Issue #19, opened 2014-11-12
Geometric zoom


tlsim:
Like the comparison series, components that draw zoomable elements should support geometric zooming. 
See also #17

tlsim:
Superseded by #216

PR #117, opened 2014-12-08
WIP - Chart Zoom

tlsim:
Not for merging yet - looking for feedback.

The idea of chartZoom is that it should be easy to add panning/zooming to a chart made with components. It manages the creation of zoom-pane rects, creates a d3 zoom behavior and sets up a function that calls each component's zoom listener on zoom..

Usage looks something like this:
```javascript 
var chartZoom = fc.utilities.chartZoom();

chartZoom
        .xScale(xScale)
        .yScale(yScale)
        .width(chartLayout.innerWidth())
        .height(chartLayout.innerHeight())
        .components([
            [candlestick, plotArea],
            [xAxis, xAxisContainer],
            [yAxis, yAxisContainer]
        ]);

chartLayout.getChartArea(chart)
        .call(chartZoom);
```
Example: https://gist.github.com/tlsim/e14ca6928db0369678de

* Components can internally manage how they should be zoomed.
* Could have the y scale domain update automatically if series data is bound.
* It's compatible with components without zoom listeners - falls back to redraw.
* Should make progress on #19 easier.


tlsim:
Superseded by chartBuilder / plotAreaOverlay. #188

PR #185, opened 2014-12-19
Fix updating clipPath in chartLayout due to WebKit bug with selecting camelCase elements

### LINKED ISSUE #184: chartLayout creates multiple clipPaths if you repeatedly 'call' it
See the attached image:

image

@jleft one for you? ;-)

jleft:
Fixes #184


ColinEberhardt:
> If the class is only used to select the the clip path which has an ID, would it not be usual to use the ID for selection and then the addition of a class is not necessary?

Good point, I've made these changes in #192

PR #193, opened 2015-01-02
Fixed the extents function

ColinEberhardt:
Minor JSDoc updates

### DIFF
--- a/components/fc.js
+++ b/components/fc.js
@@ -28,42 +28,5 @@ window.fc = {
      *
      * @namespace fc.utilities
      */
-    utilities: {},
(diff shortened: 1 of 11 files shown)


ColinEberhardt:
This simplifies the extents function. I'm pretty sure the original did not work, this line here invokes teh D3 mn function:

tunerscafe:
Much neater, I like it. 

I originally had the extents function in the fc namespace to mirror the d3 function but it is, in reality, a utility? (I know this isn't actually a question, but thoughts on this would be appreciated)

Issue #130, opened 2014-12-09
Removed hard-coded 'date' properties


ColinEberhardt:
Similar to #105 - the `date` property is hard-coded for a number of the components.

chrisprice:
#244 #245 should fix the remaining tools.

chrisprice:
* [ ] Bollinger Bands

chrisprice:
Just needs the following removing from point snap -
```
xValue = series.xValue ? series.xValue() : function(d) { return d.date; },
```

chrisprice:
Removed in #349

Issue #147, opened 2014-12-11
Create a consistent structure for the series SVG output


ColinEberhardt:
I'd like to see the same structure for each series / indicator type, with an outer `g` element with a suitable class, and inner elements:

```
<g class="moving-average">
  <path ... />
<g>
```

We also need to confirm that each series supports the data-join update pattern. You should be able to update you data and repeatedly invoke `call` passing the component with the structure above being updated

```
  chartLayout.getPlotArea(chart)
    .datum(dataOne)
    .call(volume);
```

Some of our current implementations are incorrect, and will duplicate the entire structure on update.

- [x] bollinger
- [x] moving average
- [x] volume
- [x] RSI
- [x] candlestick
- [x] comparison
- [x] line
- [x] ohlc

ColinEberhardt:
#145 should certainly test this

PR #195, opened 2015-01-05
Comparison series updates

### LINKED ISSUE #28: Documentation: Comparison series

tlsim:
- Removed ordinal colour scale (use css instead)
- No longer have to specify series names. Data input is now an array of data arrays.
- Configurable x and y values using valueAccessor #130 #105 
- Like d3 brush / axis, store element specific state on `__chart__`
  This allows for using a single instance of the component to create multiple charts which are independently interactive on zoom.
- Add JSDoc. Closes #28

### DIFF
--- a/components/fc.js
+++ b/components/fc.js
@@ -20,10 +20,17 @@ window.fc = {
      * @namespace fc.scale
      */
     scale: {},
+    /**
+     * Components which plot elements in 2 dimensional space using specified x and y scales.
+     * @namespace fc.series
+     */
     series: {},
+    /**
+     * @namespace fc.tools
+     */
     tools: {},
     /**
-     * Utility components to shorted long winded implementations of common operations.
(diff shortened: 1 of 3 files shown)


ColinEberhardt:
Thanks for the update - I was wondering, do you have a test page that demonstrates some of these updates / new features? If so, perhaps you could share in a comment or gist?

I think @jleft is currently pulling together a test page. Would be good to integrate your tests into this.

ColinEberhardt:
Does this update result in an AVG structure that mirrors the other components as per #147 ?

tlsim:
> Does this update result in an SVG structure that mirrors the other components as per #147 ?

Yes.

Here's an example which uses a single instance of fc.series.comparison to make 2 independent interactive charts. https://gist.github.com/tlsim/932320b54d9b725b440a

tlsim:
superseded by #279

PR #197, opened 2015-01-05
Extents function updated

tunerscafe:
Modified to catch some errors, mirror d3.extent behavior if necessary and handle arrays of arrays for multiple data series.

### DIFF
--- a/components/utilities/extent.js
+++ b/components/utilities/extent.js
@@ -2,16 +2,16 @@
     'use strict';
 
     /**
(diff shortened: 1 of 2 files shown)


ColinEberhardt:
These changes look good to me,

One minor thing ... just spotted, in D3 it is called `extent` rather than `extents`, can you make that change also?

ColinEberhardt:
Thanks :-)

PR #199, opened 2015-01-05
Fix Chart Layout clipPath ID

### LINKED ISSUE #198: When called on a selection multiple times, Chart Layout creates multiple clipPaths for elements with no ID
fc-chartlayout-clippath

jleft:
Fixes #198


ColinEberhardt:
Thanks :-)

PR #200, opened 2015-01-05
Series zoom

tlsim:
#19 for all the series components.

- Adds series zoom listener that sets the SVG transform attribute on the series container. The x-transform comes from d3.event, while the y-transform is computed by comparing the state of the y scale when the series was drawn to its current state. It's assumed that there's a d3.behavior.zoom with the xScale bound to it. See example https://gist.github.com/tlsim/df72009314a9e61b6f80

- Fixes bug in line series: redraw after toggling underFill did not show/hide the under-filled area.

- Removes ohlc 'line' draw method and default ohlc colours.


ColinEberhardt:
Just exploring a few alternatives ... definitely lots of good code here, just want to try re-arranging the responsibilities a little.

PR #208, opened 2015-01-05
Add getter for defs element in Chart Layout

### LINKED ISSUE #191: Getter for <defs> element in Chart Layout
Just taking a look at the Chart Layout and we could really do with a getter for the <defs> element.

Imagine you wanted to use a gradient on one of your styles, SVG gradients are defined in the <defs> element. 

You could get it from the DOM using the usual method, but a getter might be quite nice.

jleft:
Resolves #191

### DIFF
--- a/components/utilities/chartLayout.js
+++ b/components/utilities/chartLayout.js
@@ -135,12 +135,12 @@
                 dummyData = [0];
 
             // Create svg
-            chartElements.svg = container.selectAll('svg').data(dummyData);
-            chartElements.svg.enter().append('svg');
-            chartElements.svg.attr('width', width)
+            var svg = container.selectAll('svg').data(dummyData);
+            svg.enter().append('svg');
+            svg.attr('width', width)
                 .attr('height', height)
                 .style('display', 'block');
-            chartElements.svg.exit().remove();
+            svg.exit().remove();
 
             // Create group for the chart
             function roundToNearestHalfInteger(n) {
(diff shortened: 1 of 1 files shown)


components/utilities/chartLayout.js
```diff
@@ -149,9 +149,9 @@
             chartElements.chartArea = chart;
 
             // Defs - for clipping path
-            var defs = chart.selectAll('defs').data(dummyData);
-            defs.enter().append('defs');
-            defs.exit().remove();
+            chartElements.defs = chart.selectAll('defs').data(dummyData);
+            chartElements.defs.enter().append('defs');
+            chartElements.defs.exit().remove();
```
ColinEberhardt: Stylistically, I am not keen on the repetition of `chartElements.defs` in the above code. Isn't the following equivalent?

       chartElements.defs = chart.selectAll('defs').data(dummyData)
                                                  .enter().append('defs');
                                                  .exit().remove();

components/utilities/chartLayout.js
```diff
@@ -159,7 +159,7 @@
             var plotAreaClipId = 'fcPlotAreaClip_' + (element.id || nextId());
 
             // Clipping path
-            var clippingPath = defs.selectAll('#' + plotAreaClipId).data(dummyData);
+            var clippingPath = chartElements.defs.selectAll('#' + plotAreaClipId).data(dummyData);
```
ColinEberhardt: similar comment as above, `clippingPath` is being repeated

ColinEberhardt:
@tunerscafe as you requested this feature, can you confirm that this PR meets your needs? Thanks

tunerscafe:
Looks good, but agree on the suggested changes above.

components/utilities/chartLayout.js
jleft: I agree, I'm not keen on the repetition, however, the above isn't equivalent - `enter()`, `append()` and `exit()` (although `exit()` is not relevant in this case) return different selections, so chaining the methods won't work.

When using both `enter()` _and_ `exit()`, I think the repetition is necessary, otherwise you'd need to perform the selection and data-join again; it's also consistent with the other components and is what you tend to see with [data-joins](http://bost.ocks.org/mike/join/) and the [general update pattern](http://bl.ocks.org/mbostock/3808234).

Would something like this be more agreeable stylistically? I think it's easier to read.

```js
var defs = chart.selectAll('defs').data(dummyData);
defs.enter().append('defs');
defs.exit().remove();

chartElements.defs = defs;
```

components/utilities/chartLayout.js
jleft: For the clipping path, as the ID should only be set once, when the element is initially appended, I think it would be safe to chain this way (and it should be more efficient too!):

```js
var clippingPath = chartElements.defs.selectAll('#' + plotAreaClipId).data(dummyData);
clippingPath.enter().append('clipPath')
    .attr('id', plotAreaClipId);
clippingPath.exit().remove();
```

components/utilities/chartLayout.js
ColinEberhardt: Just re-read the page on data-joins, and yes, I was wrong, my suggestion wouldn't work!

I think I do prefer your proposed simplification. I find the chartLayout code a little hard to read due to this repetition.

When faced with repitition, it's also worth thinking about the use of local functions. I was able to remove some repetition here as a result:

https://github.com/ScottLogic/d3-financial-components/pull/188/files#diff-0f9b6998675c281bf770ec5c14c81ea1R195

ColinEberhardt:
@jleft - thanks for responding to my comments. Can you please do a little bit of tidyup (as per your response to my suggestion), then feel free to just merge this in.

jleft:
> @jleft - thanks for responding to my comments. Can you please do a little bit of tidyup (as per your response to my suggestion), then feel free to just merge this in.

No problem :-). Sure, will do.

PR #209, opened 2015-01-05
Move pixel alignment functionality to the top level.

### LINKED ISSUE #206: Move pixel alignment functionality to the top level
Currently the scales have an option (default enabled) to produce "pixel-rounded values" (i.e. to the nearest x.5). However, there's no equivalent for non-scaled elements e.g. axes.

An alternative approach would be to have a top level offset (0.5, 0.5) optionally applied by the chart layout to chartArea. The scales would then optionally produce integer-rounded values.

chrisprice:
Fixes #206. 

It does raise the question of whether the pixel alignment is something that should be part of the fc.axes or a wrapper but I don't think this should be fixed now.

### DIFF
--- a/components/scale/dateTime.js
+++ b/components/scale/dateTime.js
@@ -58,8 +58,7 @@
                 // When scaling dates.
                 n = linear(linearTime(x));
             }
-            var m = Math.round(n);
-            return alignPixels ? (n > m ? m + 0.5 : m - 0.5) : n;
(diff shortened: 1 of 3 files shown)


chrisprice:
@jleft BTW it also seems to fix the alignment issue you were seeing in the visual tests.

jleft:
> @jleft BTW it also seems to fix the alignment issue you were seeing in the visual tests.

:+1:

PR #214, opened 2015-01-06
Initial Visual Tests

jleft:
I expect this will change as more tests are added, but hopefully it's a reasonable starting point.


chrisprice:
I think it makes sense to get merged in even unfinished. It gives a
structure to any visual tests others may want to add.

Issue #210, opened 2015-01-06
OHLC has three render modes, keep? or simplify?


ColinEberhardt:
Currently the OHLC series has three different methods that can be used to render the series, each one having a different performance characteristic.

Do we want to keep this feature or not?

If we do keep it, we need to make it more obvious what this feature does. By choosing the more high-performance version I presume you sacrifice some functionality?

Ideally I'd like to see an equivalent visitor function as per #202

tunerscafe:
If we don't sacrifice any functionality in the higher performance mode, then it makes no sense to keep all three.

chrisprice:
Agreed, not sure if there's a difference when it comes to styling?

PR #215, opened 2015-01-07
chart zoom PR updates

ColinEberhardt:
Replaces #200 - keeping the following changes:

+ Fixes bug in line series: redraw after toggling underFill did not show/hide the under-filled area.
+ Removes ohlc 'line' draw method and default ohlc colours.

Removes the chart zoom concepts. These should be added back at a later date as a 'utility' function:

https://github.com/ScottLogic/d3-financial-components/issues/216


tlsim:
:+1: 

#210 also.

ColinEberhardt:
Thanks @tlsim

PR #218, opened 2015-01-07
Update visual tests to use renamed extent (previously extents) function

jleft:


jleft:
Looks like I was (just) beaten to it! Duplicate of #217

PR #220, opened 2015-01-08
chartLayout no longer data-joins the plot area

ColinEberhardt:
This allows the user (and chartBuilder) to data-join their own data - without this chartBuilder doesn't work.

I've also added a new visual test to verify chartBuilder works (yay!)

@jleft - finally remembered why i made that change ;-)


jleft:
Looks good :+1:, thanks for the test too :-)

> @jleft - finally remembered why i made that change ;-)

Cool, makes sense.

PR #222, opened 2015-01-08
Fix #221 - measure tool now uses correct field for origin point

### LINKED ISSUE #221: Measure tool doesn't use the correct field for the origin point
By default, the measure tool uses the same field (the target field) for both the origin and target point, which results in incorrect measurements:

measure-bug

The origin field, not the target field, should be used for the origin point.

jleft:
Fixes #221


ColinEberhardt:
Thanks :-)

PR #226, opened 2015-01-08
OHLC now has a single render mode

### LINKED ISSUE #210: OHLC has three render modes, keep? or simplify?

ColinEberhardt:
Closes #210

### DIFF
--- a/components/series/ohlc.js
+++ b/components/series/ohlc.js
@@ -84,12 +84,16 @@
             });
         };
 
-        ohlc.xScale = function(value) {
-            if (!arguments.length) {
-                return xScale;
+        function fn(value, args, prop) {
+            if (!args.length) {
+                return prop;
             }
-            xScale = value;
-            return ohlc;
+            prop = value;
+            return prop;
+        }
+
+        ohlc.xScale = function(value) {
+            return fn(value, arguments, xScale);
         };
 
         ohlc.yScale = function(value) {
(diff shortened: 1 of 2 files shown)


jleft:
Looks good :+1: 

It looks like a visual test has been added (an shown in the HTML), but the JavaScript for it hasn't - is this script to be added before the merge?

ColinEberhardt:
Thanks @jleft well spotted. I've added the missing file.

jleft:
> Thanks @jleft well spotted. I've added the missing file.

No problem :-). Though, it looks like that commit (01a8498) actually made a change to the source of the component, rather than adding the missing file.

ColinEberhardt:
> No problem :-). Though, it looks like that commit (01a8498) actually made a change to the source of the component, rather than adding the missing file.

Yeah, that was a hasty, just about to leave the office, commit. Embarrassingly rectified later ... https://github.com/ScottLogic/d3-financial-components/commit/44b6c8fd7cbbbafc51d1b1aeeef81f70917380ee

PR #227, opened 2015-01-08
Add overlay to chartLayout

### LINKED ISSUE #187: Crosshairs - overlay should be part of the chartLayout
The overlay should be constructed by the chartLayout component. Also, it should just have a rectangular path.

jleft:
Resolves #187

### DIFF
--- a/components/utilities/chartLayout.js
+++ b/components/utilities/chartLayout.js
@@ -199,13 +199,31 @@
                     .attr('class', 'plotArea');
             }
 
+            // A rectangular path, the size of the plot area, for the overlay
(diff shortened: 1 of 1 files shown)


ColinEberhardt:
@chrisprice have the changes you made meant that we do not need this overlay now?

chrisprice:
The original intention behind the overlay was to add components which
needed to listen for events to it and to add series to the plot area.
However I now think it makes more sense to just as them all to plot area
and let the consumer stack things appropriately. Therefore I don't think
there's any need for the overlay anymore.

PR #229, opened 2015-01-09
Decorate function added to bar

ColinEberhardt:
This includes:
+ This demonstrates a decorate function that is a more versatile replacement for `classForBar`
+ Some extra tests for chartBuilder (ensuring it handles resizing) and the series (ensuring that they can handle dynamic data)
+ A simplified property syntax

If OK'ed / merged, the property accessor and decorate function need to be rolled-out.


components/utilities/property.js
```diff
@@ -0,0 +1,20 @@
...
+
+    // a property that follows the D3 component convention for accessors
+    // see: http://bost.ocks.org/mike/chart/
+    fc.utilities.property = function(initialValue) {
+
+        var accessor = function(newValue) {
+                if (!arguments.length) {
+                    return accessor.value;
+                }
+                accessor.value = newValue;
```
ColinEberhardt: > Just wondering if something like this might be worthwhile?

That's a very good idea, I was not aware of functor :+1:

visual-tests/test-scripts/functional/decorate.js
```diff
@@ -0,0 +1,55 @@
...
+        .orient('bottom')
+        .ticks(5);
+
+    var priceAxis = d3.svg.axis()
+        .scale(priceScale)
+        .orient('left')
+        .ticks(5);
+
+    var color = d3.scale.ordinal()
+        .range(['#98abc5', '#8a89a6', '#7b6888', '#6b486b', '#a05d56']);
```
ColinEberhardt: Oh yes, I'd seen them before but had completely forgotten about them!

ColinEberhardt:
I'll roll-out the decorate function before merging this PR, and add a new issue to roll-out the property implementation as a separate task (it doesn't provide a functional change, so is a lower priority).

ColinEberhardt:
OK - roll-out complete. I'm leaving the task of simplifying all of our existing component properties to a later task ... #232

PR #233, opened 2015-01-11
Add a test which makes use of all indicators and updating data

jleft:


visual-tests/test-scripts/indicators/indicators-combined.js
```diff
@@ -0,0 +1,119 @@
...
+        data.shift();
+
+        // Update main chart
+        dateScale.domain(fc.utilities.extent(data, 'date'));
+        priceScale.domain(fc.utilities.extent(data, ['high', 'low']));
+        chartBuilder.setData(data);
+        chartBuilder.render();
+
+        // Update RSI chart
+        var rsi = fc.indicators.rsi()
```
ColinEberhardt: Is there any reason why rsi has to be special-cased here? I'd hope it could be treated just like all the other series

jleft: There was a reason, but I've just realised it's invalid - my mistake! I'll update the test now, before the merge.

jleft: Updated :-)

ColinEberhardt: > Updated :-)

Thanks

PR #234, opened 2015-01-11
Bar width rework

ColinEberhardt:
The width of a bar can now either be specified in pixels:

    series.barWidth(20)

Or as a fraction:

    series.(fc.utilities.fractionalBarWidth(0.75))

This allows the series to work with ordinal, linear and date axes.

This PR also removes the hard-coded date in all of the series.


jleft:
Looks great to me :+1:

---

I wonder if the fractional bar width functionality could benefit from a rounding option? As it is, it can lead to some anti-aliasing artifacts; the example below has the fractional bar width set to 1, but shows some small gaps between the first, second and third columns.

bar-aa

D3's scales have the option to round ranges: ordinal.[rangeRoundPoints](https://github.com/mbostock/d3/wiki/Ordinal-Scales#ordinal_rangeRoundPoints), ordinal.[rangeRoundBands](https://github.com/mbostock/d3/wiki/Ordinal-Scales#ordinal_rangeRoundBands), linear.[rangeRound](https://github.com/mbostock/d3/wiki/Quantitative-Scales#linear_rangeRound) etc.

ColinEberhardt:
That's a very good suggestion, I'll add rounding :-)

ColinEberhardt:
Actually - I'd like to discuss an alternative `shape-rendering = crisp-edges` ...

https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/shape-rendering

PR #236, opened 2015-01-11
Bollinger Bands and Moving Average selections are more specific to avoid conflicts

### LINKED ISSUE #235: Bollinger Bands and Moving Average selections conflict
Both [Bollinger Bands][1] and [Moving Average][2] have a default CSS class called `moving-average`. Selections are made based on this class ([Bollinger Bands][3], [Moving Average][4]).

If both components are called on the same selection, such as the plot area, the elements relating to Bollinger Bands could be selected by Moving Average - this can lead to undesirable behaviour, such as Moving Average's path being appended as a child of a Bollinger Bands path, rather than its own group.

[1]: https://github.com/ScottLogic/d3-financial-components/blob/32a4db8cc673080eb0a8d1c9ee20c36d2441115d/components/indicators/bollingerBands.js#L24
[2]: https://github.com/ScottLogic/d3-financial-components/blob/32a4db8cc673080eb0a8d1c9ee20c36d2441115d/components/indicators/movingAverage.js#L20
[3]: https://github.com/ScottLogic/d3-financial-components/blob/32a4db8cc673080eb0a8d1c9ee20c36d2441115d/components/indicators/bollingerBands.js#L157
[4]: https://github.com/ScottLogic/d3-financial-components/blob/32a4db8cc673080eb0a8d1c9ee20c36d2441115d/components/indicators/movingAverage.js#L53

jleft:
Selections are now made on element type as well as class name, to reduce the chance of selecting an unintended element. Resolves #235.

There are also a few minor changes to the style for consistency/efficiency.

Another possible solution to #235 would be to use a different class name for `moving-average` in one of the components.

### DIFF
--- a/components/indicators/bollingerBands.js
+++ b/components/indicators/bollingerBands.js
@@ -18,11 +18,6 @@
             movingAverage = 20,
             standardDeviations = 2;
 
-        var cssBandArea = 'band-area',
-            cssBandUpper = 'band-upper',
-            cssBandLower = 'band-lower',
-            cssAverage = 'moving-average';
-
         /**
         * Constructs a new instance of the bollinger bands component.
         *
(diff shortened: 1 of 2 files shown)


ColinEberhardt:
Looks good to me.

As a general point, I am not sure that we really need to ensure that our series can co-exist within the same parent `g` element. It is quite straightforward to just append a new container as the parent for a series.

jleft:
> As a general point, I am not sure that we really need to ensure that our series can co-exist within the same parent g element. It is quite straightforward to just append a new container as the parent for a series.

Agreed. However, if I understand correctly, the Chart Builder component will behave this way - all components will be called on the same parent element (the plot area)? Maybe Chart Builder could also create a group for each component and call that component on its own group only (not the plot area)?

PR #237, opened 2015-01-12
Update Scale visual tests

jleft:
- Added a test for fc.scale.dateTime with hidden weekends
- Changed draw order, so OHLC series is above gridlines


ColinEberhardt:
:+1: to more tests!

Issue #207, opened 2015-01-05
The tools don't follow the renderTargetSelection.call(component) pattern


chrisprice:
The tools need two selections -
* A selection into which to render
* A selection to bind the event listeners to

Search [this article](http://www.scottlogic.com/blog/2014/09/19/interactive.html) for overlay for an explanation of why both are needed.

The pattern used currently is -
```
var tool = fc.tools.tool().target(renderTargetSelection);
eventTargetSelection.call(tool);
```

However, this doesn't match the pattern of other components which don't require an event target -
```
renderTargetSelection.call(fc.example.component());
```

I recommend we switch to something like -
```
var tool = fc.tools.tool().on(eventTargetSelection);
renderTargetSelection.call(tool);
```

We should probably also consider adding an optional overlay to the chart layout.

Comments?

chrisprice:
When I've gone to fix this I've realised that the above is incorrect: the target selection is used for both. However, I still plan on creating a single plotAreaOverlay in chartLayout and I'll now split the selections as detailed above.

chrisprice:
Renamed ```tool().on(eventTargetSelection)``` to ```tool().eventTarget(eventTargetSelection)```.

chrisprice:
Also updated the #245 fan and #244 measure tools.

PR #240, opened 2015-01-12
Reimplement crosshairs adding snapping, decoration and events #207

chrisprice:


components/tools/crosshairs.js
```diff
@@ -3,306 +3,121 @@
...
-                var xDiff = Math.abs(xTarget - xScale(data.date));
-
-                if (xDiff < dx) {
-                    dx = xDiff;
-                    nearest = data;
+            selection.each(function() {
+                var data = this.__data__ || [];
+                if (!data.__crosshairs__) {
+                    data.__crosshairs__ = {};
+                    this.__data__ = data;
```
jleft: I wonder if rather than adding `__crosshairs__` to `data`, it could be added directly to `this`? Some D3 components stash values in `this.__chart__`, which is a similar concept (e.g [Axis][1] and [Brush][1]).

Does this need to be in its own loop or could in be within the following `selection.each()` (Line 18)?

[1]: https://github.com/mbostock/d3/blob/4e4709ce47d5c9406290dc2a467cef08002fc9b7/src/svg/axis.js#L24
[2]: https://github.com/mbostock/d3/blob/master/src/svg/brush.js#L95

jleft:
Couple of minor comments but it looks really good to me, the tests are great too - I'd be happy to see this merged :+1:

I also noticed, in the first test, that sometimes I'd hover over one bar, but the crosshairs would snap to the adjacent bar. Turning off `hideWeekends` seemed to fix this, so I expect it's caused by scale, rather than the component but something to bear in mind.

components/tools/crosshairs.js
chrisprice: Yea I'm not so keen on these bits, I'm sure there's a better way. Maybe even just doing away with data altogether like the d3 brush does...

Issue #212, opened 2015-01-06
Components lack transition support


jleft:
Currently, the components can be applied to a [selection](https://github.com/mbostock/d3/wiki/Selections), but they don't automatically support [transitions](https://github.com/mbostock/d3/wiki/Transitions) - to animate smoothly between two states. This may be a good feature to add.

[D3's axis component](http://bl.ocks.org/mbostock/1166403) is a nice example of a component which supports transitions.

chrisprice:
* [ ] /indicators/bollingerBands.js
* [ ] /indicators/movingAverage.js
* [ ] /indicators/relativeStrengthIndex.js
* [ ] /scale/gridlines.js
* [ ] /series/bar.js
* [ ] /series/candlestick.js
* [ ] /series/comparison.js
* [ ] /series/line.js
* [ ] /series/ohlc.js
* [ ] /tools/annotation.js
* [ ] /tools/callouts.js
* [ ] /tools/crosshairs.js
* [ ] /tools/fibonacciFan.js
* [ ] /tools/measure.js

jleft:
Most components have a single parameter, [`selection`](https://github.com/ScottLogic/d3-financial-components/blob/f832e51b1f8cbddb2aea0c8dce52740c5a898362/components/series/bar.js#L10), implying the component accepts a selection.

As the transition support is rolled out, components will support a _selection_ or a _transition_. I think _selection_ should be renamed in these components to something more generic.

ColinEberhardt:
Need to add transitions to error bar series #742 

I'd really like to know how we can keep on top of this one. If we get it right my feeling is that this could be a very useful and important part of d3fc..

chrisprice:
This [post](http://bost.ocks.org/mike/path/) covers transitioning lines, although it looks like it'll be hard to make that generalisable.

PR #242, opened 2015-01-13
Add transition support to bar series

jleft:
This might be a bit early to merge, but it shows a basic example of how to add transition support to a component. There is an accompanying visual test to demonstrate its behavior. See #212.

### DIFF
--- a/components/series/bar.js
+++ b/components/series/bar.js
@@ -26,25 +26,43 @@
                     .selectAll('rect')
                     .data(data, bar.xValue.value);
 
+                // "Caution: avoid interpolating to or from the number zero when the interpolator is used to generate
+                // a string (such as with attr).
+                // Very small values, when stringified, may be converted to scientific notation and
+                // cause a temporarily invalid attribute or style property value.
+                // For example, the number 0.0000001 is converted to the string "1e-7".
+                // This is particularly noticeable when interpolating opacity values.
+                // To avoid scientific notation, start or end the transition at 1e-6,
+                // which is the smallest value that is not stringified in exponential notation."
+                // - https://github.com/mbostock/d3/wiki/Transitions#d3_interpolateNumber
+                var effectivelyZero = 1e-6;
+
                 // enter
+                // entering elements fade in (from transparent to opaque)
                 series.enter()
-                    .append('rect');
+                    .append('rect')
+                    .style('opacity', effectivelyZero);
 
                 // exit
-                series.exit()
+                // exiting elements fade out (to transparent)
+                d3.transition(series.exit())
+                    .style('opacity', effectivelyZero)
                     .remove();
 
                 var width = bar.barWidth.value(data.map(x));
 
                 // update
-                series.attr('x', function(d) {
-                        return x(d) - width / 2.0;
+                // all properties of the bars will transition
+                d3.transition(series).attr('x', function(d) {
+                        return x(d) - width / 2;
                     })
                     .attr('y', function(d) { return y(d); })
                     .attr('width', width)
-                    .attr('height', function(d) { return bar.yScale.value(0) - y(d); });
+                    .attr('height', function(d) { return bar.yScale.value(0) - y(d); })
+                    .style('opacity', 1);
 
-                bar.decorate.value(series);
+                // properties set by decorate will transition too
+                bar.decorate.value(d3.transition(series));
             });
         };
 
--- a/components/series/bar.js
+++ b/components/series/bar.js
@@ -8,23 +8,19 @@
         var y = function(d) { return bar.yScale.value(bar.yValue.value(d)); };
 
         var bar = function(selection) {
-            var series, container;
-
             selection.each(function(data) {
 
                 // add a 'root' g element on the first enter selection. This ensures
                 // that it is just added once
-                container = d3.select(this)
+                var container = d3.select(this)
                     .selectAll('.bar-series')
                     .data([data]);
+
                 container.enter()
                     .append('g')
                     .classed('bar-series', true);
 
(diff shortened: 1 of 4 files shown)


ColinEberhardt:
Thanks for this - great job :-) I'm glad to see that there wasn't much work involved in adding transition support. A few random comments ...

I see you are using `d3.transition(selection)` rather than `selection.transition()` - is there a particular reason for this?

In order to make the transition configurable are you planning on adding duration and ease properties to the series?

One significant drawback of this change is that the bars always fade in, I guess the series could be implemented such that if the duration is set to zero, that no transitions are applied to the selections.

It's great that decorations are also transitioned as well :-) Could you also use this to update the enter selection so that bars grow from the bottom of the screen upwards when the first enter? And 

@chrisprice when you added your simplified data-join I think you mentioned that it required a parent `g` for each element in the series? It is probably worth checking that this works with the transitions as expected.

chrisprice:
I'm guessing the growing transition could be added by a decorator on the
(fake) enter selection?

What's the behaviour of a transition selection if you chain 2 calls to
update the same attribute? If it just overrides the old value (rather than
transition to it) then decorate could also be used to customise the
transition. Just a thought...

Is it possible to define different times for the enter, update, exit
transitions?

I suppose I'm trying to get to the bottom of whether transitions have to be
an all or nothing row affair.

chrisprice:
...and I agree about testing with simple data join. As it's pretty much a
one liner, any chance you could give it a go James?

ColinEberhardt:
> What's the behaviour of a transition selection if you chain 2 calls to 
update the same attribute? If it just overrides the old value (rather than 
transition to it) then decorate could also be used to customise the 
transition. Just a thought... 

Yes, I tried this, you can specify the same attribute value twice. The last value is the one that gets used, so ideal for our plans with decorate.

jleft:
> Thanks for this - great job :-) I'm glad to see that there wasn't much work involved in adding transition support.

No problem, thank you :-)

> I see you are using d3.transition(selection) rather than selection.transition() - is there a particular reason for this?

Yes - in this context, within a component, `selection.transition()` would apply the same, default transition every time - regardless of whether the component was called on a selection or transition.

`d3.transition(selection)` allows the user to call the component on a transition and the transition's properties (duration, delay, ease) will be inherited by the component; it also means if a selection (with no transitions) is passed to the component, the component will function as expected but with no transitions applied. See the following on [`d3.transition()`](https://github.com/mbostock/d3/wiki/Transitions#d3_transition):

> When called with an optional selection, this method typically returns the specified selection; i.e., it is a no-op. However, within the context of transition.each, this method will create a new transition for the specified selection that inherits the delay, duration and other properties of the parent transition. This is useful for implementing reusable components that can be called either on selections or on transitions, in the latter case supporting deriving concurrent transitions. An example of this is D3’s axis component.
-- *https://github.com/mbostock/d3/wiki/Transitions#d3_transition*

<br>

> In order to make the transition configurable are you planning on adding duration and ease properties to the series?

This is already supported. The transitions in the component will inherit the properties of the transition which it is called upon, the following properties would be applied to the transition in the component:

```js
// delay, duration and ease would be inherited by the bar component's transitions
chartLayout.getPlotArea()
    .datum(data)
    .transition()
    .delay(1000)
    .duration(2500)
    .ease('bounce')
    .call(bar);
```

> One significant drawback of this change is that the bars always fade in, I guess the series could be implemented such that if the duration is set to zero, that no transitions are applied to the selections.

If called on a transition this is true, but if called upon a selection no transitions would be applied (including the fade in). Setting the duration to 0 would also have the same effect.

```js
//  Bars would not fade in
chartLayout.getPlotArea()
    .datum(data)
    .call(bar);
```

> It's great that decorations are also transitioned as well :-) Could you also use this to update the enter selection so that bars grow from the bottom of the screen upwards when the first enter?

> I'm guessing the growing transition could be added by a decorator on the (fake) enter selection?

I think so, but I'm not certain, I'll try this out.

> Is it possible to define different times for the enter, update, exit transitions?

Not currently. I'm hoping this can be achieved using _named transitions_ - I'm going to look into this.

> ...and I agree about testing with simple data join. As it's pretty much a
one liner, any chance you could give it a go James?

Sure :-)

---

I think I covered everything... let me know if I've missed anything!

chrisprice:
By the looks of it, adding the transition selection into simpleDataJoin works :+1: 

There are two remaining shortcomings -

* Whilst it is possible to override any exit transitions e.g. to remove the fade out, the elements themselves will still only be removed once the transition has completed. This can lead to some strange behaviour although it's probably work-around-able by toggling the visibility of the element or somesuch.
* The update and exit transitions share the same transition settings (i.e. duration, easing function). It may be possible to fix this with named transitions but I don't think @jleft had much luck with that.

Having now enumerated the shortcomings, I'm not actually convinced that either of them are very restrictive. I think we should spend some time establishing whether it is possible to solve them but I don't think they should ultimately block this solution.

jleft:
@chrisprice - thanks for comments, I agree with all of the above :+1: 

> The update and exit transitions share the same transition settings (i.e. duration, easing function). It may be possible to fix this with named transitions but I don't think @jleft had much luck with that.

No luck so far, but I'm going to have a deeper look into how the transitions are implemented and try out a few other ideas too.

chrisprice:
Is this ready to merge (apart from the conflicts)? I think there's general acceptance of (read silence on this thread about) the shortcomings identified above, so I think it should go in.

jleft:
> Is this ready to merge (apart from the conflicts)?

Yes, I think so! :-)

chrisprice:
Nice one!

PR #244, opened 2015-01-14
Update measure tool to follow enter/update/exit pattern

chrisprice:


components/tools/measure.js
```diff
@@ -3,377 +3,170 @@
...
+                    .attr('y', function(d) { return d.source.y - paddingValue; })
+                    .style('visibility', function(d) { return d.state !== 'MEASURED' ? 'hidden' : 'visible'; })
+                    .text(measure.xLabel.value);
+
+                g.select('text.vertical')
+                    .attr('x', function(d) { return d.target.x + paddingValue; })
+                    .attr('y', function(d) { return d.source.y + (d.target.y - d.source.y) / 2; })
+                    .style('visibility', function(d) { return d.state !== 'MEASURED' ? 'hidden' : 'visible'; })
+                    .text(measure.yLabel.value);
+
```
jleft: I think `g.exit().remove();` is missing.

chrisprice: That should be handled by simpleDataJoin I think.

chrisprice: As strange as it sounds yes :) It was me trying to think up something that using data join offered over properties. I'm not convinced by it as a use-case but left it in as one of the tests.

This is the code responsible -
```
...
.attr('class', 'crosshairs-container')
.on('click', function(d) {
    d.push(d[0]);
})
.call(crosshairs);
```

chrisprice: As a general note I think we'll eventually need to revisit the use of data in the tools. Whilst it is nice, in that it provides the decorate pattern, now that we know a little more about selections I think we can easily fake that behaviour and simplify our code.

jleft: All makes sense now - thanks for the explanation,

PR #245, opened 2015-01-14
Update fan tool to use enter/update/exit pattern

chrisprice:
Apologies for the slightly messy diff, the changes need to sit on top of the measure PR. I can rebase those changes away once #244 is merged (or we can just live with it :)).

### DIFF
--- a/components/tools/fibonacciFan.js
+++ b/components/tools/fibonacciFan.js
@@ -3,395 +3,192 @@
 
     fc.tools.fibonacciFan = function() {
 
-        var target = null,
-            series = null,
-            xScale = d3.time.scale(),
-            yScale = d3.scale.linear(),
-            active = true;
+        var event = d3.dispatch('fansource', 'fantarget', 'fanclear');
 
-        var circleOrigin = null,
-            circleTarget = null,
-            lineSource = null,
-            lineA = null,
-            lineB = null,
-            lineC = null,
-            fanArea = null;
+        var fan = function(selection) {
 
-        var phase = 1,
-            locationOrigin = null,
-            locationTarget = null;
-
-        var fibonacciFan = function() {
-
-            var root = target.append('g')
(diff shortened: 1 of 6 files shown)


chrisprice:
FYI, this removes the bespoke animation.

components/tools/fibonacciFan.js
```diff
@@ -3,395 +3,192 @@
...
-                if (field !== null) {
+                var container = d3.select(this)
+                    .style('pointer-events', 'all')
+                    .on('mouseenter.fan', mouseenter);
 
-                    return {point: point, field: field};
+                if (!data.__fan__.overlay) {
+                    container.append('rect')
+                        .style('visibility', 'hidden');
+                    data.__fan__.overlay = true;
```
jleft: I appreciate this is the same for all tools which have had similar updates, so I'm happy with this for now, but when clearing the data and calling the component it can lead to a lot of unnecessary elements being appended:

image

What do you think about using a data join with dummy data instead, like this:

```js
// Data join with a single datum - to create only one overlay
var overlay = container.selectAll('rect.overlay').data([0]);

// Enter
overlay.enter().append('rect')
    .attr('class', 'overlay')
    .style('visibility', 'hidden');

// Update
overlay.attr('x', fan.xScale.value.range()[0])
    .attr('y', fan.yScale.value.range()[1])
    .attr('width', fan.xScale.value.range()[1])
    .attr('height', fan.yScale.value.range()[0]);

// Exit - remove any unnecessary overlays
overlay.exit().remove();
```

It would also mean that we don't have to mess about with data, the opening loop could be removed and something like this could be added to the main loop:

```js
this.__data__ = this.__data__ || [];
```

jleft:
Looks good :-) once the merge conflicts are resolved, I'd say it's good to merge :+1:

PR #251, opened 2015-01-17
Stacked bar implementation

ColinEberhardt:
A stacked bar. The API for this series is a little different to the others, and slightly more constrained. The other series types expose accessor functions for x & y values, whereas this one expects you to supply an array of objects where the properties are the y values, apart from the one identified as the x value.

Currently the example for this series does not work with crosshair because the crosshair component doesn't work ordinal axes. I'm going to take a look at this next.


ColinEberhardt:
I've added pixel snapping into the visual test, good to see the crosshair is sufficiently versatile to accommodate this!

I'll have to look into a way of packaging up the pixel snapping logic.

components/series/stackedBar.js
```diff
@@ -0,0 +1,100 @@
...
+
+                // create a join for each bar
+                var bar = g.selectAll('rect')
+                    .data(function(d) { return objectDatapointToArray(d); })
+                    .enter()
+                    .append('rect');
+
+                // compute the bar width from the x values
+                var xValues = data.map(function(d) {
+                    return stackedBar.xScale.value(d[stackedBar.xValueKey.value]);
```
chrisprice: Sorry I haven't checked the implementation, but does doing this in enter mean that something about the xScale could change such that this value would need to update?

chrisprice: Actually scratch that, it's not being done in enter... It's too early...

components/series/stackedBar.js
```diff
@@ -0,0 +1,100 @@
...
+            //   { xValue: 'North Tyneside', name: 'conservative', previousValue: 23, currentValue: 78},
+            // ]
+            function objectDatapointToArray(obj) {
+                var values = [];
+                var yTotal = 0;
+                var xVal = obj[stackedBar.xValueKey.value];
+                for (var propertyName in obj) {
+                    if (obj.hasOwnProperty(propertyName) && propertyName !== stackedBar.xValueKey.value) {
+                        var previous = yTotal;
+                        yTotal += +obj[propertyName];
```
chrisprice: Prefer Number(x) to +x.

chrisprice:
Looks good, glad to see it integrated with crosshairs (with only a few minor snags) :+1: 

Playing devils advocate I have 2 concerns with the stacked bar -

* There's quite a lot of data transformation done in the update. Would it be better to supply the data in a different format?
* I've not seen a stacked bar on a financial chart. Other than it being a good test of the flexibility of the components it integrates with, does it really have a place in the library?

components/series/stackedBar.js
ColinEberhardt: :+1:

ColinEberhardt:
> There's quite a lot of data transformation done in the update. Would it be better to supply the data in a different format?

My thinking here was to (hopefully) make it easier for the user of this series, they don't have to sum the data in order to stack it.

> I've not seen a stacked bar on a financial chart. Other than it being a good test of the flexibility of the components it integrates with, does it really have a place in the library?

Erm... other than trade-stack, no I haven't seen it used at all! I guess I am still very much in the mindset that this could eventually be a more generic set of charting components.

chrisprice:
Yea, I suppose there isn't an ideal solution, it has to be done somewhere. I hadn't quite appreciated it's role in creating an array item per (stacked) bar.

I agree that it eventually could be included, but is it better to keep the focus on financial related components for this first release? I'm worried that we'll get carried away implementing all the combinations e.g. side-by-side  bars, stacked areas, etc. when we could be polishing what we have and getting it used in the real world. 

Don't get me wrong, I think we should definitely still include the stacked bar but maybe in the visual tests for now. That way we can make sure we don't restrict our ability to include it in the future without it being a distraction.

ColinEberhardt:
Thanks for the comments, we'll keep this out of the docs (or whatever we decide to release with) for now.

PR #253, opened 2015-01-18
WIP: Update line series to use properties/decorate pattern

chrisprice:


chrisprice:
I've got an alternative idea, I'd like to finish that first.

ColinEberhardt:
replaced by #254

PR #254, opened 2015-01-18
WIP: Separate out line area point

chrisprice:
As an alternative to putting area and point into line, separating them all out so that they can be used by the indicators?

### DIFF
--- a/components/series/area.js
+++ b/components/series/area.js
@@ -1,0 +1,45 @@
+(function(d3, fc) {
+    'use strict';
+
+    fc.series.area = function() {
+
+        // convenience functions that return the x & y screen coords for a given point
+        var x = function(d) { return area.xScale.value(area.xValue.value(d)); };
+        var y0 = function(d) { return area.yScale.value(area.y0Value.value(d)); };
+        var y1 = function(d) { return area.yScale.value(area.y1Value.value(d)); };
+
+        var area = function(selection) {
+
+            selection.each(function(data) {
+
+                var container = d3.select(this)
+                    .selectAll('.area-series')
+                    .data([data]);
+
+                container.enter()
+                    .append('g')
+                    .classed('area-series', true)
+                    .append('path');
+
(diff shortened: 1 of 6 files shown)


ColinEberhardt:
Nice job, just had a quick skim through the code, all looks good to me. I agree that area should have properties of `yValue` and `y0Value` - and the functor concept works **very** well here, allowing you to set a constant, to shift the baseline, or supply a function that turns this into a band.

This should definitely be rolled out to bar as well :-)

I hacked around with decorate a bit, and all works well:

image

Code:

```
var point = fc.series.point()
    .xScale(dateScale)
    .yScale(priceScale)
    .yValue(fc.utilities.valueAccessor('close'))
    .decorate(function(sel) {
        sel.select('circle')
            .attr('style', 'fill: red')
            .attr('r', function(d) {return (d.high - d.low) * 10; });

        sel.append('text')
            .attr('x', function(d) { return dateScale(d.date); })
            .attr('y', function(d) { return priceScale(d.close); })
            
            .text(function(d) { return d.close.toFixed(0); });
    });
```

My only thought is, now that we have a group element (courtesy of simple data-join), I think it would be better if bar and point series positioned their elements using a translate on the group. 

With the example above, where I am adding text to each point, the x & y attributes of the text element would simply be an offset from the origin of each datapoint location, so no need to mess about with scales.

chrisprice:
Good stuff :+1: 

I'd thought about something similar but initially dismissed it to support someone who wanted to use only the x value and fix y at some pixel position (or vice versa). However, thinking about it that's a very niche edge case. Also, if they wanted to do that, they could always just use the scale to apply the opposite offset.

chrisprice:
Created #258 to cover the above improvement.

chrisprice:
I think that means this can go in unless there are any objections?

ColinEberhardt:
Yep, looks good to me

Issue #247, opened 2015-01-16
Roll out simpleDataJoin/decorate pattern


chrisprice:
* [x] /indicators/bollingerBands.js
* [x] /indicators/movingAverage.js
* [x] /indicators/relativeStrengthIndex.js
* [x] /scale/gridlines.js
* [x] /series/bar.js
* [x] /series/candlestick.js
* [x] /series/comparison.js
* [x] /series/line.js
* [x] /series/ohlc.js
* [x] /tools/annotation.js
* [x] /tools/callouts.js
* [x] /tools/crosshairs.js
* [x] /tools/fibonacciFan.js
* [x] /tools/measure.js

chrisprice:
I'll take candlestick

chrisprice:
Done: #250

chrisprice:
OHLC #252

chrisprice:
Line #253, although it doesn't actually use simpleDataJoin. There's only 2 path elements, no matter the length of the line.

chrisprice:
I'll pick up comparison

chrisprice:
Done

Issue #232, opened 2015-01-09
Roll out the simplified approach to exposing properties


ColinEberhardt:
As introduced here: https://github.com/ScottLogic/d3-financial-components/pull/229/files#diff-a30185c4c94d14d4cf8dabce084e834fR6

jleft:
#### Indicators
 + [ ] Bollinger Bands
 + [ ] Moving Avergage
 + [ ] RSI

#### Scale
 + [ ] Date Time
 + [ ] Linear
 + [ ] Gridlines

#### Series
 + [ ] Bar
 + [ ] Candlestick
 + [ ] Comparison
 + [ ] Line
 + [ ] OHLC

#### Tools
 + [ ] Annotation
 + [ ] Callouts
 + [ ] Crosshairs
 + [ ] Fibonacci Fan
 + [ ] Measure

#### Utilities
 + [ ] Chart Builder
 + [ ] Chart Layout
 + [ ] Data Generator

chrisprice:
#244 #245 should fix fan and measure

chrisprice:
This has been superseded by #275

Issue #255, opened 2015-01-19
Standardise the SVG element structure


chrisprice:
We should either -
* Always have a root element appended to the selection by the component to which all other in the component. This root element will have a CSS class and all CSS will use a descendant selector on that class name.
* Never add a root element. All CSS should target CSS classes as appropriate.

chrisprice:
Having worked with the components as part of the bollinger bands refactoring I'm strongly for not adding root elements.

jleft:
> Having worked with the components as part of the bollinger bands refactoring I'm strongly for not adding root elements.

:+1: I agree, I've found this to be the most appropriate solution. For what it's worth, D3's axis component does the same thing.

PR #256, opened 2015-01-19
Update Annotation tool

### LINKED ISSUE #223: Annotation tool generates elements with duplicate (non-unique) IDs
Unless an [annotation's `index`](https://github.com/ScottLogic/d3-financial-components/blob/67a49925df10007a31ea5b692e4c66e1c4bf2fba/components/tools/annotation.js#L6) is explicitly set, multiple instances of the component will generate elements with the same ID:

annotation-ids

I think the component's default behavior should avoid this issue, similar to the clipping path's ID in chart layout.

I'm not sure if a default ID is necessary at all? If not, then the simplest solution would be to remove it.

Otherwise, I'd suggest allowing the user the explicitly set an ID for the element (it's their responsibility to ensure it's unique), but if the ID isn't explicitly set, then the index should be incremented (in a similar way to [chartLayout](https://github.com/ScottLogic/d3-financial-components/blob/67a49925df10007a31ea5b692e4c66e1c4bf2fba/components/utilities/chartLayout.js#L455)).

Thoughts?

### LINKED ISSUE #224: Annotation tool applies same padding horizontally and vertically to text label
It would be nice if different values could be specified for the horizontal and vertical position of the text label.

### LINKED ISSUE #225: Annotation tool label text formatting could be more flexible
The annotation's label text [follows the following format](https://github.com/ScottLogic/d3-financial-components/blob/67a49925df10007a31ea5b692e4c66e1c4bf2fba/components/tools/annotation.js#L37):

```js
.text(yLabel + ': ' + formatCallout(yValue))
```

This is fairly restrictive and, at least, forces the label to contain a colon, followed by a space.

This could be simplified (which will also make the `yLabel` property redundant):

```js
.text(formatCallout(yValue))
```

This affords more flexibility to the label formatting; any additional text can still be added relatively easily to the label, for example:

```js
var annotation101 = fc.tools.annotation()
    .xScale(dateScale)
    .yScale(priceScale)
    .yValue(101)
    .formatCallout(function(d) { return 'Some text: ' + d3.format('.1f')(d); });
```

### LINKED ISSUE #230: Annotation component doesn't follow the general update pattern
The annotation component should follow the general update pattern (enter, update, exit...).

jleft:
- Follows general update pattern (resolves #230)
 - Follows simplified approach to exposing properties (#232)
 - More flexible label formatting with formatLabel (resolves #225)
 - Uses simpleDataJoin (#247)
 - Decorate added (can be used to override padding - resolves #224)
 - ID generation removed and is no longer the responsibility of the component (resolves #223)

### DIFF
--- a/components/tools/annotation.js
+++ b/components/tools/annotation.js
@@ -3,95 +3,53 @@
 
     fc.tools.annotation = function() {
 
-        var index = 0,
-            xScale = d3.time.scale(),
-            yScale = d3.scale.linear(),
-            yLabel = '',
-            yValue = 0,
-            padding = 2,
-            formatCallout = function(d) { return d; };
-
-        var root = null,
-            line = null,
-            callout = null;
-
         var annotation = function(selection) {
+            selection.each(function(data) {
+                var x = annotation.xScale.value,
+                    y = annotation.yScale.value,
+                    yValue = annotation.yValue.value;
 
-            root = selection.append('g')
-                .attr('id', 'annotation_' + index)
-                .attr('class', 'annotation');
+                var container = d3.select(this);
 
-            line = root.append('line')
-                .attr('class', 'marker')
-                .attr('x1', xScale.range()[0])
-                .attr('y1', yScale(yValue))
-                .attr('x2', xScale.range()[1])
-                .attr('y2', yScale(yValue));
+                // Create a group for each annotation
+                var g = fc.utilities.simpleDataJoin(container, 'annotation', data, yValue);
 
+                // Added the required elements - each annotation consists of a line and text label
+                var enter = g.enter();
+                enter.append('line')
+                    .attr('class', 'marker');
+                enter.append('text')
+                    .attr('class', 'label');
 
-            callout = root.append('text')
-                .attr('class', 'callout')
-                .attr('x', xScale.range()[1] - padding)
-                .attr('y', yScale(yValue) - padding)
-                .attr('style', 'text-anchor: end;')
-                .text(yLabel + ': ' + formatCallout(yValue));
+                // Update the line
+                g.selectAll('line.marker').attr('x1', x.range()[0])
+                    .attr('y1', function(d) {
+                        return y(yValue(d));
+                    })
+                    .attr('x2', x.range()[1])
+                    .attr('y2', function(d) {
+                        return y(yValue(d));
+                    });
+
+                // Update the text label
+                var paddingValue = annotation.padding.value.apply(this, arguments);
+                g.selectAll('text.label').attr('x', x.range()[1] - paddingValue)
+                    .attr('y', function(d) {
+                        return y(yValue(d)) - paddingValue;
+                    })
+                    .attr('style', 'text-anchor: end;')
+                    .text(annotation.formatLabel.value);
+
+                annotation.decorate.value(container);
+            });
         };
 
-        annotation.index = function(value) {
-            if (!arguments.length) {
-                return index;
-            }
-            index = value;
-            return annotation;
-        };
-
-        annotation.xScale = function(value) {
-            if (!arguments.length) {
-                return xScale;
-            }
-            xScale = value;
-            return annotation;
-        };
-
-        annotation.yScale = function(value) {
-            if (!arguments.length) {
-                return yScale;
-            }
-            yScale = value;
-            return annotation;
-        };
-
-        annotation.yValue = function(value) {
-            if (!arguments.length) {
-                return yValue;
-            }
-            yValue = value;
-            return annotation;
-        };
-
-        annotation.yLabel = function(value) {
-            if (!arguments.length) {
-                return yLabel;
-            }
-            yLabel = value;
-            return annotation;
-        };
-
-        annotation.padding = function(value) {
-            if (!arguments.length) {
-                return padding;
-            }
-            padding = value;
-            return annotation;
-        };
-
-        annotation.formatCallout = function(value) {
-            if (!arguments.length) {
-                return formatCallout;
-            }
-            formatCallout = value;
-            return annotation;
-        };
+        annotation.xScale = fc.utilities.property(d3.time.scale());
+        annotation.yScale = fc.utilities.property(d3.scale.linear());
+        annotation.yValue = fc.utilities.functorProperty(fc.utilities.fn.identity);
+        annotation.formatLabel = fc.utilities.functorProperty(annotation.yValue.value);
+        annotation.padding = fc.utilities.functorProperty(2);
+        annotation.decorate = fc.utilities.property(fc.utilities.fn.noop);
 
         return annotation;
     };
--- a/components/tools/annotation.js
+++ b/components/tools/annotation.js
@@ -5,9 +5,9 @@
 
         var annotation = function(selection) {
             selection.each(function(data) {
-                var x = annotation.xScale.value,
-                    y = annotation.yScale.value,
-                    yValue = annotation.yValue.value;
+                var yValue = annotation.yValue.value,
+                    x = annotation.xScale.value,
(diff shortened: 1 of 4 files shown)


components/tools/annotation.js
```diff
@@ -3,95 +3,53 @@
...
+                var enter = g.enter();
+                enter.append('line')
+                    .attr('class', 'marker');
+                enter.append('text')
+                    .attr('class', 'label');
+
+                // Update the line
+                g.selectAll('line.marker').attr('x1', x.range()[0])
+                    .attr('y1', function(d) {
+                        return y(yValue(d));
```
chrisprice: I've been using y as the combination of yValue and yScale. Maybe that would be appropriate here as well?

components/tools/annotation.js
```diff
@@ -3,95 +3,53 @@
...
+
+                var container = d3.select(this);
+
+                // Create a group for each annotation
+                var g = fc.utilities.simpleDataJoin(container, 'annotation', data, yValue);
+
+                // Added the required elements - each annotation consists of a line and text label
+                var enter = g.enter();
+                enter.append('line')
+                    .attr('class', 'marker');
```
chrisprice: I've avoiding adding classes if they aren't required to perform a selection e.g. in this case you could just use the element name?

components/tools/annotation.js
```diff
@@ -3,95 +3,53 @@
...
-            }
-            formatCallout = value;
-            return annotation;
-        };
+            selection.each(function(data) {
+                var x = annotation.xScale.value,
+                    y = annotation.yScale.value,
+                    yValue = annotation.yValue.value;
+
+                var container = d3.select(this);
```
chrisprice: This component follows the pattern of not creating a container element (which seems appropriate here). This is more of a comment for the issue #255 than a request to change anything.

components/tools/annotation.js
```diff
@@ -3,95 +3,53 @@
...
+                        return y(yValue(d));
+                    });
+
+                // Update the text label
+                var paddingValue = annotation.padding.value.apply(this, arguments);
+                g.selectAll('text.label').attr('x', x.range()[1] - paddingValue)
+                    .attr('y', function(d) {
+                        return y(yValue(d)) - paddingValue;
+                    })
+                    .attr('style', 'text-anchor: end;')
```
chrisprice: Should this be moved into CSS?

components/tools/annotation.js
```diff
@@ -3,95 +3,53 @@
...
+                    .text(annotation.formatLabel.value);
+
+                annotation.decorate.value(container);
+            });
+        };
+
+        annotation.xScale = fc.utilities.property(d3.time.scale());
+        annotation.yScale = fc.utilities.property(d3.scale.linear());
+        annotation.yValue = fc.utilities.functorProperty(fc.utilities.fn.identity);
+        annotation.formatLabel = fc.utilities.functorProperty(annotation.yValue.value);
```
chrisprice: formatLabel or just label?

components/tools/tools.css
```diff
@@ -56,15 +56,12 @@
 }
 
 /* Annotations */
-.annotation .marker {
+.annotation>line.marker {
```
chrisprice: See CSS comment above

chrisprice:
Looks like a vast improvement, nice one!

components/tools/annotation.js
jleft: Agreed.

components/tools/annotation.js
jleft: Yeah, good spot!

components/tools/annotation.js
jleft: This was based on #225 

I think `label` would be more appropriate.

jleft:
> Looks like a vast improvement, nice one!

Thanks for the feedback :-)

components/tools/annotation.js
chrisprice: Ops, memory fail! Agreed label is better though.

components/tools/annotation.js
jleft: Sure, just elements, without classes, would work here.

My initial though was, if the elements had class names, it might be easier for the user if they were making use of decorate. For example, if they wanted to append an additional line, they may not want the default annotation styling applied (I know there are easy ways around this, but it could be annoying for the user). I suppose this is an edge case and wouldn't affect the vast majority of use cases.

I agree that using just element is more appropriate here - I'll make the change.

jleft:
@chrisprice - I've made the amends based on your comments, how does this look to you now? Let me know if I should squash the commits.

components/tools/annotation.js
```diff
@@ -3,96 +3,48 @@
...
-
-        annotation.formatCallout = function(value) {
-            if (!arguments.length) {
-                return formatCallout;
-            }
-            formatCallout = value;
-            return annotation;
-        };
+            selection.each(function(data) {
+                var yValue = annotation.yValue.value,
```
chrisprice: Sorry I'm being picky but I find these variables confusing. 
* yValue is an alias to an property accessor function
* x is an alias to a scale function
* y is an alias to a scaled property accessor function

Would we be better with -
* xScaleRange = annotation.xScale.value.range()
* y as is
* drop yValue?

components/tools/annotation.js
```diff
@@ -3,96 +3,48 @@
...
+                // Create a group for each annotation
+                var g = fc.utilities.simpleDataJoin(container, 'annotation', data, yValue);
+
+                // Added the required elements - each annotation consists of a line and text label
+                var enter = g.enter();
+                enter.append('line');
+                enter.append('text');
+
+                // Update the line
+                g.selectAll('line')
```
chrisprice: Why selectAll?

components/tools/annotation.js
```diff
@@ -3,96 +3,48 @@
...
+                    .attr('x1', x.range()[0])
+                    .attr('y1', function(d) { return y(d); })
```
chrisprice: The following identity holds, so I think this should simplify :)
```function(d) { return y(d); } === y```

(also the same 2 lines down)

File: components/tools/annotation.js [hunk 10]
```diff
@@ -3,96 +3,48 @@
...
+                    .attr('x2', x.range()[1])
+                    .attr('y2', function(d) { return y(d); });
+
+                // Update the text label
+                var paddingValue = annotation.padding.value.apply(this, arguments);
+                g.selectAll('text')
```
chrisprice: selectAll?

visual-tests/index.html
```diff
@@ -191,6 +192,8 @@
     <script src="test-scripts/scale/fc-time-d3-linear.js"></script>
     <script src="test-scripts/scale/d3-time-d3-linear.js"></script>
 
+    <script src="test-scripts/tools/annotation.js"></script>
```
chrisprice: Nit pick: can this line be moved down to keep the files alphabetical?

visual-tests/test-scripts/tools/annotation.js
```diff
@@ -0,0 +1,104 @@
...
+        .padding(7)
+        .label(function(d) {
+            return 'Last close: ' + d3.format('.6f')(d.close);
+        });
+
+    var annotationDecimal = fc.tools.annotation()
+        .xScale(dateScale)
+        .yScale(priceScale)
+        .decorate(function(selection) {
+            selection.selectAll('line')
```
chrisprice: selectAll?

chrisprice:
Almost...

I think a handful of commits are fine btw, if it gets more than 5 then I might consider squashing or if all of the commit messages are "fix review comments" :)

jleft:
Thanks again for the comments, @chrisprice, really appreciate the feedback :)

I agree with all of the above and have made the appropriate changes.

PR #261, opened 2015-01-22
WIP: Updated dateTime axis

ColinEberhardt:
Just thought I'd share my current work in progress - any feedback welcome!

I've made quite a few changes to the way our dateTime axis works:
 + It now adapts a d3 time axis rather than a d3 linear axis. This means that we benefit from d3's built in logic for formatting ticks on a time axis (which is actually pretty good).
 + The task of computing the discontinuities is delegated to a `discontinuityProvider`, which should hopefully make it more extensible.

It should be possible to generalise this pattern so that it works for any axis type - although for now I want to focus on date-time.

There's still quite a bit more work to be done before this can be merged, but as it is a big change I wanted to give people a heads-up.


ColinEberhardt:
see #262

PR #262, opened 2015-01-23
Date time axis refactor

ColinEberhardt:
I've made quite a few changes to the way our dateTime axis works:

 + It now adapts a d3 time axis rather than a d3 linear axis. This means that we benefit from d3's built in logic for formatting ticks on a time axis (which is actually pretty good).
 + The task of computing the discontinuities is delegated to a `discontinuityProvider`, which should hopefully make it more extensible.
 + Also added unit tests (thanks @chrisprice )


ColinEberhardt:
Just realised, I committed the d3 source so that I could run unit tests offline. Is there a better way to do this?

chrisprice:
I have a gitignore'd version that I use for offline use as well... I don't like the idea of committing one to the repo though.

components/scale/dateTime.js
```diff
@@ -1,401 +1,65 @@
...
+    function dateTimeScale(adaptedScale, discontinuityProvider) {
 
         if (!arguments.length) {
-            linear = d3.scale.linear();
-            baseDomain = [new Date(0), new Date(0)];
-            alignPixels = true;
-            hideWeekends = false;
-            padEnds = false;
+            adaptedScale = d3.time.scale();
+            discontinuityProvider = fc.scale.discontinuity.identity;
```
chrisprice: Why identity rather than noop?

components/scale/dateTime.js
```diff
@@ -1,401 +1,65 @@
...
+
+        function scale(date) {
+            var domain = adaptedScale.domain();
+            var range = adaptedScale.range();
+
+            // The discontinuityProvider is responsible for determine the distance between two points
+            // along a scale that has discontinuities (i.e. sections that have been removed).
+            // the scale for the given point 'x' is calculated as the ratio of the discontinuous distance
+            // over the domain of this axis, versus the discontinuous distance to 'x'
+            var totalDomainDistance = discontinuities().getDistance(domain);
```
chrisprice: With either the identity or noop default implementations it looks like this line will error?

ColinEberhardt:
Yeah, to be honest I only spotted the d3 library had been committed when I saw the massive number of additions in this PR. I was aiming for a net minus ;-)

components/scale/dateTime.js
ColinEberhardt: getDistance is not a noop in either implementations

chrisprice: Sorry, misread ```fc.scale.discontinuity.identity``` as ```fc.utilities.fn.identity```....

components/scale/dateTime.js
```diff
@@ -1,401 +1,65 @@
...
-        scale.hideWeekends = function(value) {
+        scale.domain = function(x) {
             if (!arguments.length) {
-                return hideWeekends;
+                return adaptedScale.domain();
             }
-            hideWeekends = value;
+            // clamp the upper and lower domain values to ensure they
+            // do not fall within a discontinuity
+            var domainLower = scale.discontinuityProvider.value.clampUp(x[0]);
```
chrisprice: ```discontinuity()```?

ColinEberhardt: Good spot.

components/scale/discontinuity/identity.js
```diff
@@ -0,0 +1,21 @@
+(function(d3, fc) {
+    'use strict';
+
+    function noop(date) { return date; }
```
chrisprice: identity?

jleft:
> Just realised, I committed the d3 source so that I could run unit tests offline. Is there a better way to do this?

How about adding d3 to the package.json (through npm) and reference it from `node_modules`? `node_modules` is already ignored by git, so d3 wouldn't be committed to the repo.

chrisprice:
I've tried to get my head around the changes but I'm still struggling a little bit. Maybe you can talk us through it? It probably also suggests it needs a bit more documentation?

ColinEberhardt:
Changes:

 + Add readme to discontinuity folder
 + change getDistance => distance, applyOffset => offset
 + remove distance overload

chrisprice:
One other thing to consider, should the discontinuity providers be
singletons?

chrisprice:
(n.b. I'm not sure)

ColinEberhardt:
> One other thing to consider, should the discontinuity providers be singletons?

They are intended to be singletons (and stateless), otherwise, they need to be made clone-able. I'll add that to the README.

chrisprice:
What if they want to have state? Is there any reason to enforce that
restriction?

ColinEberhardt:
> How about adding d3 to the package.json (through npm) and reference it from node_modules? node_modules is already ignored by git, so d3 wouldn't be committed to the repo.

Good call - I've done that!

ColinEberhardt:
I think I've addressed all the review comments, if there's nothing else, I'll merged later today. Thanks for the feedback.

visual-tests/axis.html
```diff
@@ -0,0 +1,136 @@
...
+        offset: <input type="range" min="1" max="50" value="1" id="offset"></input>
+        ticks: <input type="range" min="5" max="50" value="5" id="ticks"></input>
+        <input type="checkbox" id="nice">nice</input>
+
+        <div>
+            <svg id="chart" width="900" height="450">
+        </div>
+    </section>
+
+    <script src="//d3js.org/d3.v3.js"></script>
```
jleft: Now that D3 is in `node_modules`, should it be referenced from there?

visual-tests/axis.html
```diff
@@ -0,0 +1,136 @@
...
+        var dateScale = d3.time.scale()
+            .domain(fc.utilities.extent(data, 'date'))
+            .range([0, 700]);
+
+        drawChart(dateScale, 100, 'd3.time.scale');
+
+
+        // ##################################################################
+        var dateScale2 = fc.scale.dateTime()
+            .hideWeekends(true)
```
jleft: `hideWeekends()` is no longer part of `dateTime()`

visual-tests/axis.html
```diff
@@ -0,0 +1,136 @@
...
+            .domain(fc.utilities.extent(data, 'date'))
+            .range([0, 700]);
+
+        drawChart(dateScale2, 250, 'fc.scale.dateTime');
+
+        // ##################################################################
+        var dateScale3 = fc.scale.dateTime2()
```
jleft: `dateTime2()` to `dateTime()`?

visual-tests/axis.html
```diff
@@ -0,0 +1,136 @@
...
+        <div>
+            <svg id="chart" width="900" height="450">
+        </div>
+    </section>
+
+    <script src="//d3js.org/d3.v3.js"></script>
+    <script src="../dist/d3-financial-components.js"></script>
+    
+
+    <script>
```
jleft: Currently, JSHint only checks .js files (not .html).

Might it be worth moving the script into its own .js file? Or adding .html files to JSHint?

Very minor point, but there is a mix of double quotes and single quotes; there are a couple of missing semicolons too (lines 83 and 130).

jleft:
Looks like a great improvement! A couple of minor comments with the visual test, but I'd be happy to see this merged :+1:

visual-tests/axis.html
ColinEberhardt: Good idea - done.

visual-tests/axis.html
ColinEberhardt: That whole file was checked in by mistake - sorry. But thanks for spotting it as part of your review :-)

ColinEberhardt:
@jleft - thanks for the review comments.

Issue #260, opened 2015-01-19
Consider alternative layout options


ColinEberhardt:
What about plottable.js? Or flex box?

ColinEberhardt:
There's a JS implementation of CSS here:

https://github.com/facebook/css-layout

Might be useful.

PR #265, opened 2015-01-29
Layout using flexbox

ColinEberhardt:
Just quickly threw something together using the Facebook css-layout code. This looks like it has potential :-) 
see #260 

You can add flex box CSS to SVG elements via `data-style`, the layout component then performs the required layout and transforms / translates each `g` element accordingly.


ColinEberhardt:
+ data-style => layout-css
 + width and height are now prefixed. I did make them properties of the SVG elements but found that this made layout debugging quite hard. Having them as attributes is easier for users IMO
 + layout now measures the parent container and sets the SVG width / height accordingly
 + Added a more complex demo with axis labels and a dummy legend. The approach works really well :-)

I think this is pretty much done.

chrisprice:
On a mobile so can't comment inline...

* I had a go using property descriptors to proxy the elements themselves. In the end I think your approach is easier to understand. 
* Is it presumptuous to assume the svg should be sized to the parent element? What about implicitly setting the "layout" dimensions of the svg based on its client dimensions (if not explicitly set)? You can then use css to make the svg fill the parent if that's the desired behaviour. 
* I can't find a reference for my assertion about custom attributes containing dashes, I'm pretty sure I was wrong about that sorry! Maybe ```data-layout```?
* I'm not convinced about writing the dimensions back onto the elements as attributes for programmatic access, but I accept the debugging argument and I suppose it is consistent with writing the transform back onto the element.
* Random thought: d3 seems to encourage programmatic construction of the structure, I assume that's still possible with this technique, but what would it look like?

I also realised I was talking yet more rubbish about referencing github repos (which don't contain package.json files), but I did find a module which allowed it; I'll share it when I get back to a laptop.

Overall though I think this is a really very neat solution!

chrisprice:
> I also realised I was talking yet more rubbish about referencing github repos (which don't contain package.json files), but I did find a module which allowed it; I'll share it when I get back to a laptop.

https://github.com/chrisprice/d3-financial-components/compare/ScottLogic:master...chrisprice:napa

chrisprice:
> Random thought: d3 seems to encourage programmatic construction of the structure, I assume that's still possible with this technique, but what would it look like?

It could look like -
```js
    // Create the structure
    var svg = d3.select('#container')
        .append('svg')
        .style({
            'width': '1024px',
            'height': '768px'
        })
        .layout('flexDirection', 'column');

    var rows = [
        svg.append('g')
            .layout({
                'flexDirection': 'row',
                'flex': 1
            }),
        svg.append('g')
            .layout({
                'flexDirection': 'row',
                'height': 100
            })
    ];

    var plotAreaContainer = rows[0].append('g')
        .layout('flex', 1);

    var yAxisContainer = rows[0].append('g')
        .layout('width', 100);

    var xAxisContainer = rows[1].append('g')
        .layout({
            'flex': 1,
            'marginRight': 100
        });

    // Perform the layout
    svg.layout();

    // Retrieve the calculated width
    var xAxisContainerWidth = xAxisContainer.layout('width');
```

ColinEberhardt:
Re: Napa - saw it, but wasn't too keen! I forked css-layout and added package.json. I made a PR and it looks like they are happy with it, so we should be able to reference the source of that framework soon.

Re: data prefixes, I didn't read the spec, but assumed that with angular using ng- prefixes, not data prefixes were fine

Also on a mobile, so can't really follow your code terribly well 😉

chrisprice:
Ah nice one, what put you off napa btw?

Finally found some decent references and, unsurprisingly, we're not the first people to have [this](https://github.com/amcss/attribute-module-specification/issues/4) [discussion](https://github.com/AmpersandJS/ampersand/issues/21). Consensus seems to be as I had originally thought spec says ```data-*``` but as most browsers ignore (most) hyphenated properties people (read angular) have adopted a ```namespace-*``` approach. 

The kicker though, is that ```data-*``` attributes [aren't actually valid](http://stackoverflow.com/questions/15532371/do-svg-docs-support-custom-data-attributes) for SVG elements as we're using them :expressionless: Therefore I can't see any reason to prefer a longer ```data-namespace-*``` format over ```namespace-*```.

chrisprice:
Gah! Read through that final SO link and it goes on to suggest custom xml namespaces...

http://www.w3.org/TR/SVG/extend.html#ForeignNamespaces

ColinEberhardt:
I love the part in the discussion that states data- attributes are included alongside ng- to 

> appease dorks and complainers

Perhaps we should use layout-css but also have namespaced attributes with a proper DTD for the dorks 😉

visual-tests/test-scripts/layout.js
```diff
@@ -0,0 +1,58 @@
+(function(d3, fc) {
```
jleft: I think this file should be in `visual-tests/test-scripts/utilities/`?

jleft:
I had a play with this yesterday - I really like it :smiley: I love the flexibility it provides (especially over chartLayout!).

visual-tests/test-scripts/layout.js
ColinEberhardt: Thanks for spotting this - that file should actually be removed.

ColinEberhardt:
@chrisprice I think you are right, this could benefit from a thin adapter layer that makes the programmatic construction of SVG a little cleaner. I've raised an issue #274

PR #268, opened 2015-01-30
Split indicator algorithms from rendering to enable re-use

### LINKED ISSUE #239: Split indicators into model and view components

chrisprice:
Fixes #239


chrisprice:
FYI the rsi algorithm might be a bit sketchy, we need some real data to test this with.

chrisprice:
Also need to change the RSI to use annotations rather than lines now that they've landed in master.

chrisprice:
> Also need to change the RSI to use annotations rather than lines now that they've landed in master.

Done

ColinEberhardt:
The RSI looks good to me - I guess this is the exponential moving average flavour, rather than the SMA.

We've actually got a book in the office which has test data for a range of studies.

PR #269, opened 2015-01-31
Remove valueAccessor to allow components to handle NaN as appropriate

### LINKED ISSUE #263: Remove valueAccessor
As far as I can tell, functionality like this is better handled by the series themselves using something like ```defined()```.

chrisprice:
Fixes #263 (highly likely to conflict with other pending PRs)


chrisprice:
Updated to reflect the latest changes on master. Should be good to go.

jleft:
Looks good :+1:

PR #281, opened 2015-02-04
Add D3 as an optional dependency

### LINKED ISSUE #280: D3 node module dependencies
#262 added D3 to the package.json - this brings in contextify (via JSDOM) as a dependency. Building this with node-gyp requires Python, Make and a C/C++ compiler on unix, and Visual Studio C++ on Windows.

Installing contextify with npm fails for me on OS X 10.10.2 with Python and the XCode command line tools installed:

> node-gyp rebuild
make: *** No rule to make target `../../../../../../../../../Volumes/Macintosh', needed by `Makefile'.  Stop.
gyp ERR! build error
...
gyp ERR! not ok 

It seems we're only doing this to have d3.js in the visual tests. Could we instead get it from http://d3js.org and download if needing it offline?

jleft:
Resolves #280

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -130,7 +130,10 @@ module.exports = function (grunt) {
         jasmine: {
             options: {
                 specs: '<%= meta.testJsFiles %>',
-                vendor: 'node_modules/d3/d3.js'
+                vendor: [
+                    'node_modules/d3/d3.js',
+                    'http://d3js.org/d3.v3.js'
+                ]
             },
             test: {
                 src: ['<%= meta.componentsJsFiles %>'],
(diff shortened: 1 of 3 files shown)


chrisprice:
Just checking if you've tested this offline?

jleft:
Yes, I tried it successfully; as long as D3 is installed in node_modules, this should work offline.

When online, I couldn't see any issues with including D3 twice in the visual tests or Jasmine - are there any downsides to this?

chrisprice:
It's certainly not best practice, but this is only a temporary measure so
if you've not seen any then let's assume it'll be fine for now.

PR #282, opened 2015-02-05
Standardise SVG element structure, upgrade gridlines + misc

### LINKED ISSUE #255: Standardise the SVG element structure

chrisprice:
I also deleted the README.md files as they're out of date.

Fixes #255, #257, #232.


chrisprice:
Also #139

components/fc.css
```diff
@@ -18,3 +18,13 @@
 .axis text { 
   font: 10px sans-serif; 
 }
+
+.gridlines { 
+  stroke: #ccc;
+  stroke-width: 1.0;
+}
+
+.chartArea rect.background {
```
ColinEberhardt: I guess whether we keep this or not depends on the future of chartLayout? Still worth merging this PR all the same

components/scale/linear.js
```diff
@@ -1,72 +0,0 @@
-(function(d3, fc) {
```
ColinEberhardt: cool - I was intending to delete this scale ...

ColinEberhardt:
Thanks @chrisprice - it's a :+1: from me.

Issue #241, opened 2015-01-13
fc.scales.dateTime with hideWeekends(true) adds excessive padding


chrisprice:
See the crosshairs examples in PR #207.

chrisprice:
@ColinEberhardt just went to check if this was still an issue and it looks like there's a random label floating to the left of the axis now. Also there's no padding what-so-ever... think something's gone wrong...

ColinEberhardt:
Thanks for checking, I've self-assigned, and will take a look

ColinEberhardt:
Found the issue, the values returned by the `ticks` function may fall within discontinuities. I'll sort this out ...

ColinEberhardt:
The same is probably true of `nice`, which can change the domain.

chrisprice:
PR replaces issue

PR #283, opened 2015-02-08
dateTime ticks and nice function are now clamped to remove discontinuiti...

ColinEberhardt:
...es

Fixeds #241

### DIFF
--- a/components/scale/dateTime.js
+++ b/components/scale/dateTime.js
@@ -5,6 +5,26 @@
         return dateTimeScale();
     };
 
+    // obtains the ticks from the given scale, transforming the result to ensure
+    // it does not include any discontinuities
+    fc.scale.dateTime.tickTransformer = function(ticks, discontinuityProvider, domain) {
+        var clampedTicks = ticks.map(function(tick, index) {
+            if (index < ticks.length - 1) {
+                return discontinuityProvider.clampUp(tick);
+            } else {
+                var clampedTick = discontinuityProvider.clampUp(tick);
+                return clampedTick < domain[1] ?
+                    clampedTick : discontinuityProvider.clampDown(tick);
+            }
+        });
+        var uniqueTicks = clampedTicks.reduce(function(arr, tick) {
+            if (arr.filter(function(f) { return f.getTime() === tick.getTime(); }).length === 0) {
+                arr.push(tick);
+            }
+            return arr;
+        }, []);
+        return uniqueTicks;
+    };
 
     /**
     * The `fc.scale.dateTime` scale renders a discontinuous date time scale, i.e. a time scale that incorporates gaps.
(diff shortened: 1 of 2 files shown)


components/scale/dateTime.js
```diff
@@ -60,6 +60,35 @@
...
+            var clampedTicks = ticks.map(function(tick, index) {
+                if (index < ticks.length - 1) {
+                    return discontinuities().clampUp(tick);
+                } else {
+                    var clampedTick = discontinuities().clampUp(tick);
+                    return clampedTick < scale.domain()[1] ?
+                        clampedTick : discontinuities().clampDown(tick);
+                }
+            });
+            var uniqueTicks = clampedTicks.reduce(function(arr, tick) {
```
chrisprice: Might be worth swapping this out for [d3.set([...]).values()](https://github.com/mbostock/d3/wiki/Arrays#set_values)

components/scale/dateTime.js
```diff
@@ -60,6 +60,35 @@
...
+            var domainLower = discontinuities().clampUp(domain[0]);
+            var domainUpper = discontinuities().clampDown(domain[1]);
+            adaptedScale.domain([domainLower, domainUpper]);
+            return scale;
+        };
+
+        scale.ticks = function() {
+            var ticks = adaptedScale.ticks();
+            var clampedTicks = ticks.map(function(tick, index) {
+                if (index < ticks.length - 1) {
```
chrisprice: Do you need the last item check or could you apply the else block to all ticks?

tests/scale/dateTimeSpec.js
```diff
@@ -21,6 +21,27 @@
             expect(clone.domain()[1]).toEqual(end);
         });
 
+        describe('tick calculations', function() {
```
chrisprice: Is it worth adding a test for nice as well?

chrisprice:
Looks good, glad it was pretty simple to fix!

components/scale/dateTime.js
ColinEberhardt: Ah, forgot to check the d3 utility functions!

components/scale/dateTime.js
ColinEberhardt: No, you have to clamp the sides inwards, otherwise you risk clamping  the upper or lower ticks outside of the axis range.

tests/scale/dateTimeSpec.js
ColinEberhardt: Probably 😉

Although, this sort of behaviour is not easily black-box tested, need to find a suitable way to inject values to test the clamping behaviour

components/scale/dateTime.js
ColinEberhardt: Actually, doesn't work so well in this context, would require a map( d.getTime() ) before using d3.set, and an inverse mapping afterwards.

tests/scale/dateTimeSpec.js
```diff
@@ -21,6 +21,69 @@
...
+
+            it('should clamp both ends inwards', function() {
+                var start = new Date(2015, 0, 4); // sunday
+                var end = new Date(2015, 0, 11); // sunday
+
+                var inputTicks = [start, end];
+
+                var filtered = fc.scale.dateTime.tickTransformer(inputTicks,
+                    fc.scale.discontinuity.skipWeekends(), [start, end]);
+                console.log(filtered);
```
chrisprice: Console.log?

ColinEberhardt: Probably not needed!

ColinEberhardt:
@chrisprice I'll remove the `console.log` then I think this one is good to merge. `nice` could benefit from testing, but is low risk (and I doubt I'll get round to it!)

chrisprice:
Sounds good. I had a go at simplifying the code myself and couldn't, I
think it's just one of those blocks which is deceptively complicated. :)

jleft:
:+1: Looks good!

> `nice` could benefit from testing, but is low risk (and I doubt I'll get round to it!)

I'm happy to attempt testing this - after my recent training, I could with the practice!

Issue #274, opened 2015-02-02
Improve layout interface for programmatic construction of charts 


ColinEberhardt:
see https://github.com/ScottLogic/d3-financial-components/pull/265#issuecomment-72372148

chrisprice:
Other issues I've faced with layout -
* Some elements (e.g. ```rect``` and ```svg```) have more appropriate attributes for the x/y/width/height values (rather than ```layout-width/height``` and ```transform=translate...```.
* I'd like to be able to call layout on a subset of previously laid-out elements (e.g. to reflow a small part of the layout). This requires a way to pass the previously calculated width and height into the layout somehow.
* The layout is trashing other transform values (e.g. rotating text labels).

Not sure if they all need to be solved by layout but thought I'd mention them here.

chrisprice:
Another one -

* Currently accessing a resulting width/height returns a string value (it's been round-tripped via an attribute), it would be better if it was forced to a number (otherwise feeding it into a scale range gives string results...)

ColinEberhardt:
I've come to the uncomfortable realisation that this probably needs unit tests  :-/

ColinEberhardt:
> The layout is trashing other transform values (e.g. rotating text labels).

This is a tricky one to fix. The layout could just add its transforms to any that are already present, however - we want our components to be idempotent, so if invoked multiple times it would need to know which ones had been added via layout and which were already there.

chrisprice:
I suppose setting x/y on text nodes would solve the particular problem I'm having?

jleft:
More thoughts here: https://github.com/ScottLogic/d3-financial-components/pull/285#issuecomment-73720583

chrisprice:
fb5d87ea added support for width/height/x/y on rect/svg elements.

PR #285, opened 2015-02-09
Added a layout function to d3.selection.

ColinEberhardt:
See #274 

Thanks for the ideas @chrisprice - I added a more complex example, and found the `layout` function to be really quite versatile and easy to use.


chrisprice:
On mobile so can't comment inline. 

* Consider changing type checks to be positive rather than negative with errors thrown? (Here's looking at you d3!!)
* Can we special case a coercion to number for reading back width/height values? That's just caught me out in a horrible way!
** Just thought about x/y, how can we help with those?

chrisprice:
What about rect/svg or are you leaving that to me?

ColinEberhardt:
> Consider changing type checks to be positive rather than negative with errors thrown? (Here's looking at you d3!!)

I'm all for better error reporting with D3FC (and D3) - although we do need to decide on what mechanism to use for reporting errors.

> What about rect/svg or are you leaving that to me?

No, I'm happy to do that. Although I'd prefer to tackle things via multiple PRs rather than one big one, so just keep the issue open rather than closing it.

jleft:
Looks good :+1: 

Conflicts resolved and merged in #294.

Issue #288, opened 2015-02-10
Restructure Visual Tests


jleft:
With the current number of examples, the visual test page is fairly unwieldy and becoming increasing sluggish.

Some tests are no longer relevant or necessary; these should be removed. I'll also look into a more flexible approach for managing these tests.

chrisprice:
I believe this is now fixed, please re-open if you disagree.

PR #289, opened 2015-02-12
Visual Tests restructure

jleft:
The visual tests have been restructured (#288). This isn't ready to merge, but I thought it'd be worth sharing at this stage - any feedback is welcome!

I'd recommend [viewing here](http://jleft.github.io/d3-financial-components/), or checking out this branch and running `npm install && grunt visual-tests`

new-tests

---

[Assemble](http://assemble.io/) is used to generate a set of pages, comprising of:
+ A page for each test
+ An index page which lists all tests and links to each test's own page

Each test requires a [handlebars](http://handlebarsjs.com/) template, as shown in [visual-tests/src/test-fixtures](https://github.com/jleft/d3-financial-components/tree/visual-tests-restructure/visual-tests/src/test-fixtures). The templates are built into HTML by the build process. Any other files in test-fixtures are copied to dist and the file hierarchy is unchanged, so additional files can be added and linked to (as shown in the current examples).

N.B. I've haven't moved many tests over yet, but it's easy to do.

**New structure:**
```
+-- visual-tests
|  +-- dist // contains the generated files to view in your browser (this should be ignored by git)
|  +-- src // edit files in here to change the output
|  |  +-- test-fixtures // tests go here; there are a few examples, hopefully fairly self-explanatory!
|  |  +-- site // contains the various layouts, includes and other files to build the site around
```


jleft:
Updated; preview available here: http://jleft.github.io/d3-financial-components/

I'm hoping this is a solid foundation for the visual tests - we could add further features in the future, if we think they'd be useful - search, tag filtering, category pages etc.

Gruntfile.js
```diff
@@ -147,18 +244,23 @@ module.exports = function (grunt) {
...
 
         clean: {
-            doc: ['doc']
+            doc: ['doc'],
+            visualTests: ['visual-tests/dist']
         }
     });
 
     require('matchdep').filterDev('grunt-*').forEach(grunt.loadNpmTasks);
+    grunt.loadNpmTasks('assemble');
```
chrisprice: Is this needed? Can't see it here https://www.npmjs.com/package/grunt-assemble

Gruntfile.js
```diff
@@ -34,6 +66,62 @@ module.exports = function (grunt) {
...
+            visualTests: {
+                options: {
+                    sourceMap: true
+                },
+                src: 'visual-tests/src/site/assets/js/**/*.js',
+                dest: 'visual-tests/dist/assets/index.js',
+            }
+        },
+
+        copy: {
```
chrisprice: Is there any way we can reduce the amount of copying going on?

chrisprice:
LGTM other than the odd comment, thanks James!

Gruntfile.js
jleft: Yes, as we're actually using _assemble_, rather than _grunt-assemble_:

https://www.npmjs.com/package/assemble

Might be worth using [load-grunt-tasks](https://github.com/sindresorhus/load-grunt-tasks) to make this a bit tidier?

chrisprice: I can't spot any difference although my phone is having a nightmare rendering the npm site...

chrisprice: Apparently grunt-assemble is something new and untested... Let's just leave it as is for now :)

Gruntfile.js
jleft: It would be possible to link directly to the assets in `/node_modules` and `/dist`, rather than copying them, but generating the links is trickier.

Copying the files doesn't take long, the files are fairly small (< 2MB) and it makes linking to these files significant easier. I also like that the generated directory, `/visual-tests/dist/`, has no dependencies above itself it the directory structure, so it can be served as the web root.

Happy to change this though, if you think it's worth it?

jleft:
> LGTM other than the odd comment, thanks James!

No problem, thanks for looking over this :)

Gruntfile.js
chrisprice: Ok sounds like it'll be snappy then, we can always revisit if we think it's getting slow.

PR #290, opened 2015-02-12
Simplify dataGenerator

tlsim:
- Replaces the internal normal distribution sampling with `d3.random.normal()`. This removes the random seed functionality - I think this would be better handled externally if required by https://github.com/davidbau/seedrandom for example. The visual tests work fine without it. Any opinions? 

- Possibly fixes an infrequently encountered bug where the price data returned explodes:
datageneratorbug

- Changes the filtering behaviour. Currently, it tries to return the requested number of points no matter what filter is set. This is a problem as it changes the units of the 'drift' and 'volatility' when filtering is applied, and it's possible to set a filter that results in an infinite loop. I've changed this to apply the filter on the generated array as the final step.

Should filtering be removed entirely or is the default filter useful?


ColinEberhardt:
Awesome - thanks for this @tlsim - I'll take a better look at this tomorrow,

components/math/randomWalk.js
```diff
@@ -0,0 +1,26 @@
+(function(d3, fc) {
+    'use strict';
+
+    fc.math.randomWalk = function(years, steps, mu, sigma, initial) {
+        var randomNormal = d3.random.normal(),
+            timeStep = years / steps,
```
jleft: Years?

jleft:
Looks like a vast improvement - thank you! Great to see the tests too :+1: 

When the conflicts are resolved, this is good to merge.

tlsim:
Thanks @jleft

Issue #258, opened 2015-01-19
Translate data point groups in series


chrisprice:
Currently the elements within the groups for series data points are individually positioned to the correct x/y offsets. The groups themselves have no translation applied.

Instead the groups should be translated and the elements given relative offsets.

chrisprice:
The [decorate visual test](http://localhost:8000/functional/decorate.html)(grunt dev:serve) suggests this is now implemented. 

If anyone disagrees feel free to re-open.

PR #291, opened 2015-02-13
Bar series update (translate data point groups in series)

jleft:
See #258

### DIFF
--- a/components/series/bar.js
+++ b/components/series/bar.js
@@ -7,27 +7,29 @@
         var x = function(d) { return bar.xScale.value(bar.xValue.value(d)); };
         var y = function(d) { return bar.yScale.value(bar.yValue.value(d)); };
 
-        var bar = function(selection) {
-            selection.each(function(data) {
+        var bar = function(container) {
+            container.each(function(data) {
                 var container = d3.select(this);
                 var series = fc.utilities.simpleDataJoin(container, 'bar', data, bar.xValue.value);
 
+                var width = bar.barWidth.value(data.map(x));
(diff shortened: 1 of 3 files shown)


chrisprice:
I know we talked about translating to the most appropriate location for the series i.e. where would it be most helpful to have the origin for transitions. Have you specifically chosen the location in this case?

There's a few too/to typos, not actually yours but whilst you're correcting grammar :)

jleft:
As discussed, the origin of each group is now translated to the centre (horizontally and vertically) of its corresponding bar (changed in 61945c7); illustrated by the white crosses:

bar-origin

I'd anticipate that most translations from this point will take into account the height and/or width or the bar. When I've tried adding some elements using decorate, this it felt a bit awkward, for example, (unless I'm missing something!) to translate a text label to the top of a bar, you'd have to perform a y-translation like this: `-(yScale(0) - yScale(d.age)) / 2`. I wonder if there is a more elegant solution, maybe exposing a couple of functions for the height and width of each bar would be useful?

ColinEberhardt:
@jleft - #373 has now added the correct translation, and has a test which shows the 'origin' of each datapoint.

I still need to add transition support back in!

jleft:
> @jleft - #373 has now added the correct translation, and has a test which shows the 'origin' of each datapoint.

Looks really good! I'll close this PR now.

> I still need to add transition support back in!

Cool, I'm happy to look into/help with this, if that would be useful?

ColinEberhardt:
> > I still need to add transition support back in!
> Cool, I'm happy to look into/help with this, if that would be useful?

That would be great - thanks. Hopefully now that the series are all rendered in a similar fashion it should be easy to add transition support to all of them.

PR #292, opened 2015-02-13
Update Candlestick series to use path

jleft:
Each candle is now drawn using a single `path`, rather than a `rect` and a `line`. This feels like the more idiomatic solution to me, and will lead to much more intuitive transitions.

### DIFF
--- a/components/series/candlestick.js
+++ b/components/series/candlestick.js
@@ -18,9 +18,8 @@
 
                 var g = fc.utilities.simpleDataJoin(container, 'candlestick', data, candlestick.xValue.value);
 
-                var enter = g.enter();
-                enter.append('line');
-                enter.append('rect');
+                g.enter()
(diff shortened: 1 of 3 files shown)


tlsim:
:+1:

chrisprice:
Thanks @jleft!

chrisprice:
Need to learn how to use GitHub... 

Looks like there are some merge conflicts, any chance you could take a quick look?

jleft:
No problem, will do!

jleft:
Having looked at this again, the wick doesn't look great with transitions, so may need a small tweak - I'll look into this.

I've also done some work on a nicer API for low-level path data, which should make to code more readable - I'll try to get this in soon.

chrisprice:
Merging as is for now, we can work on the transition support separately.

Issue #275, opened 2015-02-02
Proposal for chained property syntax


chrisprice:
In #232 we introduced a new convention for the property accessors. Having spent some time with it, I really don't believe this new approach enhances readability (any clarity in defining the property is offset by the verboseness of every access). Therefore I propose we adopt one of the following syntaxes -

# d3

## Plain
```js
d3.layout.histogram = function() {
  var frequency = true,
      valuer = Number,
      ranger = d3_layout_histogramRange,
      binner = d3_layout_histogramBinSturges;
  
  // ...

  histogram.value = function(x) {
    if (!arguments.length) return valuer;
    valuer = x;
    return histogram;
  };
```

## Functor
```js
d3.layout.histogram = function() {
  var frequency = true,
      valuer = Number,
      ranger = d3_layout_histogramRange,
      binner = d3_layout_histogramBinSturges;
  
  // ...

  histogram.range = function(x) {
    if (!arguments.length) return ranger;
    ranger = d3_functor(x);
    return histogram;
  };
```

# Alternative terse syntax 

## Plain
```js
d3.layout.histogram = function() {
  var frequency = true,
      valuer = Number,
      ranger = d3_layout_histogramRange,
      binner = d3_layout_histogramBinSturges;
  
  // ...

  histogram.value = function(x) {
    return arguments.length ? (valuer = x, histogram) : valuer;
  };
```

## Functor
```js
d3.layout.histogram = function() {
  var frequency = true,
      valuer = Number,
      ranger = d3_layout_histogramRange,
      binner = d3_layout_histogramBinSturges;
  
  // ...

  histogram.range = function(x) {
    return arguments.length ? (ranger = d3_functor(x), histogram) : ranger;
  };
```

chrisprice:
I forgot to mention the big win of these approaches, side effects can now easily be incorporated into the setter rather than being applied as part of the called function e.g. -

## Old
```js
       var bollingerBands = function(selection) {

            algorithm.inputValue(bollingerBands.yValue.value)
                .outputValue(bollingerBands.writeCalculatedValue.value);

            area.xScale(bollingerBands.xScale.value)
                .yScale(bollingerBands.yScale.value)
                .xValue(bollingerBands.xValue.value);

            upperLine.xScale(bollingerBands.xScale.value)
                .yScale(bollingerBands.yScale.value)
                .xValue(bollingerBands.xValue.value);

            averageLine.xScale(bollingerBands.xScale.value)
[...]

PR #299, opened 2015-04-09
Fixes #284 - Stacked bar series now uses stacked layout.

### LINKED ISSUE #284: Check d3 stackedlayout to see if we should be using it for stackedbar
Stumbled across it in the API reference, no idea how relevant it is.

janakerman:
I settled on an approach that exposed the `stackedBar`s internal `stack` layout. This allows users who want to show a simple bar with a fixed range simple feed the data to their `stackedBar`. It also allows for more advanced use cases:
- Dynamic Y domain based on the data (as in the `stacked-bar` visual test).
- Modification of the how data is pulled from the selection using the `stack` layout's accessors.
- Modification of the stack's baseline algorithm using the `stack` layout's offset function.
- Modification of the bar order using the `stack` layout's `order` function .

Eager to hear any feedback!


ColinEberhardt:
Thanks for this - I'll take a look shortly :-)

chrisprice:
Thanks Jan. I'm keen to review this as well but I'm not going to get a chance until over the weekend.

janakerman:
There's no rush at all - let me know what you think. I'm happy to discuss my original thoughts or any suggestions anyone might have.

components/series/stackedBar.js
```diff
@@ -6,77 +6,39 @@
...
 
                 // add a 'root' g element on the first enter selection. This ensures
                 // that it is just added once
                 container = d3.select(this);
 
-                var keyFunction = function(d) {
-                    return d[stackedBar.xValueKey.value];
-                };
-                var g = fc.utilities.simpleDataJoin(container, 'stacked-bar', data, keyFunction);
+                // If the user has already set stacked data then we don't need to re-compute it.
```
chrisprice: Is this purely a performance optimisation? If so I think we should remove it, if not we might need to revisit how it's done.

By caching a selection's data with the component, it prevents the same component being re-used with another selection. N.B. we're not actually sure if this is a good design decision yet, but it is one we've stuck to where at all possible. 

Also as a general rule we've opted to leave out performance optimisations at this stage to. The intention is to revisit specific areas with poor performance once we have the first release out.

components/series/stackedBar.js
```diff
@@ -6,77 +6,39 @@
...
+                var layers = stackedBar.stackedData.value;
+                if (typeof layers === 'undefined' || layers.length === 0) {
```
chrisprice: JS stylistic things... 
* Prefer ```!= null``` to anything involving undefined. Generally ```null``` and ```undefined``` are conflated as being the same value, using the bang equals null approach will match either (notice the use of double rather than triple equality). An exception to the rule can be for testing for the existence of arguments but testing the length of the ```arguments``` object is normally more appropriate.
* Either use an empty array to represent empty and operate on it as if it were the real thing, or choose null to represent empty and special case it. I would avoid using an empty array as a special case.

components/series/stackedBar.js
```diff
@@ -6,77 +6,39 @@
...
+                }
+
+                var g = fc.utilities.simpleDataJoin(container, 'stacked-bar', layers);
 
+                stackedBar.decorate.value(g);
 
-                // create a join for each bar
                 var bar = g.selectAll('rect')
-                    .data(function(d) { return objectDatapointToArray(d); })
+                    .data(function(d) { return d; })
```
chrisprice: Prefer ```fc.utilities.fn.identity``` rather than creating another new function.

components/series/stackedBar.js
```diff
@@ -6,77 +6,39 @@
...
                     .enter()
                     .append('rect');
 
-                // compute the bar width from the x values
-                var xValues = data.map(function(d) {
-                    return stackedBar.xScale.value(d[stackedBar.xValueKey.value]);
-                });
+                // Compute the bar width from the x values
```
chrisprice: If I'm reading the code correctly, it is no longer agnostic of the input data format i.e. the user must now use ```x```, ```y``` & ```y0``` properties. In other places we've used accessor functions, would that work here?

components/series/stackedBar.js
```diff
@@ -6,77 +6,39 @@
...
+                // Assumes first series contains all possible X values.
```
chrisprice: Any reason to have this restriction?

components/series/stackedBar.js
```diff
@@ -88,7 +50,9 @@
 
         stackedBar.yScale = fc.utilities.property(d3.scale.linear());
 
-        stackedBar.xValueKey = fc.utilities.property('name');
+        stackedBar.layout = fc.utilities.property(d3.layout.stack().offset('zero'));
+
+        stackedBar.stackedData = fc.utilities.property([]);
```
chrisprice: See above, I don't think it's possible to expose this at the same time as allow reuse with multiple selections.

visual-tests/test-scripts/series/stacked-bar.js
```diff
@@ -9,17 +9,38 @@
...
+                    });
                 }
             }
-            return total;
         });
 
+        var stack = fc.series.stackedBar();
+
+        // Manually compute stacked data - use this to calculate the max Y domain.
+        stack.stackedData(stack.layout.value(seriesData));
```
chrisprice: Oooohhhhhhh, nice one. This is a good example of why you should get other people to try out your code! It hadn't occurred to me that this would be a consequence of our approach to properties, we should definitely revisit this separately (#275).

Our normal pattern is to use the d3 accessor i.e. ```stack.layout()(seriesData)```.

visual-tests/test-scripts/series/stacked-bar.js
```diff
@@ -9,17 +9,38 @@
 
     d3.csv('stackedBarData.csv', function(error, data) {
 
-        var categories = data.slice(1).map(function(d) { return d.State; });
-        var yMax = d3.max(data, function(d) {
-            var total = 0;
-            for (var prop in d) {
-                if (prop !== 'State') {
-                    total += Number(d[prop]);
+        var categories = data.map(function(d) { return d.State; });
```
chrisprice: I would be tempted to split this into something which [reduced](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce) the array to be grouped by layer (in this case age I think) and then use a [values accessor](https://github.com/mbostock/d3/wiki/Stack-Layout#values) on the stack layout.

visual-tests/test-scripts/series/stacked-bar.js
```diff
@@ -43,18 +65,18 @@
...
+                sel.attr('fill', function(d, i) {
+                    return color(i);
+                });
+            });
 
         chartLayout.getPlotArea(chart)
           .append('g')
           .attr('class', 'series')
-          .datum(data.slice(1))
+          .datum(seriesData)
```
chrisprice: Couldn't work out why your code was doing something different until I spotted @ColinEberhardt's glorious hack :)

visual-tests/test-scripts/tools/measure.js
```diff
@@ -70,7 +70,6 @@
         .xLabel(function(d) {
             return !(d.source && d.target) ? '' :
                 d3.time.day.range(d.source.datum.date, d.target.datum.date).length + ' days';
-                // d3.time.format('%a, %e %b')(d.target.datum.date - d.source.datum.date);
```
chrisprice: Good spot.

chrisprice:
Thanks again @janakerman. I think it's a good start and the flexibility of the d3 layout which you call out is definitely preferable to rolling our own solution. 

As a general point, I would prefer to see us rebind the useful properties of the sub-components (i.e. in this case layout) rather than exposing them directly. I've tried to [explain my rationale on the wiki](https://github.com/ScottLogic/d3-financial-components/wiki) because I think it's important that not only do we share a common design approach but any future users understand our reasoning.

components/series/stackedBar.js
```diff
@@ -6,77 +6,39 @@
...
-                    }
-                }
-                return values;
-            }
-
             selection.each(function(data) {
 
                 // add a 'root' g element on the first enter selection. This ensures
                 // that it is just added once
                 container = d3.select(this);
```
jleft: Very minor: unless there's a reason not to, for consistency with other components, `container` should be declared here, rather than before the `selection.each`.

components/series/stackedBar.js
jleft: FWIW - in my experience with this problem, caching directly on the element - instead of the component or selection - seemed to work relatively well.

components/series/stackedBar.js
```diff
@@ -6,77 +6,39 @@
...
-                var g = fc.utilities.simpleDataJoin(container, 'stacked-bar', data, keyFunction);
+                // If the user has already set stacked data then we don't need to re-compute it.
+                var layers = stackedBar.stackedData.value;
+                if (typeof layers === 'undefined' || layers.length === 0) {
+                    layers = stackedBar.layout.value(data);
+                }
+
+                var g = fc.utilities.simpleDataJoin(container, 'stacked-bar', layers);
 
+                stackedBar.decorate.value(g);
```
jleft: I think decorate should be moved to the end of this `selection.each` - as it is, attributes set in decorate could be overridden by the component.

Making this change this would be consistent with the other components and any manipulations which are made in decorate will take precedence over those set by the component - which I think is the right way to do it.

components/series/stackedBar.js
jleft: Possibly not important at the moment, but I don't think this will work with transitions, as `.data()` isn't defined on a transition.

components/series/stackedBar.js
```diff
@@ -88,7 +50,9 @@
 
         stackedBar.yScale = fc.utilities.property(d3.scale.linear());
 
-        stackedBar.xValueKey = fc.utilities.property('name');
+        stackedBar.layout = fc.utilities.property(d3.layout.stack().offset('zero'));
```
jleft: I think `'zero'` is the default offset, does it need to be explicitly set here?

visual-tests/test-scripts/series/stacked-bar.js
```diff
@@ -9,17 +9,38 @@
...
 
-        var categories = data.slice(1).map(function(d) { return d.State; });
-        var yMax = d3.max(data, function(d) {
-            var total = 0;
-            for (var prop in d) {
-                if (prop !== 'State') {
-                    total += Number(d[prop]);
+        var categories = data.map(function(d) { return d.State; });
+
+        // Format data into a map of series key to series data & a data array to pass to our stacked-bar series.
```
jleft: I wonder if it might be worth creating a utility from this? I.E. something to convert a typical dataset, which is an array of objects (like that created by our data generator), into the format used by the stack layout:
```js
var dataset = [
    [
        { x: 0, y: 5 },
        //...
    ],
    //...
];
```

components/series/stackedBar.js
chrisprice: Be sure to consider how to manage cache invalidation of you do cache on the
element.

janakerman: Yeah, this was purely a performance optimisation. I normally side against premature optimisation, but this situation felt a little weird. It was like the visual test's use case was repeating most of the benefit that the stacked layout actually gives. 

Seeing the use case whereby a chart's domain is automatically scaled seemed fairly common to me, it tilted me towards trying to optimise it.

@chrisprice - what's the final verdict, would you like this removed?

components/series/stackedBar.js
janakerman: Thanks for the tips, I can see the reasoning there. I'll update.

components/series/stackedBar.js
janakerman: Yes, it would. We should be using the accessor functions defined the stacked layout, as it has accessor functions defined on it for the `x` & `y` values. However, it doesn't define an accessor function for `y0`, so that's something we'd need to define on the series itself. 

Since the stack layout also has a setter function for `y0`, it's worth noting that there would be an implicit link between `y0`'s setter and getter and user's would have to remember to update both. This would probably be nicer once I've rebound the methods (as you suggested) as both the getter and the setter would be accessible on the series.

components/series/stackedBar.js
janakerman: No reason other than simplicity. I could create a set of x values from all of the provided arrays. In hindsight, this probably wouldn't be much more difficult. I will update!

components/series/stackedBar.js
janakerman: Yeah, I was trying to be explicit. I'll update this to remove it.

visual-tests/test-scripts/series/stacked-bar.js
janakerman: I'll update it to `stack.layout()(seriesData)`.

visual-tests/test-scripts/series/stacked-bar.js
janakerman: Yeah. I couldn't work out what the original reason for that was!

visual-tests/test-scripts/series/stacked-bar.js
janakerman: I'll look into that - thanks for the suggestion.

components/series/stackedBar.js
janakerman: As discussed, we'll remove this. We shouldn't try to predict how the user will want to calculate their min/max Y domain. The use case of the user using the stack layout to do this is probably an edge case and we shouldn't try to optimise it.

components/series/stackedBar.js
janakerman: I haven't had a chance to fully look into this but I've seen your suggestion and I'll take a look as soon as I get a chance.

visual-tests/test-scripts/series/stacked-bar.js
janakerman: Still haven't had that much experience with higher order functions yet so their use cases don't instantly jump out at me! I'll take a look at this and update it.

janakerman:
@chrisprice @jleft - I think this is ready for another review.

One thing that's worth noting is that the stack layout actually modifies the original data set. This could be useful when configuring a crosshair for the stacked columns. However, it did occur to me that it the fact that it modifies the original data might be an issue.

components/series/stackedBar.js
```diff
@@ -88,8 +52,12 @@
...
         stackedBar.yScale = fc.utilities.property(d3.scale.linear());
 
-        stackedBar.xValueKey = fc.utilities.property('name');
+        // Implicitly dependant on the implementation of the stack layout's `out`.
+        stackedBar.y0 = fc.utilities.property(function(d) {
+            return d.y0;
+        });
 
-        return stackedBar;
+        var stackLayout = d3.layout.stack();
```
chrisprice: FYI, it's considered a bad practice to have a variable declared after it's first referenced. It can cause confusing bugs and is counter-intuitive to where you'd expect a declaration to be when you first look at a code block. 

It's working because JavaScript will implicitly hoist the declaration, but not the assignment, to the top of the enclosing function.

components/series/stackedBar.js
```diff
@@ -4,79 +4,43 @@
...
-                };
-                var g = fc.utilities.simpleDataJoin(container, 'stacked-bar', data, keyFunction);
+                var layers = stackLayout(data);
 
+                var g = fc.utilities.simpleDataJoin(container, 'stacked-bar', layers);
 
-                // create a join for each bar
                 var bar = g.selectAll('rect')
-                    .data(function(d) { return objectDatapointToArray(d); })
+                    .data(function(d) { return stackLayout.values()(d); })
```
chrisprice: Sorry, I wasn't clear in my previous feedback and this yet again points to why we should remove this pattern.

Within the component declaring a property, our convention has been to reference the value using the ```.value``` member. Outside of the component, references are made by invoking the function with zero-args. 

So in this case it would be ```data(function(d) { return stackLayout.values.value(d); })```, which would then simplify to ```data(stackLayout.values.value)```.

components/series/stackedBar.js
```diff
@@ -4,79 +4,43 @@
...
                     .enter()
                     .append('rect');
 
-                // compute the bar width from the x values
-                var xValues = data.map(function(d) {
-                    return stackedBar.xScale.value(d[stackedBar.xValueKey.value]);
-                });
-                var width = stackedBar.barWidth.value(xValues);
+                var xValues = stackedBar.xScale().domain();
+                var xPositions = xValues.map(function(d) { return stackedBar.xScale()(d); });
```
chrisprice: As above, I'd also be tempted to chain the method calls from the previous line.

components/utilities/object.js
```diff
@@ -0,0 +1,21 @@
+(function(d3, fc) {
```
chrisprice: Any reason not to use [Object.Keys](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys)?

visual-tests/src/test-fixtures/series/stacked-bar.js
```diff
@@ -9,26 +9,48 @@
...
-        var yMax = d3.max(data, function(d) {
-            var total = 0;
-            for (var prop in d) {
-                if (prop !== 'State') {
-                    total += Number(d[prop]);
-                }
-            }
-            return total;
+        // Collect all series keys.
+        var seriesKeys = data.reduce(function(previous, current) {
```
chrisprice: I would have expected the data array to objects with the same keys for each item, is it possible for the keys to be different? If not then I think most of this code is not required.

components/utilities/object.js
janakerman: Yeah, the README says that the components are Firefox 3+ and Safari 3.2, but I don't think Object.Keys was supported until version 4 and 5 respectively.

components/series/stackedBar.js
janakerman: I can see how this could be confusing so I'll change that. I was aware of variable hoisting but I didn't consider that the assignment wasn't - that's good to know.

components/utilities/object.js
chrisprice: Another good spot, I didn't even realise we had committed to a particular
set of browsers. Can you raise an issue to revisit the supported browser
list and for now it's fine to assume Object.keys support?

janakerman: #302

components/series/stackedBar.js
janakerman: I think this is only possible with your comment on line 24. Here, `stackLayout` is the d3 component not our custom component, `stackedBar`. I don't think the d3 accessors define a `.value` member.

chrisprice: Fair point, I hadn't spotted that! Looks like there are a few occurrences in this file though e.g. lines 26, 29, etc..

visual-tests/src/test-fixtures/series/stacked-bar.js
```diff
@@ -9,26 +9,42 @@
...
-        var categories = data.slice(1).map(function(d) { return d.State; });
-        var yMax = d3.max(data, function(d) {
-            var total = 0;
-            for (var prop in d) {
-                if (prop !== 'State') {
-                    total += Number(d[prop]);
-                }
-            }
-            return total;
+        /*  Build series objects for each series in the data set.
```
chrisprice: This is much clearer, nice one.

components/series/stackedBar.js
```diff
@@ -3,80 +3,45 @@
...
-                }
-                return values;
-            }
 
             selection.each(function(data) {
 
                 // add a 'root' g element on the first enter selection. This ensures
                 // that it is just added once
-                container = d3.select(this);
+                var container = d3.select(this);
```
jleft: The comment doesn't seem to correspond to this line?

janakerman: Yeah - it didn't seem to make sense before the change either. I'll remove it as I'm not sure what it relates to.

chrisprice:
For me, other than some potential naming improvements around y0 and out, I think this looks good to go in now. As naming things is fickle, I'll just merge it and if we think it really needs changing it's a quick thing to do.

PR #304, opened 2015-04-20
Add fc.utilities.rebind

chrisprice:
Allows a mapping object to be specified when rebinding e.g. -
```js
fc.utlities.rebind(target, source, {
  targetName: 'sourceName'
});
```

### DIFF
--- a/components/utilities/rebind.js
+++ b/components/utilities/rebind.js
@@ -1,0 +1,18 @@
+(function(d3, fc) {
+    'use strict';
+
+    fc.utilities.rebind = function(target, source, mappings) {
+        if (typeof(mappings) !== 'object') {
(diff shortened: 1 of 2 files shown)


components/utilities/rebind.js
```diff
@@ -0,0 +1,18 @@
+(function(d3, fc) {
+    'use strict';
+
```
janakerman: I'm not sure what our conventions are on documentation but it might be helpful to add a documentation header here explaining the mapping object, i.e whether the key relates to the source or value.

chrisprice:
Thanks @janakerman, I've added some docs in.

janakerman:
Looks good!

Issue #305, opened 2015-04-22
Visual tests documentation


jleft:
The visual tests would benefit from some brief documentation explaining how they're structured, can be built and can be used.

PR #306, opened 2015-04-22
Add developing instructions to README

jleft:
Feedback encouraged! :)

I've tried to cover the important parts clearly and fairly concisely. Does it all make sense? Is there anything missing? Anything superfluous?

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,43 @@ The components included in this project support the following browsers:
 * Safari 3.2+
 * Opera
 
+## Developing
+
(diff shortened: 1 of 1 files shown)


janakerman:
I think that's at a suitable level of detail!

ColinEberhardt:
Looks good :-)

One thing though, this documents how to build, etc ... but what next? I think the final step should be to direct them to the visual tests.

Thanks.

ColinEberhardt:
Ah - I see you'd already thought of this :-)  #305

chrisprice:
Thanks @jleft

PR #308, opened 2015-04-27
Uses an official release of css-layout

ColinEberhardt:
When we first started using Facebook's CSS layout it didn;t have a package.json file. This now exists and they are releasing to npm. This commit updates our repo to use this.

### DIFF
--- a/components/utilities/layout.js
+++ b/components/utilities/layout.js
@@ -1,5 +1,5 @@
 /* globals computeLayout */
-(function(d3, fc) {
+(function(d3, fc, cssLayout) {
     'use strict';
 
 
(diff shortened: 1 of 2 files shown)


components/utilities/layout.js
```diff
@@ -110,7 +110,7 @@
                 layoutNodes.style.height = height;
 
                 // use the Facebook CSS goodness
-                computeLayout(layoutNodes);
```
chrisprice: Nitpick: Any reason for the difference between cssLayout and computeLayout?

ColinEberhardt: They moved to exposing an object with a few functions, rather than a single function. But they kept with the rather ambiguous name of computeLayout for the entry point into their API.

So, we could keep with their naming, resulting in:

    computeLayout.computeLayout(...)

I'm pretty happy either way!

chrisprice: I'd actually misread the diff and didn't spot that the method was now on the object... At least it's not a singleton factory adapter!

Let's leave it as it is, happy to merge.

PR #310, opened 2015-04-27
Removed chartLayout / chartBuilder from functional visual tests - DO NOT MERGE

ColinEberhardt:
Thought I'd share my work-in-progress. I'm removing chartLayout & chartBuilder, starting by removing them from the visual tests.

To hopefully remove test-cross over I am not replacing them with the new layout, instead I am stripping away any visual components that are not relevant to the given test.

image

It also means that charts in the test no longer re-size with the browser.

Let me know if you have any thoughts? I'll just keep trudging onwards!


ColinEberhardt:
Replaced by #313

PR #311, opened 2015-04-27
Linear time series

chrisprice:
For now the rebound properties are just what I needed to get the visual tests to run. I expect that we can add more as we need them.

### DIFF
--- a/components/chart/linearTimeSeries.js
+++ b/components/chart/linearTimeSeries.js
@@ -3,7 +3,7 @@
 
     fc.charts.linearTimeSeries = function() {
 
-        var xScale = d3.time.scale();
+        var xScale = fc.scale.dateTime();
         var yScale = d3.scale.linear();
         var xAxis = d3.svg.axis()
             .scale(xScale)
@@ -100,6 +100,7 @@
         };
 
         fc.utilities.rebind(linearTimeSeries, xScale, {
+            xDiscontinuityProvider: 'discontinuityProvider',
             xDomain: 'domain',
             xNice: 'nice'
         });
--- a/components/chart/linearTimeSeries.js
+++ b/components/chart/linearTimeSeries.js
@@ -18,18 +18,18 @@
 
                 var container = d3.select(this);
 
-                var mainContainer = container.selectAll('svg')
-                    .data([data]);
+                var mainContainer = fc.utilities.simpleDataJoin(container,
+                        'linear-time-series', [data]);
                 mainContainer.enter()
-                    .append('svg')
+                    .layout('flex', 1);
+
+                var plotAreaContainer = mainContainer.selectOrAppend('svg', 'plot-area');
+                plotAreaContainer.enter()
                     .attr('overflow', 'hidden')
                     .layout('flex', 1);
 
-                var background = mainContainer.selectAll('rect.background')
-                    .data([data]);
-                background.enter()
-                    .append('rect')
-                    .attr('class', 'background')
+                mainContainer.selectOrAppend('rect', 'background')
+                    .enter()
                     .layout({
                         position: 'absolute',
                         top: 0,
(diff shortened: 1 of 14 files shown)


components/chart/linearTimeSeries.js
```diff
@@ -0,0 +1,127 @@
+(function(d3, fc) {
+    'use strict';
+
+    fc.charts.linearTimeSeries = function() {
+
+        var xScale = d3.time.scale();
```
ColinEberhardt: should this be out own time scale - that permits discontinuities? you'd just have to rebind `discontinuityProvider` to allow people to use it.

components/chart/linearTimeSeries.js
```diff
@@ -0,0 +1,127 @@
...
+        var yScale = d3.scale.linear();
+        var xAxis = d3.svg.axis()
+            .scale(xScale)
+            .orient('bottom');
+        var yAxis = d3.svg.axis()
+            .scale(yScale)
+            .orient('left');
```
ColinEberhardt: Minor point, but I think most financial charts use a right-hand axis. I guess the reason being that you are more likely to be interested in the most recent prices.

components/chart/linearTimeSeries.js
```diff
@@ -0,0 +1,127 @@
...
+
+        fc.utilities.rebind(linearTimeSeries, xAxis, {
+            xTicks: 'ticks'
+        });
+
+        fc.utilities.rebind(linearTimeSeries, yAxis, {
+            yTicks: 'ticks'
+        });
+
+        linearTimeSeries.layout = fc.utilities.property(fc.utilities.layout());
```
ColinEberhardt: is there any need to expose layout? What might it be used for?

visual-tests/src/test-fixtures/series/line.js
```diff
@@ -1,12 +1,9 @@
 (function(d3, fc) {
     'use strict';
 
-    var data = fc.utilities.dataGenerator().startDate(new Date(2014, 1, 1))(50);
-
-    var chart = d3.select('#line'),
-        chartLayout = fc.utilities.chartLayout();
-
-    chart.call(chartLayout);
+    var dataGenerator = fc.utilities.dataGenerator()
```
ColinEberhardt: I think `fc.charts.linearTimeSeries` should have its own visual test, rather than adapting the current line series test.

ColinEberhardt:
This looks really cool - a few minor comments, but otherwise :+1: from me!

visual-tests/src/test-fixtures/series/line.js
chrisprice: Agreed, I'll add one.

components/chart/linearTimeSeries.js
chrisprice: I think we have a problem when it comes to the composability of layouts i.e. a layout of charts which themselves use layout internally is going to do a lot of laying out! 
However, I don't think this is the right approach either. I'll remove it for now.

components/chart/linearTimeSeries.js
chrisprice: I was taking heavy inspiration from http://www.highcharts.com/stock/demo. Unfortunately this is something that ends up baked into the chart as it stands...

components/chart/linearTimeSeries.js
chrisprice: Good point, I'll add it.

components/chart/linearTimeSeries.js
ColinEberhardt: OK, I'm happy either way. The fact that it is baked-in isn't an issue in my opinion.

We really should promote the fact that the components are small, and encourage copy-paste-modify as a way of using this code.

ColinEberhardt:
FYI, your failing build is likely due to a breaking change in jscs, where unfortunately the major version was not bumped: https://github.com/ColinEberhardt/d3-financial-components/commit/1eb04f96950fb77a21693fedff38b07a3f0ba7f0

chrisprice:
Thanks for the heads up but I think this one was me :(

chrisprice:
Merging despite the excessive commits, I'd like to keep the history of the selectOrAppend.

PR #313, opened 2015-04-28
Removed chartLayout and chartBuilder components

ColinEberhardt:
The chartLayout and chartBuilder components have been moved into the visual tests folder. They have been removed from a number of the simpler visual tests, but for the more complex tests they are still used. In future we will probably replace them entirely with our first simple 'chart' component.

I have also removed the following tests:

 - indicators-combined - we already combine series / indicators in a number of places, so this felt superfluous.
 - d3-time-d3-linear - we don't need to test D3 components!
 - fc-time-d3-linear - we already have a test for our axis with discontinuities, I think that is enough. This functionality also has unit tests
 - bar - this just rendered a bar chart


chrisprice:
Thanks! I think there's a race condition between this PR and the linearTimeSeries one so I'll merge them both then deal with the fallout.

PR #314, opened 2015-04-29
Fixes #303 - Stacked bar series reuses bar series

### LINKED ISSUE #303: Stacked bar series should use stacked bar series.
The implementation for the bar series and the stacked bar series are quite similar. The stacked bar series would wrap the bar series to reduce code duplication. A good example of where we do this already is `movingAverage.js` where we use `fc.series.line()` internally.

janakerman:
#303 The stacked series simply sets the various accessors, the baseline,
and the scales on the bar series, letting the bar series do the
work.


chrisprice:
Maybe y0 rather than baseline?

components/series/bar.js
```diff
@@ -23,9 +24,13 @@
                     .attr('x', function(d) {
                         return x(d) - width / 2;
                     })
-                    .attr('y', function(d) { return y(d); })
+                    .attr('y', function(d) {
```
chrisprice: The anonymous function isn't needed here.

components/series/bar.js
```diff
@@ -38,6 +43,7 @@
         bar.barWidth = fc.utilities.functorProperty(fc.utilities.fractionalBarWidth(0.75));
         bar.yValue = fc.utilities.property(function(d) { return d.close; });
         bar.xValue = fc.utilities.property(function(d) { return d.date; });
+        bar.baseline = fc.utilities.property(function(d) { return 0; });
```
chrisprice: If you use a functor property then you won't need the anonymous function.

components/series/stackedBar.js
```diff
@@ -7,39 +7,29 @@
...
-                // update
-                bar.attr('x', function(d) { return stackedBar.xScale.value(stackLayout.x()(d)) - width / 2; })
-                    .attr('y', function(d) {
-                        return stackedBar.yScale.value(stackLayout.y()(d) + stackedBar.y0.value(d));
-                    })
-                    .attr('width', width)
-                    .attr('height', function(d) {
-                        var baselineValue = stackedBar.y0.value(d);
-                        var topValue = stackLayout.y()(d);
+                var series = container.selectAll('g.stacked-bar').data(layeredData);
```
chrisprice: This doesn't follow the line break conventions.

components/series/stackedBar.js
```diff
@@ -7,39 +7,29 @@
...
 
-                        var bottomPixel = stackedBar.yScale.value(baselineValue);
-                        var topPixel = stackedBar.yScale.value(topValue + baselineValue);
+                series.enter().append('g').attr('class', 'stacked-bar');
```
chrisprice: And here.

components/series/stackedBar.js
```diff
@@ -52,10 +42,17 @@
         stackedBar.yScale = fc.utilities.property(d3.scale.linear());
 
         // Implicitly dependant on the implementation of the stack layout's `out`.
-        stackedBar.y0 = fc.utilities.property(function(d) {
+        stackedBar.y0Value = fc.utilities.property(function(d) {
```
chrisprice: I think these names are good.

components/utilities/rebind.js
```diff
@@ -18,6 +18,7 @@
                     return value === source ? target : value;
                 };
             });
+        return target;
```
chrisprice: Good catch!

chrisprice:
This really tidies things up, nice one!

chrisprice:
I'm guessing this is good to merge now. Thanks @janakerman!

Issue #249, opened 2015-01-16
Use update pattern for mouse over overlay in tools


jleft:
```js
 if (!data.__fan__.overlay) {
    container.append('rect')
        .style('visibility', 'hidden');
    data.__fan__.overlay = true;
}
```

PR #317, opened 2015-05-01
Low barrel example

chrisprice:
Currently a visual test but it may be moved to the website.

It also removes the crazyness from crosshairs as referenced by #249, this needs rolling out though...


ColinEberhardt:
Cool example, I'll go through the code later, and probably merge it into the website :-)

ColinEberhardt:
I've merged in your component changes, and will add low-barrel to the website shortly.

PR #319, opened 2015-05-01
gh-pages - ready for review

ColinEberhardt:
The gh-pages are almost complete, needs a few additional bits of information, but I think now is a good time for review:

http://colineberhardt.github.io/d3-financial-components/


chrisprice:
Looks good, minor points:

* [ ] The series are rendering outside their borders on the components screen.
* [ ] Maybe drop the stroke on the band series, the stroke down the side stops it looking very good.

ColinEberhardt:
- [ ] Prevent google from crawling forks

jleft:
Looking good :+1:

PR #322, opened 2015-05-01
Change dataGenerator namespace from fc.utilities.dataGenerator to fc.dataGenerator

jleft:


chrisprice:
Thanks @jleft. 

A good reason we should be using proper imports (#275).

PR #324, opened 2015-05-01
Restructure algorithms/math

jleft:


ColinEberhardt:
Thanks, let me know when you at done with all the restructuring I'll update the website

jleft:
@ColinEberhardt - I've done all of the restructuring I was intending to do. As long as everyone is happy with this, then I think it's fine to update the website. Sorry if the changes have caused extra work - hopefully it won't be much more than find & replace!

jleft:
The only other change which may be worth considering is renaming the `components` directory to `src` - I'm not sure if this is worth it now though?

chrisprice:
I would also like to see components renamed but I'm not sure how much
effort is required?

PR #326, opened 2015-05-01
Add grunt-contrib-connect, a simple web server for viewing the visual tests

jleft:
My aim is to make it even easier for someone to get a web server started to view the visual tests. If it's not worth the extra bloat, then feel free to close :-).

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -76,6 +76,19 @@ module.exports = function (grunt) {
             }
         },
 
+        connect: {
+            options: {
+                base: 'visual-tests/dist',
+                useAvailablePort: true
+            },
+            dev: { },
+            keepalive: {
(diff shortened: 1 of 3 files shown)


ColinEberhardt:
Nice idea, anything that makes it easier for people to build and contribute gets my vote.

In the readme, perhaps rather than an exhaustive list of the tasks, just detail the key ones? Or highlight the two (one for development, one for release builds) that are most commonly used?

jleft:
> In the readme, perhaps rather than an exhaustive list of the tasks, just detail the key ones? Or highlight the two (one for development, one for release builds) that are most commonly used?

Sure, I've updated :-)

ColinEberhardt:
Awesome!

PR #331, opened 2015-05-05
Interpolation support

### LINKED ISSUE #296: Expose d3 line path data generator in line series
The [line series uses D3's line generator](https://github.com/ScottLogic/d3-financial-components/blob/15ed4891537a32252c743096baa29fc09008537f/components/series/line.js#L10) to generate the path data for the line.

This line generator is great, and has lots of useful features, such an [interpolation](https://github.com/mbostock/d3/wiki/SVG-Shapes#line_interpolate), but the user isn't able to modify these properties; it would be nice if they could.

jleft:
Resolves #296

### DIFF
--- a/src/series/area.js
+++ b/src/series/area.js
@@ -42,6 +42,6 @@
         area.xValue = fc.utilities.property(function(d) { return d.date; });
 
 
-        return d3.rebind(area, areaData, 'interpolate');
+        return d3.rebind(area, areaData, 'interpolate', 'tension');
     };
 }(d3, fc));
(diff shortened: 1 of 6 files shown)


ColinEberhardt:
This looks like a useful addition. Although if  `interpolate` is being exposed does it also make sense to expose `tension`?

jleft:
> Although if `interpolate` is being exposed does it also make sense to expose `tension`?

Yes, I think so - good idea! I've updated to expose `tension` too.

ColinEberhardt:
Nice one - thanks.

PR #332, opened 2015-05-05
Add seedrandom to support seeded random number generation for the visual tests

jleft:
Many visual tests rely upon `fc.dataGenerator`, which uses `Math.random()` to generate the data values. This is great for producing varied data, but sometimes, when debugging for example, it it is useful to use the same dataset.

This pull request adds [seedrandom](https://github.com/davidbau/seedrandom) as a development dependency, which allows `Math.random()` to be seeded and `fc.dataGenerator` to produce consistent values. This can be included in the visual tests by running `grunt build:visual-tests-debug`.

Does this seem like a good idea? Is this a reasonable approach?

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -19,6 +19,22 @@ module.exports = function (grunt) {
             visualTestJsFiles: [
                 'visual-tests/src/**/*.js'
             ],
+            visualTestSiteFiles: [
+                // Index page
+                {
+                    expand: true,
+                    cwd: 'visual-tests/src/site/pages/',
+                    src: ['index.hbs'],
+                    dest: 'visual-tests/dist/'
+                },
+                // Test fixtures
+                {
+                    expand: true,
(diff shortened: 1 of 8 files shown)


ColinEberhardt:
Definitely a worthwhile addition. Although I wonder whether it is worth adding another test target just for this? I'd be happy for the visual tests to always run with the seeded random number.

@chrisprice any thoughts?

chrisprice:
Maybe we could always include seedrandom in the page but seed it with the query string from the URL if it exists?

jleft:
Thanks for the feedback :)

> Although I wonder whether it is worth adding another test target just for this?

Agreed, I didn't really like adding the extra target - I've removed this.

> Maybe we could always include seedrandom in the page but seed it with the query string from the URL if it exists?

Nice idea, I've updated to include this.

chrisprice:
LGTM, thanks!

PR #333, opened 2015-05-06
Update Travis to run against Node.js 0.12 and io.js

jleft:
As Node.js is now at version 0.12.x, I thought it would be worth Travis running the build against this. I've also added io.js as an allowed failure - this isn't as important, but may be useful/interesting to have.


ColinEberhardt:
Thanks, definitely worth keeping up to date.

PR #338, opened 2015-05-07
Added a sparkline component

ColinEberhardt:
I've had enough of the website for a while, thought it was time to write an actual component!

image

The implementation makes use of the multi-series mapping and point series decoration. It's really good to see how the lower-level components can be assembled into higher-level charts :-)

### DIFF
--- a/src/chart/sparkline.js
+++ b/src/chart/sparkline.js
@@ -1,0 +1,117 @@
+(function(d3, fc) {
+    'use strict';
+
+    fc.charts.sparkline = function() {
+
+        // creates an array with four elements, representing the high, low, open and close
+        // values of the given array
+        function highLowOpenClose(data) {
+            var xValueAccessor = sparkline.xValue(),
+                yValueAccessor = sparkline.yValue();
+
+            var high = d3.max(data, yValueAccessor);
+            var low = d3.min(data, yValueAccessor);
+
+            function elementWithYValue(value) {
+                return data.filter(function(d) {
+                    return yValueAccessor(d) === value;
+                })[0];
+            }
+
+            return [{
+                    x: xValueAccessor(data[0]),
+                    y: yValueAccessor(data[0])
+                }, {
+                    x: xValueAccessor(elementWithYValue(high)),
+                    y: high
+                }, {
+                    x: xValueAccessor(elementWithYValue(low)),
+                    y: low
+                }, {
+                    x: xValueAccessor(data[data.length - 1]),
+                    y: yValueAccessor(data[data.length - 1])
+                }];
+        }
+
+        var xScale = fc.scale.dateTime();
+        var yScale = d3.scale.linear();
+        var line = fc.series.line();
+
+        // configure the point series to render the data from the
+        // highLowOpenClose function
+        var point = fc.series.point()
+            .xValue(function(d) { return d.x; })
+            .yValue(function(d) { return d.y; })
+            .decorate(function(sel) {
+                sel.attr('class', function(d, i) {
+                    switch (i) {
+                        case 0: return 'open';
+                        case 1: return 'high';
+                        case 2: return 'low';
+                        case 3: return 'close';
+                    }
+                });
+            });
+
+        var multi = fc.series.multi()
+            .series([line, point])
+            .mapping(function(data, series) {
+                switch (series) {
+                    case point:
+                        return highLowOpenClose(data);
+                    default:
+                        return data;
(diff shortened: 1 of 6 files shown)


src/chart/sparkline.js
```diff
@@ -0,0 +1,117 @@
...
+                multi.xScale(xScale)
+                    .yScale(yScale);
+
+                mainContainer.call(multi);
+
+            });
+        };
+
+        fc.utilities.rebind(sparkline, xScale, {
+            xDiscontinuityProvider: 'discontinuityProvider',
```
ColinEberhardt: Not really sure this is worth re-binding. Sparklines are so small that discontinuities would be a bit pointless.

src/chart/sparkline.js
```diff
@@ -0,0 +1,117 @@
...
+        // creates an array with four elements, representing the high, low, open and close
+        // values of the given array
+        function highLowOpenClose(data) {
+            var xValueAccessor = sparkline.xValue(),
+                yValueAccessor = sparkline.yValue();
+
+            var high = d3.max(data, yValueAccessor);
+            var low = d3.min(data, yValueAccessor);
+
+            function elementWithYValue(value) {
```
chrisprice: I can't think of a better way of doing this but it feels wrong...

src/chart/sparkline.js
```diff
@@ -0,0 +1,117 @@
...
+        var yScale = d3.scale.linear();
+        var line = fc.series.line();
+
+        // configure the point series to render the data from the
+        // highLowOpenClose function
+        var point = fc.series.point()
+            .xValue(function(d) { return d.x; })
+            .yValue(function(d) { return d.y; })
+            .decorate(function(sel) {
+                sel.attr('class', function(d, i) {
```
chrisprice: Could you use classed?

src/chart/sparkline.js
```diff
@@ -0,0 +1,117 @@
...
+
+        var sparkline = function(selection) {
+
+            point.radius(sparkline.radius.value);
+
+            selection.each(function(data) {
+
+                var container = d3.select(this);
+
+                var mainContainer = container.selectAll('g')
```
chrisprice: Do we need the container?

src/chart/sparkline.js
```diff
@@ -0,0 +1,117 @@
...
+                    .data([data]);
+                mainContainer.enter()
+                    .append('g')
+                    .classed('sparkline', true)
+                    .layout({
```
chrisprice: Do we need the layout?

chrisprice: As in could we just pad the ranges?

src/chart/sparkline.js
ColinEberhardt: Yeah, I know what you mean - having to scan the array twice, is not ideal. Perhaps I should replace this with my own reduce-based implementation.

src/chart/sparkline.js
ColinEberhardt: Yes, but I think it would be more messy ...

```
sel.classed({
  open: function(d, i) { return i === 0; },
  high: function(d, i) { return i === 1; },
  low: function(d, i) { return i === 2; },
  close: function(d, i) { return i === 3; },
})
```

src/chart/sparkline.js
ColinEberhardt: > As in could we just pad the ranges?

Not easily - if I understand you correctly!

The margin around the container is in the screen coordinate system, it is the margin required to ensure that the circles do not fall outside of the bounds on the `container` element. 

To achieve this via the scale ranges would require converting this into the domain coordinate system.

src/chart/sparkline.js
chrisprice: Oh yea, kind of forgot we didn't have lamdas (was watching a talk on them!)

src/chart/sparkline.js
chrisprice: Could we not just pad the range (pixel) values by the radius value?

ColinEberhardt:
> Could we not just pad the range (pixel) values by the radius value?

This is what I've done, although it did involve pulling out the logic from the layout component for measuring the dimensions of an element (minus padding). Which probably isn't a bad thing!

chrisprice:
Looks good to me. It'll definitely add some "spark" to the library...

ColinEberhardt:
Thanks Chris - I'll merge :-)

PR #339, opened 2015-05-07
Added sparklines to website

ColinEberhardt:
This assumes that #338 is merged ;-)


ColinEberhardt:
As #338 has landed, I'll merge this also.

PR #347, opened 2015-05-14
Always pass index to value accessors (fixes #316)

### LINKED ISSUE #316: Make xValue accessor more flexible
The xValue function used by point series (and others) should be changed to supply both the current datapoint and the index to the accessor function:

https://github.com/ScottLogic/d3-financial-components/blob/master/components/series/point.js#L7

Changing from this:

    var x = function(d) { return point.xScale.value(point.xValue.value(d)); };

To something like this:

    var x = function(d, i) { return point.xScale.value(point.xValue.value(d, i)); };

This would allow you to use the series with data where the x-value is based on the order within the data array

chrisprice:
Also tries to minimise accessor call counts. This made testing easier but I assume it can't hurt performance either.

### DIFF
--- a/src/series/area.js
+++ b/src/series/area.js
@@ -7,17 +7,17 @@
             xScale = d3.time.scale(),
             yScale = d3.scale.linear(),
             y0Value = d3.functor(0),
-            y1Value = function(d) { return d.close; },
-            xValue = function(d) { return d.date; };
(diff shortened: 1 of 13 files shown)


ColinEberhardt:
Wow - unit tests :-)

PR #348, opened 2015-05-14
Remove out parameters (fixes #278)

### LINKED ISSUE #278: Change outputValue function invocation semantics
Currently looks like -
```js
percentageChange.outputValue(function(obj, value) { obj.foo = value; return obj; });
```

Which means that you can't use it with ```fc.utilities.fn.identity``` but we could make it possible by changing it to -
```js
percentageChange.outputValue(function(value) { this.foo = value; return obj; });
```

chrisprice:


tests/indicators/algorithms/relativeStrengthIndicatorSpec.js
```diff
@@ -77,14 +77,14 @@
...
             it('should calculate correctly with a windowSize of 2', function() {
                 rsi.windowSize(2);
-                expect(rsi(data)).toEqual([50, 50]);
+                expect(rsi(data)).toEqual([undefined, 50, 50]);
             });
 
             it('should calculate correctly with a windowSize of 3', function() {
                 rsi.windowSize(3);
-                expect(rsi(data)).toEqual([22.222222222222214]);
+                expect(rsi(data)).toEqual([undefined, undefined, 22.222222222222214]);
```
ColinEberhardt: I presume this is a side-effect of zip/mapping the two data series together, they have to be of the same length, so an algorithm must output an array of the same length as the input array?

chrisprice: Yep, that's it. Previously we manipulated the incoming data so we implicitly had correlation between the input and calculated values. Obviously now we no longer have that...

Issue #346, opened 2015-05-14
Series are not consistent in their handling of undefined values


chrisprice:
Area and line (which wrap their d3 counterparts) contain a ```defined``` filter function which allows undefined data values to be skipped. The rest of the series do not have this functionality.

Therefore swapping from a line series to a bar series will cause a runtime error.

I'm not sure if we should remove the functionality from area/line or add it to the others. It seems like nice functionality but forcing an extra filter pass of the data before doing the data-join doesn't sound great either.

ColinEberhardt:
The addition of `xyBase` and `ohlcBase` has resolved this.

PR #350, opened 2015-05-14
Check if result from Bollinger Bands algorithm is defined before attempting to access its properties

jleft:
`tuple[1]`, the result of the Bollinger Bands algorithm, can be `undefined`. When the Bollinger Bands component attempts to access properties of this, a `TypeError` will be thrown.

We should check this to avoid the error. There are several approaches to this; is there are preferred solution?

### DIFF
--- a/src/indicators/bollingerBands.js
+++ b/src/indicators/bollingerBands.js
@@ -56,9 +56,11 @@
 
                 data = d3.zip(data, algorithm(data))
                     .map(function(tuple) {
-                        tuple[0].upper = tuple[1].upper;
-                        tuple[0].average = tuple[1].average;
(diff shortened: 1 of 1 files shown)


chrisprice:
Doh, must have missed this one. Thanks for the fix!

jleft:
No problem :smiley: 

Don't know if it's worth doing something similar with moving average and RSI? I know a `TypeError` wouldn't occur there, but it would be consistent and I'm not sure if adding `undefined` properties has any benefit?

chrisprice:
I suppose difference is summarised by #346. The undefined values are now required to allow correlation between the two arrays see https://github.com/ScottLogic/d3-financial-components/pull/348#discussion_r30331654.

PR #351, opened 2015-05-15
Fix point snapping edge cases

chrisprice:
Now works with any scale and supports snapping to data values as well as pixel values.


src/tools/crosshairs.js
```diff
@@ -6,12 +6,30 @@
         var event = d3.dispatch('trackingstart', 'trackingmove', 'trackingend'),
             xScale = d3.time.scale(),
             yScale = d3.scale.linear(),
-            snap = function(x, y) { return {x: x, y: y}; },
+            snap = function(x, y) {
+                // ordinal axes don't invert pixel values (interpolation doesn't
```
ColinEberhardt: Is this a no-op snap function? If so, looks like it could do with being made a utility function so that it doesn't need to be repeated.

ColinEberhardt:
Thanks for the update - I'd be happy for this to be merged once the conflicts are resolved.

PR #352, opened 2015-05-15
Allow undefinedValue to be specified for slidingWindow

chrisprice:
This allows for the logic in ```series.bollingerBands``` to be cleaned up.


jleft:
:+1:

PR #353, opened 2015-05-15
Finish removing chartBuilder/Layout

### LINKED ISSUE #315: Tidy up fallout of #313/#311
* [x] #311 introduced a new chart layout which is appropriate for replacing the chart builder in #313.
* [x] #313 introduced the concept of simplified visual tests, #311 added a complex multi visual test which should be simplified.
* [x] #313 set an [explicit width on the iframe](https://github.com/ScottLogic/d3-financial-components/pull/313/files#diff-d03ea6571b32bddbca4b7f3afede1b2cL23), this should be removed if possible.

chrisprice:
Fixes #315


src/tools/measure.js
```diff
@@ -102,6 +102,7 @@
         }
 
         function mouseenter() {
+            console.log('enter');
```
jleft: Remove?

chrisprice:
Good spot, fixed.

PR #360, opened 2015-05-15
Extract SVG path generator from candlestick/ohlc series. 

### LINKED ISSUE #354: Split path string generators out of ohlc and candlestick series
They can be re-used in other ways, the series should just combine scales with renderers.

chrisprice:
Fixes #354.

I think it's ultimately the right thing to do but the fake index stuff is horrible. Can anyone think of a better way?

### DIFF
--- a/src/fc.js
+++ b/src/fc.js
@@ -20,6 +20,7 @@
             discontinuity: {}
         },
         series: {},
+        svg: {},
         tools: {},
         utilities: {}
     };
--- a/src/series/candlestick.js
+++ b/src/series/candlestick.js
@@ -13,8 +13,6 @@
             yCloseValue = function(d, i) { return d.close; },
             barWidth = fc.utilities.fractionalBarWidth(0.75);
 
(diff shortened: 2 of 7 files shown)


chrisprice:
I've removed the fake index stuff in c8bec0c. I'm happy with it now, let me know what you think.

jleft:
:+1:

ColinEberhardt:
:+1: from me ... although it is a shame that there is so much duplication of the configuration properties. Could the pathGenerator properties be re-bound onto candlestick / ohlc? Although I guess an issue here is that the path generator has is stateful.

@jleft / @chrisprice  - it would be great if one of you could a few notes to `simpleDataJoin` that describes what it does and why it exists!

chrisprice:
Good point, added #366 to track the simpleDataJoin documentation.

chrisprice:
As discussed the value accessor index arguments prevent simple re-binding. The svg path generators operate on each data item individually so would always pass an index of 0.

PR #361, opened 2015-05-15
Use d3.pairs in place of custom implementation

### LINKED ISSUE #359: fc.utilities.fractionalBarWidth re-implements d3.pairs
From fc.utilities.fractionalBarWidth -

```js
// creates a new array as a result of applying the 'fn' function to
// the consecutive pairs of items in the source array
function pair(arr, fn) {
    var res = [];
    for (var i = 1; i < arr.length; i++) {
        res.push(fn(arr[i], arr[i - 1]));
    }
    return res;
}
```

From the docs -

d3.pairs(array)

For each adjacent pair of elements in the specified array, returns a new array of tuples of element i and element i - 1. For example:
```
d3.pairs([1, 2, 3, 4]); // returns [[1, 2], [2, 3], [3, 4]]
```
If the specified array has fewer than two elements, returns the empty array.

chrisprice:
Fixes #359

### DIFF
--- a/src/utilities/fractionalBarWidth.js
+++ b/src/utilities/fractionalBarWidth.js
@@ -15,20 +15,11 @@
 
             pixelValues.sort();
 
-            // creates a new array as a result of applying the 'fn' function to
-            // the consecutive pairs of items in the source array
-            function pair(arr, fn) {
-                var res = [];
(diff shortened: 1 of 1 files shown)


ColinEberhardt:
Heh, this has happened far too often! I never thought to check d3 in this case.

PR #364, opened 2015-05-18
Give me an 'E', give me an 'M', give me an 'A' - what have you got?

ColinEberhardt:
Exponential Moving Average!

### DIFF
--- a/src/indicators/algorithms/exponentialMovingAverage.js
+++ b/src/indicators/algorithms/exponentialMovingAverage.js
@@ -3,7 +3,7 @@
 
     fc.indicators.algorithms.exponentialMovingAverage = function() {
 
-        var days = 9,
+        var windowSize = 9,
             value = fc.utilities.fn.identity;
 
         function undefinedArrayOfLength(length) {
@@ -11,15 +11,15 @@
         }
 
         var exponentialMovingAverage = function(data) {
-            if (data.length < days) {
+            if (data.length < windowSize) {
                 return undefinedArrayOfLength(data.length);
             }
-            var ema = undefinedArrayOfLength(days - 1);
-            var previous = d3.mean(data.slice(0, days).map(value));
+            var ema = undefinedArrayOfLength(windowSize - 1);
+            var previous = d3.mean(data.slice(0, windowSize).map(value));
             ema.push(previous);
 
-            var alpha = 2 / (days + 1);
-            for (var index = days; index < data.length; index++) {
+            var alpha = 2 / (windowSize + 1);
+            for (var index = windowSize; index < data.length; index++) {
                 var nextValue = value(data[index]) * alpha + (1 - alpha) * previous;
                 ema.push(nextValue);
                 previous = nextValue;
@@ -28,11 +28,11 @@
             return ema;
         };
 
-        exponentialMovingAverage.days = function(x) {
+        exponentialMovingAverage.windowSize = function(x) {
             if (!arguments.length) {
-                return days;
+                return windowSize;
             }
-            days = x;
+            windowSize = x;
             return exponentialMovingAverage;
         };
 
--- a/src/indicators/algorithms/exponentialMovingAverage.js
+++ b/src/indicators/algorithms/exponentialMovingAverage.js
@@ -6,26 +6,27 @@
         var windowSize = 9,
             value = fc.utilities.fn.identity;
 
-        function undefinedArrayOfLength(length) {
-            return Array.apply(undefined, new Array(length));
-        }
-
         var exponentialMovingAverage = function(data) {
-            if (data.length < windowSize) {
-                return undefinedArrayOfLength(data.length);
-            }
-            var ema = undefinedArrayOfLength(windowSize - 1);
-            var previous = d3.mean(data.slice(0, windowSize).map(value));
-            ema.push(previous);
 
             var alpha = 2 / (windowSize + 1);
-            for (var index = windowSize; index < data.length; index++) {
-                var nextValue = value(data[index]) * alpha + (1 - alpha) * previous;
-                ema.push(nextValue);
-                previous = nextValue;
-            }
+            var previous;
+            var initialAccumulator = 0;
 
-            return ema;
+            return data.map(function(d, i) {
+                    if (i < windowSize - 1) {
+                        initialAccumulator += d;
+                        return undefined;
+                    } else if (i === windowSize - 1) {
+                        initialAccumulator += d;
+                        var initialValue = initialAccumulator / windowSize;
+                        previous = initialValue;
+                        return initialValue;
+                    } else {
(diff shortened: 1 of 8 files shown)


src/indicators/movingAverage.js
```diff
@@ -3,70 +3,17 @@
...
-        movingAverage.yValue = function(x) {
-            if (!arguments.length) {
-                return yValue;
-            }
-            yValue = x;
-            return movingAverage;
+            indicator(selection);
         };
 
         d3.rebind(movingAverage, algorithm, 'windowSize');
```
ColinEberhardt: I think this is probably better re-bound as 'days'

chrisprice: I'm not so sure. It so happens that our "data generatored" data is in days but nothing about our code enforces or requires this.

src/indicators/algorithms/exponentialMovingAverage.js
```diff
@@ -0,0 +1,49 @@
+(function(d3, fc) {
+    'use strict';
+
+    fc.indicators.algorithms.exponentialMovingAverage = function() {
+
+        var days = 9,
```
chrisprice: See general comment about days.

src/indicators/movingAverage.js
ColinEberhardt: good point - I'll stick with `windowSize` for both SMA and EMA

src/indicators/algorithms/exponentialMovingAverage.js
```diff
@@ -0,0 +1,49 @@
...
+            value = fc.utilities.fn.identity;
+
+        function undefinedArrayOfLength(length) {
+            return Array.apply(undefined, new Array(length));
+        }
+
+        var exponentialMovingAverage = function(data) {
```
chrisprice: I wonder if this could be rewritten more in the style of the sliding window function? My preference is based on the top level ```data.map``` function making it obvious that the resulting array will always be of the same length.

src/indicators/algorithms/exponentialMovingAverage.js
```diff
@@ -0,0 +1,49 @@
...
+            if (data.length < days) {
+                return undefinedArrayOfLength(data.length);
+            }
+            var ema = undefinedArrayOfLength(days - 1);
+            var previous = d3.mean(data.slice(0, days).map(value));
+            ema.push(previous);
+
+            var alpha = 2 / (days + 1);
+            for (var index = days; index < data.length; index++) {
+                var nextValue = value(data[index]) * alpha + (1 - alpha) * previous;
```
chrisprice: value is not invoked with an index

src/indicators/algorithms/exponentialMovingAverage.js
```diff
@@ -0,0 +1,49 @@
+(function(d3, fc) {
+    'use strict';
+
+    fc.indicators.algorithms.exponentialMovingAverage = function() {
+
+        var days = 9,
+            value = fc.utilities.fn.identity;
+
+        function undefinedArrayOfLength(length) {
```
chrisprice: slidingWindow makes the undefined value configurable. We should probably decide whether that makes sense or not, then be consistent here.

src/indicators/exponentialMovingAverage.js
```diff
@@ -0,0 +1,19 @@
...
+    'use strict';
+
+    fc.indicators.exponentialMovingAverage = function() {
+
+        var algorithm = fc.indicators.algorithms.exponentialMovingAverage();
+
+        var indicator = fc.indicators.singleValuedIndicator(algorithm);
+
+        var exponentialMovingAverage = function(selection) {
+            indicator(selection);
```
chrisprice: selection.call(indicator) ?

src/indicators/singleValuedIndicator.js
```diff
@@ -0,0 +1,75 @@
...
+                return yValue;
+            }
+            yValue = x;
+            return singleValuedIndicator;
+        };
+        singleValuedIndicator.algorithm = function(x) {
+            if (!arguments.length) {
+                return yValue;
+            }
+            yValue = x;
```
chrisprice: copy/paste error?

src/indicators/singleValuedIndicator.js
```diff
@@ -0,0 +1,75 @@
+(function(d3, fc) {
+    'use strict';
+
+    fc.indicators.singleValuedIndicator = function(algorithm) {
```
chrisprice: I can see why the algorithm is passed in the constructor (it is required to function) but other components have similar restrictions and just default it to something sensible (e.g. series default x/y scales). Consider removing it?

src/indicators/singleValuedIndicator.js
```diff
@@ -0,0 +1,75 @@
...
+            algorithm.value(yValue);
+
+            outputLine.xScale(xScale)
+                .yScale(yScale)
+                .xValue(xValue)
+                .yValue(function(d, i) { return d.singleValuedIndicator; });
+
+            selection.each(function(data) {
+
+                data = d3.zip(data, algorithm(data))
```
chrisprice: I know I came up with this pattern but I'm now thinking it should be changed to make it's intent clearer. Instead of a map (which implies that nothing is going to be modified), I wonder if it should be a forEach -
```js
 d3.zip(data, algorithm(data))
  .forEach(function(tuple) {
    tuple[0].singleValuedIndicator = tuple[1];
  });
```

src/indicators/singleValuedIndicator.js
```diff
@@ -0,0 +1,75 @@
...
+                    .map(function(tuple) {
+                        tuple[0].singleValuedIndicator = tuple[1];
+                        return tuple[0];
+                    });
+
+                d3.select(this)
+                    .datum(data)
```
chrisprice: This datum call isn't needed, we're modifying the original data.

src/indicators/algorithms/exponentialMovingAverage.js
ColinEberhardt: I was thinking about that too - I'll add that as part of the undefinedInputAdapter that I am working on seperately, as this will no doubt need to make use of this adapted property.

src/indicators/algorithms/exponentialMovingAverage.js
ColinEberhardt: Yes, it can be re-written using `data.map`: https://github.com/ColinEberhardt/d3-financial-components/commit/a772faa0ae34f52fa981239c7749f10830ceb1ae - the implementation style I came up with was as a result of test-first development ;-)

I'm not sure I like the `map` implementations because both have side-effects as a result of the mapping function. However, your point about the array length is a good one.

chrisprice:
Other than the tests, :+1: from me.

PR #368, opened 2015-05-19
Updated to use an official release

ColinEberhardt:
As we are 0.x.x I haven't followed the sem-ver rules, I don't think we are stable enough for that yet!


chrisprice:
LGTM (we should probably start considering release notes when we get to semver as well)

ColinEberhardt:
I've started the ball rolling on release notes ...

https://github.com/ScottLogic/d3-financial-components/releases/tag/0.1.1

PR #372, opened 2015-05-19
support negative values with the bar series

ColinEberhardt:
This is a simple fix, although doesn't produce terribly ideal transitions @jleft - I think you mentioned this before?

Anyhow, this change does put us in a better place, removing the runtime errors that the code currently spews out. Also, I need this for MACD ;-)

### DIFF
--- a/src/series/bar.js
+++ b/src/series/bar.js
@@ -32,14 +32,21 @@
                         var x = xValueScaled(d, i),
                             y0 = y0Value(d, i),
                             barBottom = yScale(y0),
-                            barTop = yScale(y0 + y1Value(d, i));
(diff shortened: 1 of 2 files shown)


ColinEberhardt:
Ah, of course, it is the only other open PR - #291

ColinEberhardt:
Actually - I have a better idea ...

jleft:
> This is a simple fix, although doesn't produce terribly ideal transitions @jleft - I think you mentioned this before?

Yes, as `rect` doesn't support negative height, the transitions won't look great unless a `path` is used (other elements might be suitable too) - I see this is covered in #373 :)

PR #373, opened 2015-05-20
Series group elements are now translated, plus negative bar height support

ColinEberhardt:
Bar series is now rendered using our own bar svg element - Prevously we used a rect, which does not support a negative height. This commit replaces rect with a path, via a path-generator (similar to OHLC / candlestick).

Squashed commits:
[29599f2] ohlc parent group is translated
[585efc5] Candlestick parent group is now translated

See #258

NOTE: This breaks transition support for bar charts (#212), however, as bar was the only series type with support, my feeling is that at this stage it is better to have a more uniform method of rendering series.

### DIFF
--- a/src/series/bar.js
+++ b/src/series/bar.js
@@ -22,8 +22,7 @@
                 g.enter()
                     .append('path');
 
-                var width = barWidth(data.map(xValueScaled)),
-                    halfWidth = width / 2;
+                var width = barWidth(data.map(xValueScaled));
 
                 var pathGenerator = fc.svg.bar()
                     .width(width);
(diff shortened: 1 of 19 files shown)


ColinEberhardt:
Replaces #291

src/series/bar.js
```diff
@@ -16,34 +16,37 @@
...
-                            });
-                    });
+                    var x = xValueScaled(d, i),
+                        y0 = y0Value(d, i),
+                        barBottom = yScale(y0),
+                        barTop = yScale(y0 + y1Value(d, i));
 
-                decorate(series);
+                    var g = d3.select(this)
+                        .attr('transform', 'translate(' + (x - halfWidth) + ', ' + barTop + ')');
```
chrisprice: For the OHLC/Candlestick, we do the halfWidth calculation inside the generator. It probably makes sense to do the same thing here.

chrisprice:
Also, what's the relationship between this and @jleft's outstanding PR  #291?

PR #376, opened 2015-05-21
Add transition support to bar series

jleft:
#212; also related to #373 and (https://github.com/ScottLogic/d3-financial-components/pull/291#issuecomment-104223841).

### DIFF
--- a/src/series/bar.js
+++ b/src/series/bar.js
@@ -18,30 +18,40 @@
                 var container = d3.select(this);
                 var g = fc.utilities.simpleDataJoin(container, 'bar', data, xValue);
 
-                // enter
-                g.enter()
-                    .append('path');
-
                 var width = barWidth(data.map(xValueScaled));
 
                 var pathGenerator = fc.svg.bar()
                     .width(width);
 
+                var x = function(d, i) { return xValueScaled(d, i); },
+                    y0 = function(d, i) { return y0Value(d, i); },
+                    barTop = function(d, i) { return yScale(y0(d, i) + y1Value(d, i)); },
+                    barBottom = function(d, i) { return yScale(y0(d, i)); };
+
+                g.enter()
+                    .each(function(d, i) {
+                        d3.select(this)
+                            .attr('transform', 'translate(' + x(d, i) + ', ' + barBottom(d, i) + ')');
+                    })
+                    .append('path')
+                    .each(function(d, i) {
+                        pathGenerator.x(0)
+                            .y(0)
+                            .height(0);
+
+                        d3.select(this)
+                            .attr('d', pathGenerator([d]));
+                    });
+
                 g.each(function(d, i) {
-
-                    var x = xValueScaled(d, i),
-                        y0 = y0Value(d, i),
-                        barBottom = yScale(y0),
-                        barTop = yScale(y0 + y1Value(d, i));
-
-                    pathGenerator.x(d3.functor(0))
-                        .y(d3.functor(0))
-                        .height(function() { return barBottom - barTop; });
+                    pathGenerator.x(0)
+                        .y(0)
+                        .height(function() { return barBottom(d, i) - barTop(d, i); });
 
                     var barGroup = d3.select(this);
 
                     d3.transition(barGroup)
-                        .attr('transform', 'translate(' + x + ', ' + barTop + ')')
+                        .attr('transform', 'translate(' + x(d, i) + ', ' + barTop(d, i) + ')')
                         .select('path')
                         .attr('d', pathGenerator([d]));
                 });
--- a/src/series/bar.js
+++ b/src/series/bar.js
@@ -21,35 +21,29 @@
                 var width = barWidth(data.map(xValueScaled));
 
                 var pathGenerator = fc.svg.bar()
-                    .width(width);
+                    .x(0)
+                    .y(0)
+                    .width(width)
+                    .height(0);
 
                 var x = function(d, i) { return xValueScaled(d, i); },
                     y0 = function(d, i) { return y0Value(d, i); },
                     barTop = function(d, i) { return yScale(y0(d, i) + y1Value(d, i)); },
                     barBottom = function(d, i) { return yScale(y0(d, i)); };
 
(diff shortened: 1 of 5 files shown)


src/series/bar.js
```diff
@@ -34,14 +34,15 @@
...
-
                     pathGenerator.x(d3.functor(0))
                         .y(d3.functor(0))
                         .height(function() { return barBottom - barTop; });
 
-                    g.select('path')
+                    var barGroup = d3.select(this);
+
+                    d3.transition(barGroup)
+                        .attr('transform', 'translate(' + x + ', ' + barTop + ')')
```
jleft: I think the enter transitions would look a bit nicer if this translation was also applied to the [enter selection](https://github.com/ScottLogic/d3-financial-components/pull/376/files#diff-8eef6e441ba0cee7bd1c8bde8fbba961L22), but it's not a big deal for now and something that can be changed later. 

Although, it seems this would be a bit harder to do now that the convenience accessors are encapsulated within the `g.each`?

ColinEberhardt:
I'll take a look at this one shortly, I want to grab a copy of the code and play around with it first!

jleft:
> I'll take a look at this one shortly, I want to grab a copy of the code and play around with it first!

Sure, sounds good :smiley:

ColinEberhardt:
Good work, I do agree that the enter transition could be better - having the bars fly in from the top left corner does look a bit odd.

I've quickly hacked a solution that creates a zero height bar on the y0 line within the enter selection:

    var bar = function(selection) {
        selection.each(function(data) {
            var container = d3.select(this);
            var g = fc.utilities.simpleDataJoin(container, 'bar', data, xValue);

            var width = barWidth(data.map(xValueScaled));

            var pathGenerator = fc.svg.bar()
                .width(width);

            // enter
            g.enter()
                .each(function(d, i) {
                    var x = xValueScaled(d, i),
                        y0 = y0Value(d, i),
                        barBottom = yScale(y0);

                    d3.select(this)
                        .attr('transform', 'translate(' + x + ', ' + barBottom + ')');
                })
                .append('path')
                .each(function(d, i) {
                    pathGenerator.x(d3.functor(0))
                        .y(d3.functor(0))
                        .height(function(d, i) { return 0; });

                    var path = d3.select(this);
                    path.attr('d', pathGenerator([d]));
                });


            g.each(function(d, i) {

                var x = xValueScaled(d, i),
                    y0 = y0Value(d, i),
                    barBottom = yScale(y0),
                    barTop = yScale(y0 + y1Value(d, i));

                pathGenerator.x(d3.functor(0))
                    .y(d3.functor(0))
                    .height(function() { return barBottom - barTop; });

                var barGroup = d3.select(this);

                d3.transition(barGroup)
                    .attr('transform', 'translate(' + x + ', ' + barTop + ')')
                    .select('path')
                    .attr('d', pathGenerator([d]));
            });

            decorate(g);
        });
    };

Ignoring the slightly ugly code duplication, the transitions do look quite a bit better. 

What do you think? Is this worth cleaning up and adding to your changes?

Its also worth thinking about how this change would be rolled out to the other series types.

ColinEberhardt:
Also, based on the usage above, I think all of the properties of the svg 'bar' path should be auto-functors, not just the width:

https://github.com/ScottLogic/d3-financial-components/blob/master/src/svg/bar.js#L57

jleft:
> What do you think? Is this worth cleaning up and adding to your changes?

Thanks for looking into this. It definitely looks a lot better to me :smiley: . I've added the changes in 232eb37.

I have slightly mixed feelings about entering the bar with 0 height, rather than just fading it in with its full height. How easily can this be applied to other series for consistency (or maybe consistency isn't important in this case)? Would this approach imply a change in value from 0 to _d_, rather than just a new value, _d_? Either way, it's fairly minor, I like how it looks and would be happy to see this merged as is.

The tests are failing at the moment (although I don't think there's anything wrong with the implementation); if we're happy with these changes, I'll fix the tests before merging.

> Its also worth thinking about how this change would be rolled out to the other series types.

Absolutely, I'll give this some more thought.

> Also, based on the usage above, I think all of the properties of the svg 'bar' path should be auto-functors, not just the width.

I had the same thought! (b546268)

src/series/bar.js
```diff
@@ -18,30 +18,41 @@
...
+                        .height(function() { return barBottom(d, i) - barTop(d, i); });
 
-                    pathGenerator.x(d3.functor(0))
-                        .y(d3.functor(0))
-                        .height(function() { return barBottom - barTop; });
+                    var barGroup = d3.select(this);
 
-                    g.select('path')
+                    d3.transition(barGroup)
+                        .attr('transform', 'translate(' + x(d, i) + ', ' + barTop(d, i) + ')')
```
chrisprice: Did you consider a scale transition?

jleft: Only briefly, I felt translate was more idiomatic, but maybe scale is worth considering further? Would there be a benefit to using scale?

chrisprice: I was thinking it could avoid the double call to the generator but that
would only be possible if the transform origin was at y0. Could that be
done?

ColinEberhardt: One disadvantage of a scale transform is that it would effect any stroke applied to the bar.  Doesn't sound viable to me.

chrisprice: That's a good point. Translate it is!

ColinEberhardt:
@jleft - it would be great to get this merged - will you have time to fix the tests? Or should I take a look?

jleft:
> @jleft - it would be great to get this merged - will you have time to fix the tests? Or should I take a look?

Sorry for the delay - I've fixed the tests.

I would definitely appreciate any feedback on them; they are fairly brittle, especially as I can't see an easy way to encapsulate the value accessors within a single `g.each`, which is a pattern the tests seemed to assume would be used. I've tried to keep them in a similar style to the previous tests and the other series test, although I feel they could benefit from being split up. Please feel free to change them though! :-)

ColinEberhardt:
Many thanks for the updates. I think the problem is that the tests are too tightly coupled to the specif

PR #379, opened 2015-05-24
Fix x-axis in linearTimeSeries

jleft:
The linear time series component was appending the x-axis to the `container` (passed into the component), rather than the `mainContainer` (created by the component); this resulted in the x-axis being added outside of the chart's SVG:

screen shot 2015-05-24 at 16 34 02

This pull request appends the x-axis to the appropriate container - `mainContainer` - (so it can be positioned and sized properly, relative to its parent) and adjusts the plot area and y axis container positioning.

### DIFF
--- a/src/chart/linearTimeSeries.js
+++ b/src/chart/linearTimeSeries.js
@@ -48,7 +48,7 @@
                         position: 'absolute',
                         top: 0,
                         right: 0,
-                        bottom: 0,
+                        bottom: 20,
                         left: 0
                     });
 
@@ -61,15 +61,21 @@
                         position: 'absolute',
                         top: 0,
                         right: 0,
-                        bottom: 0
+                        bottom: 20
                     });
 
-                var xAxisContainer = container.selectAll('g.x-axis')
+                var xAxisContainer = mainContainer.selectAll('g.x-axis')
                     .data([data]);
                 xAxisContainer.enter()
                     .append('g')
                     .attr('class', 'axis x-axis')
-                    .layout('height', 20);
+                    .layout({
+                        height: 20,
+                        position: 'absolute',
+                        right: 0,
+                        bottom: 0,
+                        left: 0
+                    });
 
                 container.layout();
 
--- a/visual-tests/src/test-fixtures/chart/linearTimeSeries.js
+++ b/visual-tests/src/test-fixtures/chart/linearTimeSeries.js
@@ -32,11 +32,6 @@
     chart.plotArea(multi);
 
     d3.select('#linear-time-series')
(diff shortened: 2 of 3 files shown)


ColinEberhardt:
Well spotted - thanks for the fix

chrisprice:
Sorry @jleft, I believe this PR should be reverted.

> The linear time series component was appending the x-axis to the container (passed into the component), rather than the mainContainer (created by the component); this resulted in the x-axis being added outside of the chart's SVG.

This was intentional. As a general rule, our components should not create container elements. The mainContainer svg element was intended as the first row of a two row flex layout. It is an svg element so that overflow hidden can be set on it rather than using a clipmask. The x-axis should render as the second row and it's labels should overflow as the parent container sees fit. 

> This pull request appends the x-axis to the appropriate container - mainContainer - (so it can be positioned and sized properly, relative to its parent) and adjusts the plot area and y axis container positioning.

Was there actually a layout problem or is this just an explanation of your changes?

jleft:
Short answer: sorry for the misunderstanding - that all seems fine, happy to revert :smiley:. An SVG element, to be used as a container, is missing from the multi series visual test and needs to be added.

---

Longer answer:

> This was intentional. As a general rule, our components should not create container elements.

Agreed. I wasn't sure if an exception to the rule might be the `fc.chart` components - if a component is creating an entire chart, one might want the containing SVG element (which is essential for the chart) to be created too - either way is fine to me.

> The mainContainer svg element was intended as the first row of a two row flex layout. It is an svg element so that overflow hidden can be set on it rather than using a clipmask. The x-axis should render as the second row and it's labels should overflow as the parent container sees fit.

Okay, that makes sense - happy to go with this approach and revert.

> Was there actually a layout problem or is this just an explanation of your changes?

I observed that the axis in the multi series visual test didn't look right - the domain line and ticks weren't visible and the tick labels weren't positioned correctly:

lineartimeseries-broken-axis

When I inspected, the structure was:

```html
<div> <!-- container -->
  <svg></svg>
  <g></g> <!-- x-axis container -->
</div>
```

Initially I thought the x-axis container `g` was appended to the incorrect container, hence this pull request. However, it just seems the test was missing a containing SVG element.

chrisprice:
Cool, I'll stick a new PR in to sort that out and revert these changes.

jleft:
> Cool, I'll stick a new PR in to sort that out and revert these changes.

Thanks, sorry for mix-up! (Also, more than happy to create a PR to sort this out, as I caused this.)

chrisprice:
No worries, #381 contains the revert and a little more.

PR #380, opened 2015-05-24
Fix var names in bar series and line series examples

### LINKED ISSUE #375: small typo in example site
in [line series example](http://scottlogic.github.io/d3-financial-components/components.html#line) refers to 'point' both in the code and description (instead of 'line'). Also in bar series code below, it has `var point` (instead of `var bar`)

jleft:
Fixes #375


ColinEberhardt:
Thanks :-)

PR #381, opened 2015-05-26
Revert the changes made in #379

chrisprice:
See the discussion on #379. This takes it a little further and fixes some minor alignment and z-index issues by tidying up the generated structure.

I would consider db1595f an optional enhancement, I was extracting the value as a constant and thought I may as well make it configurable in case someone wants a crazy font-size.


ColinEberhardt:
Looks good to me, and adding the axis height configuration gets a :+1: from me too

Issue #369, opened 2015-05-19
Examples on the website could do with some padding


ColinEberhardt:
via Rob "One thing I noticed though was that all the “bar” charts (OHLC, candlestick, bar, etc) cut the last bar in half."

chrisprice:
Fixed in #383.

PR #383, opened 2015-05-26
Update homepage chart to fully show last bar

chrisprice:
Fixes one of the #369 issues, not sure whether we want to fix all of the component samples in the same way?

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,13 +1,23 @@
 ---
 layout: default
 example-code: |
-  // create some test data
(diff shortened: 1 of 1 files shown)


ColinEberhardt:
Looks good to me.

I'd vote against adding this logic into all of the components samples. I think people should understand that the rendered extent of some series types is beyond that which is implied by the scale domain, and that it is their job to compensate for this (if they wish).

chrisprice:
Cool, happy with that (less work!)

PR #385, opened 2015-05-29
Series accessor tests are now more flexible

ColinEberhardt:
See comment here: https://github.com/ScottLogic/d3-financial-components/pull/376#issuecomment-106623087

The tests only ensure that the first arg of an accessor is an element from the data array and that the second is an integer. They are no longer tightly coupled to the order and number of invocations of the accessors.

### DIFF
--- a/tests/beforeEachSpec.js
+++ b/tests/beforeEachSpec.js
@@ -9,8 +9,8 @@ beforeEach(function() {
                 .forEach(function(call, i) {
                     // check that the first argument is always an element from the data array
                     expect(data.indexOf(call.args[0])).not.toEqual(-1);
-                    // check that the second argument is awlays an integer
-                    expect(call.args[1]).toEqual(parseInt(Number(call.args[1])));
+                    // check that the second argument is always an integer
+                    expect(call.args[1]).toEqual(Math.round(call.args[1]));
                 });
         }
     };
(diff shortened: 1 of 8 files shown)


tests/beforeEachSpec.js
```diff
@@ -0,0 +1,17 @@
...
+    this.utils = {
+        // verifies that each time a property accessor is called that the first arg is a datapoint
+        // from the given array of data, and that the second arg is an integer
+        verifyAccessorCalls: function(spy, data) {
+            spy.calls.all()
+                .forEach(function(call, i) {
+                    // check that the first argument is always an element from the data array
+                    expect(data.indexOf(call.args[0])).not.toEqual(-1);
+                    // check that the second argument is awlays an integer
+                    expect(call.args[1]).toEqual(parseInt(Number(call.args[1])));
```
jleft: Is it worth specifying the radix for parseInt?

jleft:
Really like this! :+1:

tests/beforeEachSpec.js
ColinEberhardt: Good point, just read the MDN page, I had no idea `parseInt` was quite so quirky in its behaviour!

chrisprice: Or use Math.round to be more explicit about what you're doing?

chrisprice:
LGTM 

Happy to see that no one else will have to debug the accessors!

PR #392, opened 2015-05-31
Fix comparison visual test - update namespace for percentage change algorithm

jleft:
Updated the namespace to reflect change in 332d9efd770860846106d7adedfe83bfe2d6b747.


ColinEberhardt:
Thanks for spotting this.

PR #394, opened 2015-06-01
Permit nested multi-series

### LINKED ISSUE #393: multi-series cannot be nested
When used in a nested context, the initial selection for `g.multi-outer` will match both the inner and outer multi-series.

ColinEberhardt:
Fixes #393

### DIFF
--- a/src/series/multi.js
+++ b/src/series/multi.js
@@ -16,10 +16,9 @@
 
                 // in order to support nested multi-series, this selector
                 // is filtered to only return immediate children of the container
-                var self = this;
                 var g = container.selectAll('g.multi-outer')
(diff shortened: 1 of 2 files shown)


src/series/multi.js
```diff
@@ -14,7 +14,13 @@
 
                 var container = d3.select(this);
 
+                // in order to support nested multi-series, this selector
+                // is filtered to only return immediate children of the container
+                var self = this;
```
chrisprice: http://stackoverflow.com/questions/20569670/d3-selector-for-immediate-children

chrisprice: Doh, that won't work :(

Issue #370, opened 2015-05-19
version information in fc.js has not been updated


ColinEberhardt:
https://github.com/ScottLogic/d3-financial-components/blob/master/src/fc.js#L5

This version has been updated locally, but for some reason wasn't included in the 0.1.1 commit: https://github.com/ScottLogic/d3-financial-components/commit/833eb424ce55b840a7c34448284d2d76b73dfe8c

chrisprice:
The release task currently runs the build (which applies the version) before bumping the version, we should probably make use of the afterBump option in grunt-release. However, watch out for geddski/grunt-release#118.

chrisprice:
There's a PR pending which should allow us to use the postBump hook. I've had a quick look around the other release plugins and they're using the workaround in the PR so I think we just wait for that to land.

chrisprice:
Updated the gruntfile to work with the new version when it's ready (https://github.com/chrisprice/d3-financial-components/tree/370)

chrisprice:
Fixed as part of #431 by removing the grunt task :)

PR #395, opened 2015-06-01
Removed version number from fc.js

ColinEberhardt:
No version number is better than the wrong version number! see: #370
I've left the versioning grunt task in there as a reminder to re-instate this code once the related issue is resolved.


chrisprice:
Fair point!

PR #398, opened 2015-06-02
Bar series now correctly applied the y0 value

### LINKED ISSUE #397: bar series incorrectly computes y0 value
if you specify a non-zero y0 value the whole bar series is offset rather than just the baseline.

ColinEberhardt:
Previously y0 was been included in teh bar top and bar bottom calculation, as a result, the two y0's were cancelled out when computng bar height. Doh!

Fixes #397

### DIFF
--- a/src/series/bar.js
+++ b/src/series/bar.js
@@ -27,24 +27,24 @@
                     .height(0);
 
                 var x = function(d, i) { return xValueScaled(d, i); },
-                    barTop = function(d, i) { return yScale(y1Value(d, i)); },
-                    barBottom = function(d, i) { return yScale(y0Value(d, i)); };
+                    y1 = function(d, i) { return yScale(y1Value(d, i)); },
+                    y0 = function(d, i) { return yScale(y0Value(d, i)); };
 
                 pathGenerator.height(0);
 
                 g.enter()
                     .attr('transform', function(d, i) {
-                        return 'translate(' + x(d, i) + ', ' + barBottom(d, i) + ')';
+                        return 'translate(' + x(d, i) + ', ' + y0(d, i) + ')';
                     })
                     .append('path')
                     .attr('d', function(d) { return pathGenerator([d]); });
 
                 g.each(function(d, i) {
-                    pathGenerator.height(barBottom(d, i) - barTop(d, i));
+                    pathGenerator.height(y0(d, i) - y1(d, i));
 
                     var barGroup = d3.select(this);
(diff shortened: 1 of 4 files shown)


src/series/bar.js
```diff
@@ -27,9 +27,8 @@
                     .height(0);
 
                 var x = function(d, i) { return xValueScaled(d, i); },
-                    y0 = function(d, i) { return y0Value(d, i); },
-                    barTop = function(d, i) { return yScale(y0(d, i) + y1Value(d, i)); },
-                    barBottom = function(d, i) { return yScale(y0(d, i)); };
+                    barTop = function(d, i) { return yScale(y1Value(d, i)); },
```
chrisprice: Minor point but should barTop and barBottom now be renamed y0 and y1?

visual-tests/src/test-fixtures/series/bar.js
```diff
@@ -25,18 +25,32 @@
...
-        .yScale(priceScale);
+        .yScale(priceScale1);
 
     // Add it to the chart
     container.append('g')
         .datum(data)
-        .call(bar);
+        .call(bar1);
+
+    // Create scale for y axis
```
chrisprice: You might have avoided this intentionally but I think you could just reuse the price scale with a new range?

src/series/bar.js
ColinEberhardt: Good point - the naming is probably a left-over from when the bar series didn't support negative values.

visual-tests/src/test-fixtures/series/bar.js
ColinEberhardt: I don't know ... mutating and re-using a scale feels wrong. I know it works, but just doesn't feel right!

jleft: If you haven't already seen, D3's scales have [`.copy()` ](https://github.com/mbostock/d3/wiki/Quantitative-Scales#linear_copy), which:

> Returns an exact copy of this linear scale. Changes to this scale will not affect the returned scale, and vice versa.

So you could do something like this (which looks quite neat to me):

```js
var priceScale2 = priceScale1.copy()
    .range([height / 2, height]);
```

ColinEberhardt: > If you haven't already seen, D3's scales have .copy()

I'd forgotten about that, thanks for the reminder. Interesting that D3 scales have a copy function, whereas most other components do not.

PR #399, opened 2015-06-03
MACD indicator

ColinEberhardt:
Added a MACD indicator:

screen shot 2015-06-03 at 06 30 17

This also modified bar series so that it doesn't barf when supplied with undefined values. However, this behaviour should be re-visited in #346


src/series/bar.js
```diff
@@ -26,12 +33,12 @@
                     .width(width)
                     .height(0);
 
+                pathGenerator.height(0);
```
jleft: Does the height need to be set here, if it's already done on line 34?

ColinEberhardt: Probably not - well spotted :+1:

chrisprice: I'm really interested to have a proper look at this but it's going to be
tomorrow afternoon at the earliest before I get a chance :(

ColinEberhardt: No worries, I've got some other things I want to be deplore in parallel.

And good luck for tomorrow!

chrisprice: I didn't realise my reviews had such a reputation!! :)

Thanks!

ColinEberhardt: Serendipitous auto correct?

ColinEberhardt: If that's the right word? I should probably google it to check!

visual-tests/src/test-fixtures/indicators/macd.js
```diff
@@ -0,0 +1,50 @@
...
+
+    // Create the macd algorithm
+    var macdAlgo = fc.indicators.algorithms.macd()
+        .fastPeriod(4)
+        .slowPeriod(10)
+        .signalPeriod(5)
+        .value(function(d) { return d.open; });
+    macdAlgo(data);
+
+    // compute the extents
```
chrisprice: This extent calculation looks funky but I think that’s only because the scale is misnamed?
Also, could be simplified to something like priceScale.domain(d3.max(extent, Math.abs))?

src/indicators/renderers/macd.js
```diff
@@ -0,0 +1,67 @@
...
+        var macdLine = fc.series.line(),
+            signalLine = fc.series.line(),
+            divergenceBar = fc.series.bar(),
+            multiSeries = fc.series.multi();
+
+        var macd = function(selection) {
+
+            macdLine
+                .xValue(xValue)
+                .yValue(function(d, i) { return d.macd.macd; })
```
chrisprice: Is it worth adding an intermediary accessor for the root data value? i.e. d.macd.signal would be (ignore naming) root(d).signal

src/indicators/renderers/macd.js
```diff
@@ -0,0 +1,67 @@
...
+            multiSeries
+                .xScale(xScale)
+                .yScale(yScale)
+                .series([divergenceBar, macdLine, signalLine]);
+
+            selection.each(function(data) {
+                d3.select(this)
+                    .append('g')
+                    .classed('macd-indicator', true)
+                    .datum(data)
```
chrisprice: I don't think this is required.

chrisprice:
Other than those comments, :+1: from me

visual-tests/src/test-fixtures/indicators/macd.js
ColinEberhardt: Good idea, not quite right but did get me thinking, hadn't realised `d3.max` has an accessor function. This simplifies things quite a bit:

    var maxExtent = d3.max(data, function(d) { return Math.abs(d.macd.macd); });
    priceScale.domain([-maxExtent, maxExtent]);

src/indicators/renderers/macd.js
ColinEberhardt: Good point :+1:

src/indicators/renderers/macd.js
ColinEberhardt: Do you mean as part of the components public API? Or an internal convenience?

ColinEberhardt:
> Also rebind xValue?

Done - and also exposed xValue for macd also (boll already exposed this). I have decided against exhaustively rebinding all the properties of the nested series. Indicators renderers are opinionated ;-)

chrisprice:
Just the one comment this time :)

ColinEberhardt:
> xValue is rebound but there's also a local property for it?

Well spotted, it's gone now (and tested this time)

chrisprice:
:+1:

PR #402, opened 2015-06-05
Linear time series plot area correctly clipped

### LINKED ISSUE #401: linear time series does not clip the plot area
Here's an area chart 'bleeding' out from the plot area:

screen shot 2015-06-05 at 07 57 01

ColinEberhardt:
fixes #401


chrisprice:
Good spot, side effect of the changes in my last PR #381 I think. :+1:

PR #403, opened 2015-06-05
Add cycle plot series

chrisprice:
Not sure how appropriate it is in the series package but it was the closest match I could find.

Demonstrates a workaround for https://github.com/ScottLogic/d3-financial-components/issues/400.

For reference, [this](http://www.cvast.tuwien.ac.at/sites/default/files/cycleplot-technique.png) is what a cycle plot looks like and [here's one](http://bl.ocks.org/godds/6132680) @godds did in d3.


ColinEberhardt:
I've been playing around with this and really like it. However, it took a while to get an example up and running, quite a few scales and value accessors to get right!

This series will benefit from a clear and simple example on the website. 

I re-implemented Graham's example using the cycle series, which went pretty well:

    d3.json('data.json', function(error, data) {
        var parseDate = d3.time.format('%Y%m%d').parse;

        // parse the data / coerce types
        data.forEach(function(d) {
          d.date = parseDate(d.date);
          d.itemsSold = +d.items_sold;
        });

        var dayScale = d3.scale.linear()
            .domain([-0.5, 6.5])
            .range([0, width]);

        var salesScale = d3.scale.linear()
            .domain(fc.utilities.extent(data, 'itemsSold'))
            .range([height, 0])
            .nice();

        var subScale = d3.scale.linear()
            .domain(fc.utilities.extent(data, 'date'));

        var lineSeries = fc.series.line()
            .xValue(function(d) { return d.date; })
            .yValue(function(d) { return d.itemsSold; });

        var cycle = fc.series.cycle()
            .barWidth(fc.utilities.fractionalBarWidth(0.9))
            .yScale(salesScale)
            .xScale(dayScale)
            .xValue(function(d) { return d.date.getDay(); })
            .subScale(subScale)
            .subSeries(lineSeries);

        container.append('g')
            .datum(data)
            .call(cycle);
    });

Although as soon as I added the 'baseline', the average price for each day of the week, it became very ugly! It is a real pain having to repeat the same nesting that the cycle series performs.

This functionality could be baked in to the cycle series, or perhaps added as a form of axis?

chrisprice:
Yea I agree, doing the algorithmic work outside it starts to feel a bit messy like the stacked bar. I toyed with the idea of using an axis too, but it felt wrong to force it into the multi. Thinking about it again though, cycle could explicitly accept an axis e.g. ```subAxis```?

ColinEberhardt:
I'm trying it it the other way around, an axis that accepts a cycle series. I'll see how it turns out.

chrisprice:
I've opted to create an adapter for the axis which allows it to be used as a series. This is then added to the multi series used as the cycle plot's subSeries.

This approach seems to offer the most potential for re-use whilst cleaning up the visual test.

What process are we following for adding stuff to the website? It would be good to have a new build containing the functionality before attempting the example...

ColinEberhardt:
Much better :+1:

PR #405, opened 2015-06-09
Refactor sdj into dataJoin and use it to remove the nested element fr…

### LINKED ISSUE #366: Document the rationale behind simpleDataJoin
The comments currently cover the implementation but not why we have it. The description should cover decorate and transitions.

chrisprice:
…om multi series

Fixes #366.

I wanted to get feedback before rolling out the new data-join pattern.

### DIFF
--- a/src/series/multi.js
+++ b/src/series/multi.js
@@ -8,40 +8,39 @@
             series = [],
             mapping = fc.utilities.fn.identity;
 
+        var dataJoin = fc.utilities.dataJoin()
+            .children(true)
+            .selector('g.multi')
+            .element('g')
+            .attrs({'class': 'multi'})
+            .key(function(d) { return d.__series__; });
+
         var multi = function(selection) {
 
             selection.each(function(data) {
 
-                var container = d3.select(this);
+                // Prototypically inherit the mapped data for a series and augment the object
+                // with a series property. This allows us to data-bind without requiring a nested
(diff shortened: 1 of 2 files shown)


ColinEberhardt:
Might be worth showing a simple example that compares D3 and D3FC data joins:

```
// -------------  a regular datajoin
var container = d3.select('#container');

var update = container.selectAll('li')
  .data(data);

update.enter()
  .append('li')
  .text(function(d) { return d; } );

// the problem with this is how do you allow someone to customise 
// the enter and update selections?

// -------------  a D3FC datajoin
var container2 = d3.select('#container2');

var dataJoin = fc.utilities.dataJoin()
  // the selection used to construct the datajoin
  .selector('li') 
  // the element appended in the enter selection
  .element('li');

var updateSelection = dataJoin(container2, data);

// updatedSelection has already appended the 'li' element,
// allowing the result to be decorated

// enter
updateSelection.enter()
  .append('span')
  .text(function(d) { return d; } );

// update
updateSelection.style('color', 'red');
```

chrisprice:
I think it's only going to be advanced users of the library who are interested in the difference between then two. Hopefully the comment I've just added along with the one at the top of the file is enough for them to appreciate the difference.

ColinEberhardt:
secret sauce is good :sake:

PR #407, opened 2015-06-09
Annotation refactor

ColinEberhardt:
The line annotation creates the following structure:

screen shot 2015-06-09 at 15 31 12

This makes it easier to add ' handles' or other decoration on the left or right-hand size of the annotation.

### DIFF
--- a/src/tools/horizontalLineAnnotation.js
+++ b/src/tools/horizontalLineAnnotation.js
@@ -1,7 +1,7 @@
 (function(d3, fc) {
     'use strict';
 
-    fc.tools.annotation = function() {
+    fc.tools.horizontalLineAnnotation = function() {
 
 
         var xScale = d3.time.scale(),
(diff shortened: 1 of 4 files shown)


src/tools/horizontalLineAnnotation.js
```diff
@@ -0,0 +1,100 @@
...
+                var xScaleRange = xScale.range(),
+                    y = function(d) { return yScale(yValue(d)); };
+
+                var container = d3.select(this);
+
+                // Create a group for each horizontalLineAnnotation
+                var g = fc.utilities.simpleDataJoin(container, 'annotation', data, keyValue);
+
+                // create the outer container and line
+                var enter = g.enter();
```
chrisprice: Missing a g.attr('transform', 'translate(' + xScaleRange[0] + ',0)');?

src/tools/horizontalLineAnnotation.js
```diff
@@ -0,0 +1,100 @@
...
+                enter.append('g')
+                    .classed('left-handle', true);
+                // create a right container with a text label
+                enter.append('g')
+                    .classed('right-handle', true)
+                    .attr('transform', function(d) {
+                        return 'translate(' + (xScaleRange[1] - xScaleRange[0]) + ', 0)';
+                    })
+                    .append('text')
+                    .attr('transform', 'translate(-5, -5)');
```
chrisprice: You can set x/y directly on text nodes

src/tools/horizontalLineAnnotation.js
```diff
@@ -0,0 +1,100 @@
...
+                var enter = g.enter();
+                enter.append('line')
+                    .attr('x2', xScaleRange[1] - xScaleRange[0]);
+                // create an empty left container
+                enter.append('g')
+                    .classed('left-handle', true);
+                // create a right container with a text label
+                enter.append('g')
+                    .classed('right-handle', true)
+                    .attr('transform', function(d) {
```
chrisprice: Don't think the function is needed.

src/tools/horizontalLineAnnotation.js
```diff
@@ -0,0 +1,100 @@
+(function(d3, fc) {
+    'use strict';
+
+    fc.tools.horizontalLineAnnotation = function() {
```
chrisprice: Drop annotation from the name?

PR #408, opened 2015-06-09
Add axis series visual test

chrisprice:
Realised we didn't have one and the implementation was nice and simple :)


ColinEberhardt:
It's pretty cool that axes can now be added directly to a chart via multi-series.

PR #409, opened 2015-06-10
line annotation now supports horizontal and vertical

ColinEberhardt:
Unfortunately the rename and changes mean git sees this as a new file. Perhaps I should have committed the rename first.

Anyhow, this change allows line annotations to support vertical and horizontal orientation.


src/tools/line.js
```diff
@@ -0,0 +1,137 @@
+(function(d3, fc) {
+    'use strict';
+
+    fc.tools.line = function() {
+
+        var xScale = d3.time.scale(),
+            yScale = d3.scale.linear(),
+            value = fc.utilities.fn.identity,
+            keyValue = fc.utilities.fn.index,
```
ColinEberhardt: keyValue property is unused

chrisprice:
Opps... didn't realise you'd split this into a separate PR. Merged as part of the #412 anyway.

PR #412, opened 2015-06-11
Crosshair refactor

ColinEberhardt:
Builds on #409 

A number of crosshair changes:
 + It now uses two line annotations to render the 'cross'
 + It constructs a container at the crosshair origin and adds a circle - this makes it much easier to create a 'trackball'
 + Line annotation and crosshair use the new `dataJoin` component
 + Line annotation has a new `domainValue` property that indicates whether the value is in domain or screen coordinates
 + Snapping now support X only and Y only


src/tools/crosshairs.js
```diff
@@ -9,14 +9,27 @@
...
+            decorate = fc.utilities.fn.noop;
 
         var x = function(d) { return d.xInDomainUnits ? xScale(d.x) : d.x; },
             y = function(d) { return d.yInDomainUnits ? yScale(d.y) : d.y; };
 
+        var dataJoin = fc.utilities.dataJoin()
+            .children(true)
+            .selector('g.crosshair')
+            .element('g')
+            .attrs({'class': 'crosshair'});
```
chrisprice: Created #420 to simplify this in the future.

src/tools/line.js
```diff
@@ -0,0 +1,166 @@
...
+    'use strict';
+
+    fc.tools.line = function() {
+
+        var xScale = d3.time.scale(),
+            yScale = d3.scale.linear(),
+            value = fc.utilities.fn.identity,
+            keyValue = fc.utilities.fn.index,
+            label = value,
+            // specifies whether the value is domain or screen coordinates
```
chrisprice: I know this pattern came from crosshairs but I now think I would expect this to be handled by supplying identity scales rather than a flag. Otherwise I'm a bit scared that over time this flag will end up being propagated everywhere the scale is. Happy to see this spun off as a separate issue though.

chrisprice: On second thoughts, the fix is just a ternary operator in the x/yScale calls in crosshairs. Maybe it can be fixed as part of this?

src/tools/crosshairs.js
```diff
@@ -45,48 +58,35 @@
...
+                    .append('g')
+                    .classed('trackball', true)
+                    .attr('transform', trackballTranslate)
+                    .append('circle')
+                    .attr('r', 5);
+
+                crosshair.select('g.trackball')
+                    .attr('transform', trackballTranslate);
+
+                if (data.length === 1) {
```
chrisprice: ...remove this bit...

src/tools/crosshairs.js
```diff
@@ -45,48 +58,35 @@
...
+                    verticalLine.domainValue(data[0].xInDomainUnits);
+                    horizontalLine.domainValue(data[0].yInDomainUnits);
+                }
+
+                var multi = fc.series.multi()
+                    .series([horizontalLine, verticalLine])
+                    .xScale(xScale)
```
chrisprice: ...and use identity scales.

src/tools/crosshairs.js
```diff
@@ -9,14 +9,27 @@
...
             y = function(d) { return d.yInDomainUnits ? yScale(d.y) : d.y; };
 
+        var dataJoin = fc.utilities.dataJoin()
+            .children(true)
+            .selector('g.crosshair')
+            .element('g')
+            .attrs({'class': 'crosshair'});
+
+        var horizontalLine = fc.tools.line()
+            .value(function(d) { return d.y; })
```
chrisprice: I think you can simplify this by using the x function from above...

src/tools/crosshairs.js
```diff
@@ -9,14 +9,27 @@
...
+            .label(function(d) { return d.y; });
+
+        var verticalLine = fc.tools.line()
+            .orient('vertical')
+            .value(function(d) { return d.x; })
```
chrisprice: ...and the y function here...

src/tools/line.js
chrisprice: On third thoughts, see comments inline in crosshairs :)

chrisprice: Created #421.

src/tools/line.js
```diff
@@ -0,0 +1,166 @@
...
+                            transform = valueScale(transform);
+                        }
+                        return translation(scaleRange[0], transform);
+                    },
+                    scaleWidth = scaleRange[1] - scaleRange[0];
+
+                var container = d3.select(this);
+
+                // Create a group for each line
+                data = Array.isArray(data) ? data : [data];
```
chrisprice: Not sure what this is protecting against?

src/tools/line.js
```diff
@@ -0,0 +1,166 @@
...
+                enter.append('g')
+                    .classed(handleOne, true);
+
+                enter.append('g')
+                    .classed(handleTwo, true)
+                    .attr('transform', translation(scaleWidth, 0))
+                    .append('text')
+                    .attr(textAttributes);
+
+                // Update
```
chrisprice: By reading the update will not handle crossScale range changes. Is that intentional?

ColinEberhardt: No, it was not :-) I've added the required code to the update selection (and kept it in enter selection for better enter transitions).

ColinEberhardt:
@chrisprice I think that's everything.

chrisprice:
LGTM :+1:

PR #413, opened 2015-06-11
linear time series now rebinds more axis properties

ColinEberhardt:
I've found myself needing to use quite a few non-rebound properties, so I've just re-bound the lot!


chrisprice:
:+1: 

Good catch with the throw, assume that caught you out? :)

PR #416, opened 2015-06-15
Legend component

ColinEberhardt:
Added a new legend component.

The legend renders a HTML table, based on an array of items:

    var legend = fc.charts.legend()
            .items([
                ['open', function(d) { return priceFormat(d.open); }],
                ['high', function(d) { return priceFormat(d.high); }],
                ['low', function(d) { return priceFormat(d.low); }],
                ['close', function(d) { return priceFormat(d.close); }]
            ]);

Each item within the array is an array of two values, the label and the value. Both are 'functored' allowing functions to be used for the label as well as the value.

Example:

screen shot 2015-06-15 at 11 56 38

The decorate function is passed a selection with each 'tr' element within the table.

### DIFF
--- a/src/chart/chart.css
+++ b/src/chart/chart.css
@@ -12,3 +12,12 @@
 .sparkline .low {
   fill: red;
 }
+
+.legend {
+  border: 1px solid #999;
+  margin: 10px;
+}
+
+.legend .label {
+  font-weight: bold;
+}
--- a/src/chart/chart.css
+++ b/src/chart/chart.css
@@ -15,9 +15,4 @@
 
 .legend {
   border: 1px solid #999;
-  margin: 10px;
-}
-
-.legend .label {
-  font-weight: bold;
 }
--- a/src/chart/legend.js
+++ b/src/chart/legend.js
@@ -1,76 +1,64 @@
 (function(d3, fc) {
     'use strict';
 
-    fc.charts.legend = function() {
-        var decorate = fc.utilities.fn.noop,
-            items = [
+    fc.chart.legend = function() {
+        var tableDecorate = fc.util.fn.noop,
+            rowDecorate = fc.util.fn.noop;
+
+        var items = [
                 ['datum', function(d) { return d.datum; }]
             ];
+
+        var tableDataJoin = fc.util.dataJoin()
+            .selector('table.legend')
+            .element('table')
+            .attrs({'class': 'legend'});
+
+        var rowDataJoin = fc.util.dataJoin()
+            .selector('tr.row')
+            .element('tr')
+            .attrs({'class': 'row'});
 
         var legend = function(selection) {
             selection.each(function(data) {
                 var container = d3.select(this);
 
-                // map the data to include the datapoint, so that
-                // decorators have access to both the 'items' and 'data'
                 var legendData = items.map(function(item, i) {
                     return [
                         {
                             value: d3.functor(item[0]),
                             index: i,
-                            datum: data,
-                            class: 'label'
+                            datum: data
                         },
                         {
                             value: d3.functor(item[1]),
                             index: i,
-                            datum: data,
-                            class: 'value'
+                            datum: data
                         }
                     ];
                 });
 
-                var table = container.selectAll('table')
-                    .data([legendData]);
-                table.enter()
-                    .append('table')
-                    .classed('legend', true);
+                var table = tableDataJoin(container, [legendData]);
 
-                // create the rows
-                var tr = table.selectAll('tr')
-                  .data(function(d) { return d; });
+                var trUpdate = rowDataJoin(table);
 
-                tr.enter()
-                  .append('tr');
-                tr.exit().remove();
+                var trEnter = trUpdate.enter();
+                trEnter.append('th');
+                trEnter.append('td');
 
-                // create the cells
-                var tdUpdate = tr.selectAll('td')
-                  .data(function(d) { return d; });
+                trUpdate.select('th')
+                    .html(function(d) {
+                        return d[0].value(d[0].datum);
+                    });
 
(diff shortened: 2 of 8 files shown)


src/chart/legend.js
```diff
@@ -0,0 +1,78 @@
...
+                    .html(function(d, i) {
+                        return d.label(d.datum, i);
+                    });
+
+                container.selectAll('td.value')
+                    .html(function(d, i) {
+                        return d.value(d.datum, i);
+                    });
+
+                var decorateSelection = container.selectAll('tr');
```
ColinEberhardt: Yuck - I am pretty sure I shouldn't need to do this. However, the `rowsUpdate ` selection above only gets called on enter. Perhaps there is something wrong with the nested datajoin?

ColinEberhardt:
OK, changed my mind about how this should work. In the previous implementation the rows were created via the datajoin, and the cells constructed manually. After further testing I found that this causes a lot of problems (d3 does not propagate changes in `__data__`, so nested elements hold stale data).

Anyhow, I've modified the behaviour so that the rows and cells are constructed via datajoin.

chrisprice:
The other components act on the supplied data, I wonder if we could do that for the legend as well? Something like -

```js
var data = [
    ['open', function(d) { return priceFormat(d.open); }],
    ['high', function(d) { return priceFormat(d.high); }],
    ['low', function(d) { return priceFormat(d.low); }],
    ['close', function(d) { return priceFormat(d.close); }]
];

var legend = fc.charts.legend()
    .label(function(d) { return d[0].apply(this, arguments); })
    .value(function(d) { return d[1].apply(this, arguments); });

selection.datum(data)
    .call(legend);
```

Also was there a restriction in dataJoin which prevented you from using it?

ColinEberhardt:
we could flip it round, and have the configuration data introduced to the legend via the data-join but that feels like doing things the wrong way round to me! Although I can see some advantages, e.g. allowing you to make the legend items dynamic.

Regarding dataJoin, I don't think this work in this case where the UI is constructed via a nested data-join, although I didn't actually give it a try to see if it would work or would be easily adapted / extended.

ColinEberhardt:
@chrisprice - rebased onto master, and updated implementation
 + Now makes use of the `dataJoin` component, which has been tweaked to allow nesting
 + The legend exposes two decorate functions, table and row
 + Decorating on row is more natural, you can just select the th or td to update their contents.

ColinEberhardt:
I also checked whether this component should add `tbody`, but it looks like this is optional as per the HTML5 spec:

> *tr element*
> ... As a child of a table element, after any caption, colgroup, and thead elements, but only if there are no tbody elements that are children of the table element.

http://dev.w3.org/html5/spec-author-view/the-tr-element.html

src/util/dataJoin.js
```diff
@@ -25,6 +25,8 @@
 
         var dataJoin = function(container, data) {
 
+            var joinedData = data || fc.util.fn.identity;
```
chrisprice: Very nice :+1:

src/chart/legend.js
```diff
@@ -0,0 +1,91 @@
...
+            .selector('tr.row')
+            .element('tr')
+            .attrs({'class': 'row'});
+
+        var legend = function(selection) {
+            selection.each(function(data) {
+                var container = d3.select(this);
+
+                var legendData = items.map(function(item, i) {
+                    return [
```
chrisprice: Is it worth just making this an object now?

src/chart/legend.js
```diff
@@ -0,0 +1,91 @@
...
+
+                var trUpdate = rowDataJoin(table);
+
+                var trEnter = trUpdate.enter();
+                trEnter.append('th');
+                trEnter.append('td');
+
+                trUpdate.select('th')
+                    .html(function(d) {
+                        return d[0].value(d[0].datum);
```
chrisprice: Should these be invoked in the normal d3 style (context: node, args: datum, index)?

src/chart/legend.js
```diff
@@ -0,0 +1,91 @@
...
+
+        var legend = function(selection) {
+            selection.each(function(data) {
+                var container = d3.select(this);
+
+                var legendData = items.map(function(item, i) {
+                    return [
+                        {
+                            value: d3.functor(item[0]),
+                            index: i,
```
chrisprice: I think because of the data-join changes you'll be ok to just use the d3 supplied index?

visual-tests/src/site/templates/layouts/index.hbs
```diff
@@ -2,6 +2,7 @@
 <html lang="en">
   <head>
     {{> header }}
+    <link href="{{ assets }}/bootstrap/css/bootstrap.min.css" rel="stylesheet">
```
chrisprice: Did you mean to move this from inclues/header.hbs?

ColinEberhardt: Yes, but the reason for the move is no longer relevant!

ColinEberhardt:
@chrisprice good spot, you're right, `legendData` is still a bit too complex due to the previous implementation. I've simplified it now.

src/chart/chart.css
```diff
@@ -11,4 +11,13 @@
 
 .sparkline .low {
   fill: red;
+}
+
+.legend {
```
chrisprice: Final point: do we need the css?

src/chart/chart.css
```diff
@@ -11,4 +11,13 @@
...
+  border: 1px solid #999;
+  margin: 10px;
+}
+
+.legend .label {
```
chrisprice: I don't think this one is being used in any case.

chrisprice:
Loving the commit message :)

PR #424, opened 2015-06-18
Added a band annotation

ColinEberhardt:
Also removed unused keyValue property within line annotation

The band annotation can be used to plot vertical or horizontal bands. In terms of implementation, it is very similar to the line.

screen shot 2015-06-18 at 23 06 45

I can't think of any suitable 'handles' for this annotation, I can't think of many realistic examples where people would want to add element to the band.

### DIFF
--- a/src/annotation/band.js
+++ b/src/annotation/band.js
@@ -1,0 +1,138 @@
+(function(d3, fc) {
+    'use strict';
+
+    fc.annotation.band = function() {
+
+        var xScale = d3.time.scale(),
+            yScale = d3.scale.linear(),
+            value0 = fc.util.fn.identity,
+            value1 = fc.util.fn.identity,
+            decorate = fc.util.fn.noop,
+            orient = 'horizontal';
+
+        var dataJoin = fc.util.dataJoin()
+            .selector('g.annotation')
+            .element('g')
+            .attrs({'class': 'annotation'});
+
+        function pickFirst(a, d) { return a; }
+        function pickSecond(a, d) { return d; }
+
+        var band = function(selection) {
+            selection.each(function(data) {
+
+                // the value scale which the annotation 'value' relates to, the crossScale
+                // is the other. Which is which depends on the orientation!
+                var valueScale, crossScale, translation, height, width;
+                switch (orient) {
+                    case 'horizontal':
+                        translation = function(a, b) { return 'translate(' + a + ', ' + b + ')'; };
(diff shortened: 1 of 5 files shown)


src/tools/line.js
```diff
@@ -70,7 +69,8 @@
 
                 // create the outer container and line
                 var enter = g.enter()
-                    .attr('transform', containerTransform);
+                    .attr('transform', containerTransform)
+                    .classed(orient, true);
```
chrisprice: Did you mean to remove this line?

src/svg/bar.js
```diff
@@ -57,6 +58,13 @@
             width = d3.functor(x);
             return bar;
         };
+        bar.centred = function(x) {
```
chrisprice: Change to align and support 'left' + 'center'?

src/tools/band.js
```diff
@@ -0,0 +1,138 @@
...
+                // Create a group for each band
+                var g = dataJoin(container, data);
+
+                // create the outer container and band path
+                var enter = g.enter()
+                    .attr('transform', containerTransform);
+                enter.append('path')
+                    .classed('band', true);
+
+                var pathGenerator = fc.svg.bar()
```
chrisprice: Are we getting a lot from using the bar rather than a rect?

chrisprice:
As discussed, maybe an implementation which accepted y0/y1/x0/x1 and defaulted the values to the extent of the scale would make for a simpler and more flexible implementation (i.e. removing orient and allowing it to be used as a background)?

src/tools/band.js
ColinEberhardt: > Are we getting a lot from using the bar rather than a rect?

SVG `rect` doesn't support negative width / height which is a bit of a pain as it adds extra logic wherever you use it. bar supports negative values.

src/annotation/band.js
```diff
@@ -0,0 +1,128 @@
+(function(d3, fc) {
+    'use strict';
+
+    fc.annotation.band = function() {
+
+        // ordinal axes have a rangeExtent function, this adds any padding that
+        // was applied to the range. This functions returns the rangeExtent
+        // if present, or range otherwise
+        function range(scale) {
```
chrisprice: This is turning up everywhere, I've reated an issue for extracting it #437.

chrisprice:
:+1:

PR #429, opened 2015-06-19
Various renames to normalise namespaces

### LINKED ISSUE #387: Double check namespaces
Check for plural vs. non-plural - namespaces should be singular, e.g `fc.chart` as opposed to `fc.charts`

chrisprice:
Fixes #387.

I'm bound to have missed some but hopefully this catches most of them!


chrisprice:
Annotation namespace changes in, should be GTG.

_SpecRunner.html
```diff
@@ -0,0 +1,79 @@
+<!doctype html>
```
ColinEberhardt: we should probably add _SpecRunner to .gitignore ?

ColinEberhardt:
One minor comment, but other than that :+1:

chrisprice:
Good spot, merging...

PR #439, opened 2015-06-24
band unit tests

ColinEberhardt:
Adds unit tests for the band series ... which uncovered a few bugs!

 + literal property values were not being functored
 + context and index were both incorrect

Regarding the second point, any component that uses our svg components are likely to be incorrect too.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -236,9 +236,6 @@ module.exports = function (grunt) {
             },
             test: {
                 src: ['dist/*.js'],
-            },
-            testDebug: {
-                src: ['dist/*.js'],
                 options: {
                     keepRunner: true
                 }
--- a/src/annotation/band.js
+++ b/src/annotation/band.js
@@ -55,7 +55,11 @@
                     });
 
                 g.select('path')
-                    .attr('d', function(d) { return pathGenerator([d]); });
+                    .attr('d', function(d, i) {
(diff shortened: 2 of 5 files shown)


Gruntfile.js
```diff
@@ -236,9 +236,6 @@ module.exports = function (grunt) {
             },
             test: {
                 src: ['dist/*.js'],
-            },
```
ColinEberhardt: I've dropped the testDebug target, now that the spec runner file is ignore, I don't see any harm in keeping this file hanging around. I find it useful for debugging.

src/annotation/band.js
```diff
@@ -55,7 +55,11 @@
                     });
 
                 g.select('path')
```
chrisprice: This problem is rife throughout our use of the `fc.svg.* ` components e.g. [bar series](https://github.com/ScottLogic/d3fc/blob/master/src/series/bar.js#L45). I do think your changes represent the pattern we should be following though, should we create an issue to fix the rest?

chrisprice:
Happy to see this merged but holding off so that you see my question.

src/annotation/band.js
ColinEberhardt: Yep, I'll raise an issue.

ColinEberhardt:
Merged and issue raised.

Issue #444, opened 2015-06-24
Update the website designs


ColinEberhardt:
Update the website design based on Tamara's visuals, @steviefordi is currently working on this.

PR #448, opened 2015-06-25
Implemented new site design.

steviefordi:
This fixes issue #444 to implement new site design.

The new design can be seen on my fork's project page here:
http://steviefordi.github.io/d3-financial-components/


ColinEberhardt:
Thanks @steviefordi - I'll take a look at this shortly

chrisprice:
@steviefordi looks a lot better! 

I'm not sure who this is feedback for (I haven't seen the designs) but...

### Homepage
* [ ] The logo isn't vertically centered in the top left and the left padding doesn't match the top level nav's right padding.
* [ ] I think the colours for the active/inactive pages in the top level nav should be inverted.
* [ ] The hero section isn't centrally vertically aligned.
* We should consider changing the main chart's colour scheme to better fit the site's colour scheme.
* [ ] The padding around the main chart looks inconsistent, not sure what the design proposed though. (If you make the svg display: block, it might make it easier to position).
* [ ] Again unsure of design, but vertical whitespace between the various sections on the main page seems inconsistent.
* [ ] To my eyes there's not enough contrast between the sub-title and body text in the quick start maybe text-transform:uppercase.
* [ ] The quick-start sub-titles and content should be re-visited to try and have consistent lengths (i.e. sub-titles fit on one line, content is similar numbers of lines).
image

* [ ] Whitespace around the D3 + d3fc = Awesome! is very inconsistent. 
* [ ] Consider changing the  body text colour to match the footer background?
* [ ] The example images are Consider making the example box title widths tight to the text rather than being 100%, then maybe align right?
image

### Components
* [ ] The bars between the component names in the nav look odd to me, maybe just stick with newlines between them?
* [ ] In my opinion, the hrs are unnecessary and bleed into the content.
* [ ] The body text feels very heavy, maybe drop the size to 16px? Consider changing the color as well as per above.
* [ ] Use caps for the main headings to match the nav.
* [ ] Content and nav have different top padding.
* [ ] Move the view component source link
* [ ] Pin the title to the top of the page or move the menu up to fill the whitespace as you scroll.

image

### Example

* [ ] Adopt the same layout as Components?

I'm sure some of the above will warrant more discussion but I thought I'd write them all down so we were forced to make a decision on them.

ColinEberhardt:
Thanks for the feedback @chrisprice - I think we'll have to sit down together and work out how to tweak the design to address some of these concerns. Unfortunately Tamara is no longer available to work on the designs themselves.

Anyhow, thanks for your work @steviefordi - I'll work out how to proceed with this.

steviefordi:
Thanks for the feedback @chrisprice, I'm happy to work on anything you and @ColinEberhardt want to change - just let me know. I should have time available for during the next two weeks, but after that it will be a little difficult.

chrisprice:
Superseded by #455

ColinEberhardt:
Closed by #455 - thanks @steviefordi for getting the ball rolling.

Issue #400, opened 2015-06-04
Styling stacked bar series is hard


ColinEberhardt:
The stacked bar series has the following structure:

![screen shot 2015-06-04 at 08 18 54](https://cloud.githubusercontent.com/assets/1098110/7978897/8a6100e0-0a92-11e5-8c67-51a6140be02c.png)

It is a collection of bar series, each with a parent `g.stacked-bar`. To set the color for each series you would expect to be able to decorate as follows:

        var stack = fc.series.stackedBar()
            .decorate(function(sel) {
                sel.attr('fill', function(d, i) {
                    return color(i);
                });
            });

This sets the `fill` attribute for the `g.stacked-bar` elements. However, the fill color for each bar is determined by the CSS selector `.bar>path`. which has a higher specificity.

This can be solved by setting the `fill` style on the path directly:

    .decorate(function(sel) {
        sel.each(function(d, i) {
            d3.select(this)
              .selectAll('path')
              .style('fill', color(i));
        }); 
    });

But the need to 'burrow' into the internal structure is a bit ugly.

ColinEberhardt:
I've been playing around with this and really like it. However, it took a while to get an example up and running, quite a few scales and value accessors to get right!

This series will benefit from a clear and simple example on the website. 

I re-implemented Graham's example using the cycle series, which went pretty well:

    d3.json('data.json', function(error, data) {
        var parseDate = d3.time.format('%Y%m%d').parse;

        // parse the data / coerce types
        data.forEach(function(d) {
          d.date = parseDate(d.date);
          d.itemsSold = +d.items_sold;
        });

        var dayScale = d3.scale.linear()
            .domain([-0.5, 6.5])
            .range([0, width]);

        var salesScale = d3.scale.linear()
            .domain(fc.utilities.extent(data, 'itemsSold'))
            .range([height, 0])
            .nice();

        var subScale = d3.scale.linear()
            .domain(fc.utilities.extent(data, 'date'));

        var lineSeries = fc.series.line()
            .xValue(function(d) { return d.date; })
            .yValue(function(d) { return d.itemsSold; });

        var cycle = fc.series.cycle()
            .barWidth(fc.utilities.fractionalBarWidth(0.9))
            .yScale(salesScale)
            .xScale(dayScale)
            .xValue(function(d) { return d.date.getDay(); })
            .subScale(subScale)
            .subSeries(lineSeries);

        container.append('g')
            .datum(data)
            .call(cycle);
    });

[...]

Issue #445, opened 2015-06-24
Adds visual tests for stacked line and stacked area


janakerman:

ColinEberhardt:
Done.

PR #453, opened 2015-06-28
Fixes #391 - Adds visual tests for stacked bar & line series

### LINKED ISSUE #391: Expand on the different stacked series types we offer
Add 100% stacked series and support more than just bar i.e. line, area...

janakerman:
These are simply duplicates of the stacked bar visual test. Since it is only a visual test I figured simply duplicating the file would suffice. I'd be happy to refactor this if desired.

In the stacked area visual test it isn't possible to see how the series stack (without using the crosshair) due to #400. Once that is fixed, the stacked series tests should be different colours.

Once this PR is merged in I believe we can close off this issue.

### DIFF
--- a/visual-tests/src/test-fixtures/series/stacked-area.hbs
+++ b/visual-tests/src/test-fixtures/series/stacked-area.hbs
@@ -1,0 +1,17 @@
+---
+title: Stacked Area
+description: Data is loaded from a CSV file.
+categories: series
+tags:
+- area
+- stacked area
+- csv
+testScripts:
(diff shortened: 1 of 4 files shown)


janakerman:
In fact, it seems we can close off #391 & #445 as I seem to have raised a separate issue to add the visual tests.

chrisprice:
Happy to see these go in as duplicates as they're only visual tests. I can see in the future quite a bit of the code will probably end up in the library anyway (e.g. snapping).

Thanks!

PR #454, opened 2015-06-29
Fixes #420 - dataJoin attrs accepts key/value params as well as an object

### LINKED ISSUE #420: dataJoin.args should support both object and key/value signatures
Currently only the object form is supported but it is commonly used just for setting a class name like -
```js
  var dataJoin = fc.utilities.dataJoin()
            .children(true)
            .selector('g.crosshair')
            .element('g')
            .attrs({'class': 'crosshair'});
```

And I'd like to be able to just type -
```js
  var dataJoin = fc.utilities.dataJoin()
            .children(true)
            .selector('g.crosshair')
            .element('g')
            .attrs('class', 'crosshair');
```

janakerman:
#420

### DIFF
--- a/src/util/dataJoin.js
+++ b/src/util/dataJoin.js
@@ -90,7 +90,16 @@
             if (!arguments.length) {
                 return attrs;
             }
-            attrs = x;
+
+            if (arguments.length === 1) {
+                attrs = arguments[0];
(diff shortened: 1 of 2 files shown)


chrisprice:
:+1: 

Nice work with the comprehensive tests.

PR #464, opened 2015-07-02
Fix #459

### LINKED ISSUE #459: Firefox Example Chart
Regression viewable in new site http://chrisprice.io/d3fc/examples/
Firefox 38
Mouseover on the main chart causes the content to stretch out.

tlsim:
Firefox getComputedStyle() in innerDimensions util returns a greater than desired width value for 100% width SVG. Setting an explicit width fixes it.

I'm not sure what introduced the bug - the example worked on the old site in the same browser, and it had the same CSS as far as I can tell.

### DIFF
--- a/site/src/examples/index.md
+++ b/site/src/examples/index.md
@@ -13,11 +13,8 @@ title: Examples
 .navigator-row>td {
   height: 80px;
 }
-.chart {
-  width: 640px;
-}
 svg {
-  width: 100%;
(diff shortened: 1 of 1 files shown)


chrisprice:
Thanks @tlsim, I'm guessing [this line](https://github.com/ScottLogic/d3fc/blob/master/site/src/style/base.less#L2) is responsible.

PR #465, opened 2015-07-02
Don't use instanceof d3.selection to fix #458

### LINKED ISSUE #458: IE9 error
When viewing a number of the website examples, the following error is thrown and the chart doesn't or partially renders -

`SCRIPT438: Object doesn't support property or method 'querySelectorAll' 
scripts.js, line 564 character 5`

chrisprice:


ColinEberhardt:
Nice and simple :+1:

PR #472, opened 2015-07-05
Moved plot-area styling into chart.css

ColinEberhardt:
Small tweak, needed to change the overflow style on an example - so moved to CSS


chrisprice:
:+1:

PR #475, opened 2015-07-06
pass index to the decorate function

ColinEberhardt:
By passing the index through to the decorate function it is now possible to colour the stacked bar series in a more conventional fashion:

screen shot 2015-07-06 at 20 56 41

PR #476, opened 2015-07-07
Allow decoration of multi series

chrisprice:


ColinEberhardt:
LGTM - I'll update the decorate signature as part of #475

PR #477, opened 2015-07-07
Add coinbase feed and tidy up docs

chrisprice:
* [x] Add description of `start` & `end` parameters to docs.
* [x] Change `return cb(err)` to long-form


site/src/components/data/coinbase.md
```diff
@@ -0,0 +1,28 @@
...
+  var feed = fc.data.feed.coinbase();
+
+  // fetch some data!
+  feed(function(error, data) {
+    d3.select("#coinbase")
+      .text(JSON.stringify(error ? error : data, null, 2));
+  });
+---
+
+Fetch data from the [Coinbase exchange API](https://docs.exchange.coinbase.com/#market-data). The `product`, `start`, `end` and `granularity` can be customised, for more information on the parameters see the Coinbase docs.
```
ColinEberhardt: Might be worth noting that the component expects `start` and `end` parameters as `Date` instances, rather than the ISO-date strings that the coinbase API expected.

src/data/feed/coinbase.js
```diff
@@ -0,0 +1,75 @@
...
+            if (end != null) {
+                params.push('end=' + end.toISOString());
+            }
+            if (granularity != null) {
+                params.push('granularity=' + granularity);
+            }
+            var url = 'https://api.exchange.coinbase.com/products/' + product + '/candles?' + params.join('&');
+            d3.json(url, function(error, data) {
+                if (error) {
+                    return cb(error);
```
ColinEberhardt: is there any reason why the error path returns the callback?

ColinEberhardt:
A useful little component, just a couple of minor comments.

site/src/components/data/coinbase.md
chrisprice: Added todo.

src/data/feed/coinbase.js
chrisprice: Habit I suppose, it's a Node-ism I've picked up ([random example](https://github.com/joyent/node/blob/master/lib/fs.js#L1465)). It saves on an else statement or separate return statement.

ColinEberhardt: Oh yes, it's all over the place. I can't say I like it though, returning something from a function for the convenience of a slightly more concise implementation seems wrong to me.

Although I did try to get a `+` through in my PR ;-)

chrisprice: Don't feel strongly about it, I'll add a todo.

chrisprice:
Should be good now.

ColinEberhardt:
:+1:

PR #489, opened 2015-07-12
re-instated class on multi series datajoin

### LINKED ISSUE #486: Website example crosshair horizontal line in the incorrect location
image

ColinEberhardt:
The crosshair is broken at the moment:

screen shot 2015-07-12 at 19 22 30

This is because the multi series is placing one of the annotations within the trackball 'g' element which has a translate transform applied.

Re-instating the class on the multi series datajoin fixes this.

Fixes #486

### DIFF
--- a/src/series/multi.js
+++ b/src/series/multi.js
@@ -18,8 +18,9 @@
             decorate = fc.util.fn.noop;
 
         var dataJoin = fc.util.dataJoin()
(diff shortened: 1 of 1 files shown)


chrisprice:
Ah fair point, I'd forgotten we mixed multi series in with other types.

PR #496, opened 2015-07-15
WIP: Adds grouped series

janakerman:
This is an WIP of the grouped series. As you can see, it's very similar to the cycle series, but differs in that:
* It always uses an ordinal subScale as there is always a discrete number of series in a group.
* It uses the subScale to translate the group, rather than the xScale, and uses the xScale to plot the series, rather than the subScale.

This feels quite similar to the Cycle but I can't quite put my finger on how to factor out any commonality, or whether the Group series should be using the Cycle series via composition some how - but it's not that important right now.

Currently, the bar size, group spacing & bar spacing is a little hard, and it won't always render a sensible looking chart if you don't choose sensible numbers yourself. I was thinking of pulling the grouped series into its own name space fc.series.grouped.group & adding a fc.series.grouped.bar series.

This grouped.bar series would rebind a bar series' barWidth, and provide automatic group spacing, depending on the barWidth implementation provided.

To save the hassle of loading up the visual test to see the result - here's an image:
screen shot 2015-07-15 at 17 20 32

### DIFF
--- a/src/series/group.js
+++ b/src/series/group.js
@@ -1,0 +1,117 @@
+(function(d3, fc) {
+    'use strict';
+
+    fc.series.group = function() {
+
+        var decorate = fc.util.fn.noop,
+            xScale = d3.scale.linear(),
+            yScale = d3.scale.linear(),
+            xValue = function(d, i) { return d.x; },
+            groupValue = function(d, i) { return d.group;},
+            subScale = d3.scale.ordinal(),
+            subSeries = fc.series.bar(),
+            groupSize = fc.util.fractionalBarWidth(0.25);
+
+        var dataJoin = fc.util.dataJoin()
+            .selector('g.group')
+            .element('g')
+            .attrs({'class': 'group'});
+
+        var group = function(selection) {
+
+            selection.each(function(data, index) {
+
+                var dataByX = d3.nest()
+                    .key(xValue)
+                    .map(data);
+
+                var xValues = Object.keys(dataByX);
+
+                var width = groupSize(xValues.map(xScale)),
+                    halfWidth = width / 2;
+
+                var dataByGroup = d3.nest()
+                    .key(groupValue)
+                    .map(data);
+
+                var groupedValues = Object.keys(dataByGroup);
+
+                subScale.domain(groupedValues);
+
+                var g = dataJoin(this, groupedValues);
+
+                g.each(function(d, i) {
+
+                    subScale.rangePoints([-halfWidth, halfWidth]);
+
+                    var g = d3.select(this);
+
+                    g.attr('transform', 'translate(' + subScale(d) + ', 0)');
+
+                    subSeries.xScale(xScale)
+                        .yScale(yScale);
+
+                    d3.select(this)
+                        .datum(dataByGroup[d])
+                        .call(subSeries);
+
+                });
+
+                decorate(g, groupedValues, index);
+            });
+        };
+
+        group.decorate = function(x) {
+            if (!arguments.length) {
+                return decorate;
+            }
+            decorate = x;
+            return group;
+        };
+        group.xScale = function(x) {
+            if (!arguments.length) {
+                return xScale;
+            }
+            xScale = x;
+            return group;
+        };
+        group.yScale = function(x) {
+            if (!arguments.length) {
+                return yScale;
+            }
+            yScale = x;
+            return group;
+        };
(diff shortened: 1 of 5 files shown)


janakerman:
I've added a commit that makes it easier to change the offset algorithm. It also supplies two algorithms.0 The `intervalGroupOffset` offsets the bars by a given interval, useful for when you know the bar width, and want to place bars next to each other (user can +/- padding to the value they give to space bars out). The `fractionalGroupOffset` essentially just wraps the `fractionalBarWidth` algorithm to calculate how much of the available space a group should occupy, and then evenly spaces out the bars within that (I'm not sure how useful this one is - opinions please!).

It's worth noting that at the moment, this group series will work with any series. There's nothing specific to bars in it. Saying that, something still doesn't quite feel right about the visual test's use case. 

You have to [specify a fixed bar width](https://github.com/ScottLogic/d3fc/commit/f2a09fe60a269b753eb7b12289787e901f243768#diff-633270d5a8980978e6fe1f8bbde86722R84), otherwise, the bars expand to fill their entire X value, and all the bars overlap. This is because the bar series' sizing algorithm isn't really aware that it is part of a group. A few options here:
* Leave it, we're happy with this behaviour.
* Change the bar width algorithm so it can be given an integer representing how many bars it will share its space with.
* Write a more specific groupedBar series that some how handles the bar width.. haven't really though this through at all.

(I'll be adding tests for the algorithms and updating the website ect once we've solidified the implementation.)

visual-tests/src/test-fixtures/series/grouped-bar.js
```diff
@@ -0,0 +1,114 @@
...
+            .nice();
+
+        var colorScale = d3.scale.category20();
+
+        var bar = fc.series.bar()
+            .xValue(function(d) { return d.flavour; })
+            .yValue(function(d) { return d.sales; })
+            .barWidth(20);
+
+        var cycle = fc.series.group()
```
chrisprice: Name

src/series/group.js
```diff
@@ -0,0 +1,110 @@
...
+
+                var g = dataJoin(this, groupedValues);
+
+                g.each(function(d, i) {
+
+                    var g = d3.select(this);
+
+                    g.attr('transform', 'translate(' + offset(d, i, groupedValues, xPixels) + ', 0)');
+
+                    subSeries.xScale(xScale)
```
chrisprice: Does this need to be done in the each?

ColinEberhardt:
Great job @janakerman - sorry I've not got much time for detailed feedback. A few random thoughts ...

 + The translation required to shift each bar from the centre, would it make sense to do this with a sub-scale as per the cycle plot? (with that in mind, is this just a specialised cycle plot?)
 + You should be able to color the series by index, via decorate, as per stacked https://github.com/ScottLogic/d3fc/blob/master/visual-tests/src/test-fixtures/series/stacked-bar.js#L62
 + I think you're going to have to find a way for grouped and bar series to collaborate to resolve the bar width / spacing issues!

ColinEberhardt:
Thinking about it a bit more, the similarities to cycle are more coincidental. With cycle plots, the datapoints in each 'group' are connected, whereas here the bar's at the same index are connected between groups.

I think the important thing from my perspective is that the index passed to decorate allows you to 'address' each bar series (to colour each series for example), and that by performing a selection and `each` on each series you can address each individual bar (to add labels to each bar for example).

For the width / spacing, could the group series have a 'bar width' property that works in exactly the same way as the bar series, which computes the width / spacing of each group? The group series could then compute the bar width required for each sub-series. Hope that makes sense!

janakerman:
@ColinEberhardt - That's exactly the path of discovery I went down. Originally, I'd modified the visual test to draw this chart, only to find exactly that, each datapoint in a 'group' was in the same <g>, whereas it felt natural to have a 'series' represent a 'group'.

RE your comment 'would it make sense to do this with a sub-scale as per the cycle plot' - the two offset algorithms I've written do actually use an ordinal scale for calculating the group offsets.

RE decoration, I actually haven't looked into the decoration of the series yet. I wanted a few opinion on the early approach before I spent any more time on this design, tests, or website. I had the exact same thing in mind for the decoration though.

I'll have a think about the bar spacing - I think it might be a tricky one! :)

I'll also have a think about how the auto bar sizing can be achieved.

ColinEberhardt:
Thanks @janakerman - we've implemented this in #600

PR #500, opened 2015-07-16
Website width fix

joelwilliamsmith:
A bunch of commits that fix various page width issues on smaller devices.


site/src/_includes/example-fixture.hbs
```diff
@@ -1,8 +1,10 @@
-<div id="{{>component-id component}}" class="chart"> </div>
```
chrisprice: Can we put this in the less?

joelwilliamsmith: Yep, just being lazy...

chrisprice:
Thanks

PR #485, opened 2015-07-09
Update README to reflect namespace changes

jleft:

PR #503, opened 2015-07-28
[WIP] A drop-in replacement for the D3 SVG axis

### LINKED ISSUE #466: IDEA: Create our own axis implementation
One thing that I find a little frustrating with the D3 axes is that there isn't an elegant way to re-position the labels. The [standard solution](http://stackoverflow.com/questions/17544546/d3-js-align-text-labels-between-ticks-on-the-axis) seems to be to render the axis then tweak the results. 

For example:

```
d3.selectAll('.x-axis text')
    .style('text-anchor', 'left')
    .attr('transform', 'translate(35, 0)');
```

Which is pretty ugly.

If the axis were re-implemented / wrapped / adapted to follow the decorate pattern, the above code could be implemented as part of the enter selection via decorate.

Thoughts?

ColinEberhardt:
Fixes #466 

This is a drop-in replacement for the D3 scale, with the aim of making it conform to the decorate pattern (which I think is quite central to d3fc). The implementation differs from the D3 equivalent in a few ways:

 + Uses SVG `path` for the ticks rather than `line`, this makes the code simpler by using a 'transpose' function that can transpose an SVG path supporting vertical / horizontal axes
 + D3 does not ship with CSS, as a result the axis implementation 'manually' positions / styles the labels. This implementation has CSS making the label positioning a bit simpler.

I've also pulled out a few scale related utility functions.

One thing I want to debate is the axis structure and the impact it has on CSS. Currently we rely on the axis being added within a `g` with a class of `axis`. This feels a bit fragile. There are two options:

 1. The axis component adds a `g` with this class.
 2. The tick containers have additional classes added that allow us unambiguously target them with selectors, perhaps having a class of `axis tick orient-bottom`. 

I am not keen on (2) because it results in a lot of repeated information. My preference would be to pull the common information into a container `g` with a class of `axis orient-bottom`.

TODO (before merge)
 + [x] Add documentation
 + [x] Finalise the CSS

TODOs (from review feedback, duplicated from https://github.com/ScottLogic/d3fc/pull/503#issuecomment-125765180 so that they appear in the PR summary)
 + [x] use `fc.util.datajoin` to create the domain path
 + [x] try moving `isVertical()` into the helper functions to see if it results in simpler code
 + [x] ensure orientation changes function properly
 + [x] test and expose `decorate`
 + [x] be self-consistent regarding ordinal axis identification

Anyhow, any feedback or comments relating the CSS / axis structure would be welcome.

### DIFF
--- a/src/svg/axis.js
+++ b/src/svg/axis.js
@@ -56,6 +56,10 @@
             return orient === 'left' || orient === 'right';
         }
 
+        function tryApply(fn, defaultVal) {
+            return scale[fn] ? scale[fn].apply(scale, tickArguments) : defaultVal;
+        }
+
         var axis = function(selection) {
 
             selection.each(function(data, index) {
@@ -63,10 +67,6 @@
                 // Stash a snapshot of the new scale, and retrieve the old snapshot.
                 var scaleOld = this.__chart__ || scale;
                 this.__chart__ = scale.copy();
-
-                function tryApply(fn, defaultVal) {
-                    return scale[fn] ? scale[fn].apply(scale, tickArguments) : defaultVal;
-                }
 
                 var ticksArray = tickValues == null ? tryApply('ticks', scale.domain()) : tickValues;
                 var tickFormatter = tickFormat == null ? tryApply('tickFormat', fc.util.fn.identity) : tickFormat;
@@ -83,7 +83,7 @@
                     ]);
 
                 var domainLine = domainPathDataJoin(container, [data]);
-                d3.transition(domainLine)
+                domainLine
                     .attr('d', svgDomainLine(domainPathData));
 
                 // datajoin and construct the ticks / label
--- a/src/svg/axis.js
+++ b/src/svg/axis.js
@@ -204,6 +204,14 @@
             return axis;
         };
 
+        axis.decorate = function(x) {
+            if (!arguments.length) {
+                return decorate;
+            }
+            decorate = x;
+            return axis;
+        };
+
         return axis;
 
     };
--- a/src/svg/axis.js
+++ b/src/svg/axis.js
@@ -1,0 +1,211 @@
+(function(d3, fc) {
+    'use strict';
+
+    // A drop-in replacement for the D3 axis, supporting the decorate pattern.
+    fc.svg.axis = function() {
+
+        var scale = d3.scale.identity(),
+              decorate = fc.util.fn.noop,
+              orient = 'bottom',
+              tickArguments = [10],
+              tickValues = null,
+              tickFormat = null,
+              outerTickSize = 6,
+              innerTickSize = 6,
+              tickPadding = 3,
+              svgDomainLine = d3.svg.line();
+
+        var dataJoin = fc.util.dataJoin()
+            .selector('g.tick')
+            .element('g')
+            .key(fc.util.fn.identity)
+            .attrs({'class': 'tick'});
+
+        // returns a function that creates a translation based on
+        // the bound data
+        function containerTranslate(s, trans) {
+            return function(d) {
+                return trans(s(d), 0);
+            };
+        }
+
+        function translate(x, y) {
+            return 'translate(' + x + ', ' + y + ')';
+        }
+
+        function translateTransposed(x, y) {
+            return 'translate(' + y + ', ' + x + ')';
+        }
+
+        function transposeArray(arr) {
+            return arr.map(function(d) {
+                return [d[1], d[0]];
+            });
+        }
+
+        function isVertical() {
+            return orient === 'left' || orient === 'right';
+        }
+
+        var axis = function(selection) {
+
+            selection.each(function(data, index) {
+
+                // Stash a snapshot of the new scale, and retrieve the old snapshot.
+                var scaleOld = this.__chart__ || scale;
+                this.__chart__ = scale.copy();
+
+                function tryApply(fn, defaultVal) {
+                    return scale[fn] ? scale[fn].apply(scale, tickArguments) : defaultVal;
+                }
+
+                var ticksArray = tickValues == null ? tryApply('ticks', scale.domain()) : tickValues;
+                var tickFormatter = tickFormat == null ? tryApply('tickFormat', fc.util.fn.identity) : tickFormat;
+
+                // orientation logic
+                var sign = orient === 'bottom' || orient === 'right' ? 1 : -1;
+                var pathTranspose = isVertical() ? transposeArray : fc.util.fn.identity;
+                var transform = isVertical() ? translateTransposed : translate;
+
+                var container = d3.select(this);
+
+                // add the domain line
+                var range = fc.util.scale.range(scale);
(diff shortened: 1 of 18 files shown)


src/svg/axis.js
```diff
@@ -0,0 +1,211 @@
...
+
+                domainLine.enter()
+                    .append('path')
+                    .classed('domain', true);
+
+                d3.transition(domainLine)
+                    .attr('d', svgDomainLine(domainPathData));
+
+                // datajoin and construct the ticks / label
+                dataJoin.attrs({
```
chrisprice: I don't think this will work if you change the orientation, the attributes will only be set on enter.

src/svg/axis.js
```diff
@@ -0,0 +1,211 @@
...
+                    // in order to get the correct enter transition - however, for ordinal
+                    // scales the tick will not exist on the old scale, so use the current position
+                    'transform': containerTranslate(fc.util.scale.isOrdinal(scale) ? scale : scaleOld, transform)
+                });
+
+                var g = dataJoin(container, ticksArray);
+
+                // enter
+                var labelOffset = sign * (innerTickSize + tickPadding);
+                g.enter().classed('orient-' + orient, true);
```
chrisprice: Could this go into the previous attrs?

src/util/scale.js
```diff
@@ -0,0 +1,16 @@
...
+    fc.util.scale = {
+        // ordinal axes have a rangeExtent function, this adds any padding that
+        // was applied to the range. This functions returns the rangeExtent
+        // if present, or range otherwise
+        range: function(scale) {
+            return scale.rangeExtent ? scale.rangeExtent() : scale.range();
+        },
+
+        isOrdinal: function(scale) {
+            return scale.rangeBands;
```
chrisprice: How come this check is different to the above one?

src/svg/axis.js
```diff
@@ -0,0 +1,211 @@
...
+
+        var axis = function(selection) {
+
+            selection.each(function(data, index) {
+
+                // Stash a snapshot of the new scale, and retrieve the old snapshot.
+                var scaleOld = this.__chart__ || scale;
+                this.__chart__ = scale.copy();
+
+                function tryApply(fn, defaultVal) {
```
chrisprice: Should this go with the other helper funcs?

src/svg/axis.js
```diff
@@ -0,0 +1,211 @@
...
+                    return scale[fn] ? scale[fn].apply(scale, tickArguments) : defaultVal;
+                }
+
+                var ticksArray = tickValues == null ? tryApply('ticks', scale.domain()) : tickValues;
+                var tickFormatter = tickFormat == null ? tryApply('tickFormat', fc.util.fn.identity) : tickFormat;
+
+                // orientation logic
+                var sign = orient === 'bottom' || orient === 'right' ? 1 : -1;
+                var pathTranspose = isVertical() ? transposeArray : fc.util.fn.identity;
```
chrisprice: Could this and the next line have the conditionals collapsed into the helper functions?

src/svg/axis.js
```diff
@@ -0,0 +1,211 @@
...
+                // add the domain line
+                var range = fc.util.scale.range(scale);
+                var domainPathData = pathTranspose([
+                      [range[0], sign * outerTickSize],
+                      [range[0], 0],
+                      [range[1], 0],
+                      [range[1], sign * outerTickSize]
+                    ]);
+
+                var domainLine = container.selectAll('path.domain')
```
chrisprice: Worth just using fc.util.datajoin?

src/svg/axis.js
```diff
@@ -0,0 +1,211 @@
...
+            return axis;
+        };
+
+        axis.tickSize = function(x) {
+            var n = arguments.length;
+            if (!n) {
+                return innerTickSize;
+            }
+            innerTickSize = Number(x);
+            outerTickSize = Number(arguments[n - 1]);
```
chrisprice: This looks bonkers but I see it matches d3 :)

src/svg/axis.js
```diff
@@ -0,0 +1,211 @@
...
+                   .attr('transform', transform(0, labelOffset))
+                   .text(tickFormatter);
+
+                // for non ordinal scales, exit by animating the tick to its new location
+                if (!fc.util.scale.isOrdinal(scale)) {
+                    g.exit()
+                        .attr('transform', containerTranslate(scale, transform));
+                }
+
+                decorate(g, data, index);
```
chrisprice: I think you're missing the function to allow this to be configured? Does that mean there's no test of it?

chrisprice:
> I am not keen on (2) because it results in a lot of repeated information. My preference would be to pull the common information into a container g with a class of axis orient-bottom.

I'd rather avoid containers if they're not necessary, is there really that much duplication?

chrisprice:
Looks good, I especially like the new `fc.util.scale` it's long overdue! I wonder if we need some examples of decoration to really prove it's usefulness though?

src/svg/axis.js
ColinEberhardt: That's pretty ironic!

src/util/scale.js
ColinEberhardt: Good spot - this was based on a refactor of some D3 code, which doesn't appear to be self-consistent regarding how it identified ordinal axes.

ColinEberhardt:
@chrisprice thanks for the review feedback, some very useful comments. 

TODOs:
 + [ ] use `fc.util.datajoin` to create the domain path
 + [ ] try moving `isVertical()` into the helper functions to see if it results in simpler code
 + [ ] ensure orientation changes function properly
 + [ ] test and expose `decorate`
 + [ ] be self-consistent regarding ordinal axis identification

ColinEberhardt:
> I wonder if we need some examples of decoration to really prove it's usefulness though?

I'm actually working on that right now on another branch. We've got a good 'quick start', and component documentation but nothing much in between.

I'm doing the following:
 + Moving the 'getting started' from the README.md into the site, this should make it less likely that this code will be out of synch again (as per #485)
 + Adding a page that describe decorate

Basically I think we need some documentation that sits in between the very high-level quick start and the much more detailed component docs.

ColinEberhardt:
@chrisprice - all updated and documentation added. Let me know if you spot anything else.

src/svg/axis.js
```diff
@@ -0,0 +1,218 @@
...
+                var range = fc.util.scale.range(scale);
+                var domainPathData = pathTranspose([
+                      [range[0], sign * outerTickSize],
+                      [range[0], 0],
+                      [range[1], 0],
+                      [range[1], sign * outerTickSize]
+                    ]);
+
+                var domainLine = domainPathDataJoin(container, [data]);
+                d3.transition(domainLine)
```
chrisprice: Don't think this transition is needed anymore.

ColinEberhardt: Good spot - it wasn't needed.

PR #504, opened 2015-08-02
Updated strapline

ColinEberhardt:
We had different straplines in various locations (GitHub project description, `package.json`, `README.md`) - this updates them all to match the website strapline which feels the most accurate at the moment.


chrisprice:
Good spot.

PR #508, opened 2015-08-05
series no longer key on xValue

ColinEberhardt:
Series currently use the x-value for the data-join key, this prohibits them being used for any chart where multiple datapoints share the same x-value.

This PR changes the series to use the default, index-based, key and rebinds to expose it as a configurable property.

As a result, you can create plots like this GitHub punchcard:

screen shot 2015-08-05 at 08 15 07

### DIFF
--- a/src/series/area.js
+++ b/src/series/area.js
@@ -82,6 +82,9 @@
             return area;
         };
 
-        return d3.rebind(area, areaData, 'interpolate', 'tension');
+        d3.rebind(area, dataJoin, 'key');
+        d3.rebind(area, areaData, 'interpolate', 'tension');
+
+        return area;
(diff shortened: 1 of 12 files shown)


ColinEberhardt:
On a related note, I'd like to propose changing the default xValue and yValue accessors in all non-financial specific series types to be something more generic, e.g. `d.x` / `d.y`

chrisprice:
Nice one, I'm a big fan of this pull request as it makes the series more flexible without changing the functionality for the majority of users (i.e. those who don't care about transitioning updating data). 

I'm less keen on your other proposal, do you want to spin it out into a separate issue and we can discuss it there?

PR #509, opened 2015-08-05
Add menu to example section and add basecoin example

chrisprice:
The example landing page needs work but I figure we can do that separately when we have an idea of how we want it to look.

Also there's an implicit dependency between this and the blog post.


chrisprice:
https://github.com/ScottLogic/blog/pull/216

chrisprice:
Yahoo finance chart is misbehaving even though it's the [latest version](http://bl.ocks.org/ColinEberhardt/raw/875bb098218a94cad7f5/) -

image

chrisprice:
Think this is now good to go.

PR #512, opened 2015-08-06
Rename fc.util.dataJoin().attrs to attr

chrisprice:
In keeping with d3.selection().attr


ColinEberhardt:
As `attr` supports `(key, value)`, is it worth updating the single-valued usages to this syntax?

chrisprice:
I can definitely do that

chrisprice:
Done :)

ColinEberhardt:
> Done :)

awesome

PR #513, opened 2015-08-07
Fixes the Yahoo example

ColinEberhardt:
+ axis styling moved to use decorate
 + linearTimeSeries updated to expose the decorate function for axes
 + bar svg was incorrectly using 'V' (absolute) rather than 'v' (relative), this took a while to track down!


chrisprice:
:+1:

PR #517, opened 2015-08-18
Introduced yAxisWidth and related layout

djmiley:
To align with use and exposure of `xAxisHeight`

### DIFF
--- a/src/chart/linearTimeSeries.js
+++ b/src/chart/linearTimeSeries.js
@@ -4,6 +4,7 @@
     fc.chart.linearTimeSeries = function() {
 
         var xAxisHeight = 20;
+        var yAxisWidth = 40;
         var plotArea = fc.series.line();
         var xScale = fc.scale.dateTime();
         var yScale = d3.scale.linear();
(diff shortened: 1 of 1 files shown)


src/chart/linearTimeSeries.js
```diff
@@ -4,6 +4,7 @@
     fc.chart.linearTimeSeries = function() {
 
         var xAxisHeight = 20;
+        var yAxisWidth = 40;
```
chrisprice: Can we leave the defaults as they were? i.e. 0

chrisprice:
One small comment but otherwise :+1:

ColinEberhardt:
Thanks for the contribution, I'm definitely :+1: to the idea of more configuration. Although I am wondering what this chart looks like if the y-axis width is != 0? Wouldn't you also have to change the y axis orientation (via `orient`) for this to make sense?

chrisprice:
@ColinEberhardt I agree long-term it's probably not the right fix, but the same holds for the `xAxisHeight` so I think being consistent short-term is an ok fix.

ColinEberhardt:
Far enough, if it fixes an immediate need, I'm happy :-)

chrisprice:
Merged, thanks @djmiley

djmiley:
:+1:

PR #518, opened 2015-08-26
Cartesian chart

ColinEberhardt:
I think we need a more versatile chart component with axis labels and configurable scales. This is a quick prototype to test out a few ideas.

The non-standard part of this component is the way the scales work. In order to expose them as properties the settings from previous scales need to be re-applied and the new scale properties rebound.

What do others think - does this look viable?

### DIFF
--- a/src/chart/cartesianChart.js
+++ b/src/chart/cartesianChart.js
@@ -26,9 +26,9 @@
             .orient('right');
 
         var containerDataJoin = fc.util.dataJoin()
-            .selector('svg.cartesianChart')
+            .selector('svg.cartesian-chart')
             .element('svg')
-            .attr({'class': 'cartesianChart', 'layout-css': 'flex: 1'});
+            .attr({'class': 'cartesian-chart', 'layout-css': 'flex: 1'});
 
         var cartesianChart = function(selection) {
 
--- a/src/chart/cartesianChart.js
+++ b/src/chart/cartesianChart.js
@@ -10,10 +10,10 @@
             yAxisWidth = 40,
             yAxisLabelWidth = 20,
             xAxisLabelHeight = 20,
-            titleHeight = 20,
+            chartLabelHeight = 20,
             yAxisLabel = 'y-axis',
             xAxisLabel = 'x-axis',
-            title = 'Chart Title',
+            chartLabel = 'Chart',
             plotArea = fc.series.line(),
             decorate = fc.util.fn.noop;
 
@@ -101,7 +101,7 @@
 
                 // position the title
                 svg.select('.title')
-                    .layout({'height': titleHeight, 'alignItems': 'center', 'justifyContent': 'center',
+                    .layout({'height': chartLabelHeight, 'alignItems': 'center', 'justifyContent': 'center',
                         'marginLeft': marginLeft, 'marginRight': marginRight});
 
                 // perform the flexbox / css layout
@@ -109,7 +109,7 @@
 
                 // update the label text
                 svg.select('.title .label')
-                    .text(title);
+                    .text(chartLabel);
 
                 svg.select('.x-axis .label')
                     .text(xAxisLabel);
@@ -155,18 +155,18 @@
         fc.util.rebindProperties(cartesianChart, xAxis, axisReboundProperties, 'x');
         fc.util.rebindProperties(cartesianChart, yAxis, axisReboundProperties, 'y');
 
-        cartesianChart.titleHeight = function(x) {
+        cartesianChart.chartLabelHeight = function(x) {
             if (!arguments.length) {
-                return titleHeight;
+                return chartLabelHeight;
             }
-            titleHeight = x;
+            chartLabelHeight = x;
             return cartesianChart;
         };
-        cartesianChart.title = function(x) {
+        cartesianChart.chartLabel = function(x) {
             if (!arguments.length) {
-                return title;
+                return chartLabel;
             }
-            title = x;
+            chartLabel = x;
             return cartesianChart;
         };
         cartesianChart.plotArea = function(x) {
--- a/src/chart/cartesianChart.js
+++ b/src/chart/cartesianChart.js
@@ -1,0 +1,242 @@
+(function(d3, fc) {
+    'use strict';
+
+    fc.chart.cartesianChart = function(xScale, yScale) {
+
+        xScale = xScale || d3.scale.linear();
+        yScale = yScale || d3.scale.linear();
+
+        var xAxisHeight = 20,
+            yAxisWidth = 40,
+            yAxisLabelWidth = 20,
+            xAxisLabelHeight = 20,
+            titleHeight = 20,
+            yAxisLabel = 'y-axis',
+            xAxisLabel = 'x-axis',
+            title = 'Chart Title',
+            plotArea = fc.series.line(),
+            decorate = fc.util.fn.noop;
+
+        var xAxis = fc.svg.axis()
+            .scale(xScale)
+            .orient('bottom');
+
+        var yAxis = fc.svg.axis()
+            .scale(yScale)
+            .orient('right');
+
+        var scaleReboundProperties = ['nice', 'domain'];
+        var axisReboundProperties = ['ticks', 'tickValues', 'tickSize', 'innerTickSize',
+            'outerTickSize', 'tickPadding', 'tickFormat', 'orient'];
+
+        function capitalizeFirstLetter(str) {
+            return str.charAt(0).toUpperCase() + str.slice(1);
+        }
+
+        function rebind(component, properties, prefix) {
+            var bindings = {};
+            properties.forEach(function(property) {
+                bindings[prefix + capitalizeFirstLetter(property)] = property;
+            });
+            fc.util.rebind(cartesianChart, component, bindings);
+        }
+
+        var containerDataJoin = fc.util.dataJoin()
+            .selector('svg.cartesianChart')
+            .element('svg')
+            .attr({'class': 'cartesianChart', 'layout-css': 'flex: 1'});
+
+        var cartesianChart = function(selection) {
+
+            selection.each(function(data, index) {
+
+                var container = d3.select(this);
+
+                var svg = containerDataJoin(container, [data]);
+                svg.enter().html(
+                    '<g class="title"> \
+                        <text class="label" layout-css="height: 0; width: 0"/> \
+                    </g> \
+                    <g layout-css="flex: 1; flexDirection: row"> \
+                        <g class="y-axis left"> \
+                            <g class="label-container"> \
+                                <g layout-css="height: 0; width: 0"> \
+                                    <text class="label" transform="rotate(-90)"/> \
+                                </g> \
+                            </g> \
+                            <g class="padding"/> \
+                            <g class="axis" layout-css="width: 0"/> \
+                        </g> \
+                        <svg class="plot-area" layout-css="flex: 1"/> \
+                        <g class="y-axis right"> \
+                            <g class="axis" layout-css="width: 0"/> \
+                            <g class="padding"/> \
+                            <g class="label-container"> \
+                                <g layout-css="height: 0; width: 0"> \
+                                    <text class="label" transform="rotate(-90)"/> \
+                                </g> \
+                            </g> \
+                        </g> \
+                    </g> \
+                    <g class="x-axis"> \
+                        <g class="axis"/> \
+                        <g class="label-container"> \
+                            <text class="label" layout-css="height: 0; width: 0"/> \
+                        </g> \
+                    </g>');
+
+                // configure the y axis containers based on y-axis orientation
+                var opposingOrientation = yAxis.orient() === 'left' ? 'right' : 'left';
+
+                var totalYAxisWidth = yAxisWidth + yAxisLabelWidth;
+                svg.select('.y-axis.' + yAxis.orient())
+                    .layout({'width': totalYAxisWidth, 'flexDirection': 'row'});
+
+                svg.select('.y-axis.' + yAxis.orient() + ' .label-container')
+                    .layout({'width': yAxisLabelWidth, 'alignItems': 'center', 'justifyContent': 'center'});
+
+                svg.select('.y-axis.' + yAxis.orient() + ' .padding')
+                    .layout('width', yAxisWidth);
+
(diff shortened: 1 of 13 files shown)


chrisprice:
I know this isn't a problem specifically with this PR (and I don't have any ideas how to avoid it) but there's so much boilerplate for building static DOM structure.

Also, the re-binding stuff does look a bit odd. As they're unlikely to be changed, could they be passed in as arguments to the factory method?

ColinEberhardt:
You're right about the DOM construction code, it is very hard to determine what
the component creates from inspecting the code. I'll see if I can create a more
generic way of creating 'static' DOM from a structured model.
Not sure what you mean about passing the rebounding arguments to the factory
method? Could you give a quick example?
Thanks
Colin E.

On Wed, Aug 26, 2015 at 4:13 pm, ScottLogic/d3fc
<reply@reply.github.com>
wrote:
I know this isn't a problem specifically with this PR (and I don't have any
ideas how to avoid it) but there's so much boilerplate for building static DOM
structure.

Also, the re-binding stuff does look a bit odd. As they're unlikely to be
changed, could they be passed in as arguments to the factory method?

—
[https://github.com/ScottLogic/d3fc/pull/518#issuecomment-135057024] .[https://github.com/notifications/beacon/ABDBfn09zHibsVueefFZHxVN9Dd1dZLtks5orc8cgaJpZM4FyU_b.gif]

chrisprice:
Sorry, that was a wordy was of saying -

```js
fc.chart.cartesianChart = function(xScale, yScale) {
```

And not letting them be subsequently changed.

ColinEberhardt:
> I know this isn't a problem specifically with this PR (and I don't have any ideas how to avoid it) but there's so much boilerplate for building static DOM structure.

You've got a point - I find ti hard to follow the DOM construction code. I've pushed a commit with a very different approach which uses `html` to 'bulk' add the entire structure then uses selectors to locate and update the various layout properties. 

https://github.com/ColinEberhardt/d3-financial-components/blob/86e1397f88f3935404dd66f5c9cb34c10073a867/src/chart/cartesianChart.js#L58-L82

I'm not so keen on having the scales provided as factory method args, it's not a pattern we use elsewhere.

chrisprice:
Is that not because we've not had a good use case?

It is used in other parts of d3 e.g.
https://github.com/mbostock/d3/wiki/Requests#d3_xhr.

ColinEberhardt:
Still not convinced about passing the scales to the factory function - it works, but the nested fluent syntax looks a bit ugly:

    var chart = fc.chart.cartesianChart(
            d3.scale.linear()
                .domain(fc.util.extent(data, 'x')),
            d3.scale.log()
                .domain(fc.util.extent(data, 'y')))
        .yOrient(chartConfig[0].value)
        .yAxisLabel(chartConfig[1].value)
        .xAxisLabel(chartConfig[2].value)
        .xAxisLabelHeight(Number(chartConfig[3].value))
        .yAxisLabelWidth(Number(chartConfig[4].value))
        .xAxisHeight(Number(chartConfig[5].value))
        .yAxisWidth(Number(chartConfig[6].value));

chrisprice:
Yea that does look odd. I'd imagined the scales would still have their
properties rebound though, would that help?

ColinEberhardt:
@chrisprice I think this one is ready for a proper review, I've dropped the [WIP] prefix! Squashed, and rebased.

> I'd imagined the scales would still have their properties rebound though, would that help?

Yeah, that did, the API works quite well now.

I was going to change `linearTimeSeries` to use the same construction pattern, but realised it requires constructing the SVG container. We could probably do with debating this one a little more!

src/chart/cartesianChart.js
```diff
@@ -0,0 +1,232 @@
...
+        yScale = yScale || d3.scale.linear();
+
+        var xAxisHeight = 20,
+            yAxisWidth = 40,
+            yAxisLabelWidth = 20,
+            xAxisLabelHeight = 20,
+            titleHeight = 20,
+            yAxisLabel = 'y-axis',
+            xAxisLabel = 'x-axis',
+            title = 'Chart Title',
```
chrisprice: `chartLabel` for consistency? Also make the default value `chart`?

src/util/rebind.js
```diff
@@ -24,4 +24,19 @@
         return target;
     };
 
+    function capitalizeFirstLetter(str) {
+        return str.charAt(0).toUpperCase() + str.slice(1);
```
chrisprice: `str[0].toUpperCase() + str.slice(1)`

src/util/rebind.js
```diff
@@ -24,4 +24,19 @@
...
+    }
+
+    /**
+     * Rebinds multiple properties from the source component, adding the given prefix.
+     */
+    fc.util.rebindProperties = function(target, source, properties, prefix) {
```
chrisprice: How about calling it `rebindAll`, dropping `properties` and treating the arguments after `prefix` as excludes?

src/chart/chart.css
```diff
@@ -19,4 +19,15 @@
...
+}
+
+.y-axis .label,
+.x-axis .label,
+.title .label {
+    dominant-baseline: middle;
+    text-anchor: middle;
+}
+
+.cartesianChart {
```
chrisprice: Should we have the rule that all classes are dash-separated rather than camel-cased?

src/chart/cartesianChart.js
ColinEberhardt: Good idea, I've also renamed `titleHeight` to `chartLabelHeight` for consistency.

src/util/rebind.js
ColinEberhardt: Nice idea - I'll give that a go.

src/chart/chart.css
ColinEberhardt: I think we already have that convention, I just ignored it!

ColinEberhardt:
Thanks for the review - `rebindAll` is a nice idea, and also forced me to put a bit more thought into which properties are actually rebound.

chrisprice:
Nice one, I'll take another look next time I'm in front of a computer

src/chart/cartesianChart.js
```diff
@@ -0,0 +1,233 @@
...
+                    </g> \
+                    <g class="x-axis"> \
+                        <g class="axis"/> \
+                        <g class="label-container"> \
+                            <text class="label" layout-css="height: 0; width: 0"/> \
+                        </g> \
+                    </g>');
+
+                // configure the y axis containers based on y-axis orientation
+                var opposingOrientation = yAxis.orient() === 'left' ? 'right' : 'left';
```
chrisprice: There are a few statically defined values in the `layout` argument object. I assume that's because `layout` doesn't support applying deltas using the object syntax? Is that something we should create an issue for?

src/chart/cartesianChart.js
```diff
@@ -0,0 +1,233 @@
...
+        var scaleExclusions = [
+            'range', 'rangeRound', // the scale range is set via the component layout
+            'tickFormat'           // use axis.tickFormat instead
+        ];
+        fc.util.rebindAll(cartesianChart, xScale, 'x', scaleExclusions);
+        fc.util.rebindAll(cartesianChart, yScale, 'y', scaleExclusions);
+
+        // exclude scale because this component associates the scale with the axis, and
+        // exclude x axis orientation, as only y axis can be repositions to the left / right
+        fc.util.rebindAll(cartesianChart, xAxis, 'x', ['scale', 'orient']);
```
chrisprice: Is it worth allowing the x-axis to be positioned at the top?

src/util/rebind.js
```diff
@@ -24,4 +24,30 @@
...
+    function capitalizeFirstLetter(str) {
+        return str[0].toUpperCase() + str.slice(1);
+    }
+
+    /**
+     * Rebinds all the methods from the source component, adding the given prefix. An
+     * optional exclusions parameter can be used to specify methods which should not
+     * be rebound.
+     */
+    fc.util.rebindAll = function(target, source, prefix, exclusions) {
```
chrisprice: Could the exclusions be made an array or var-args? I can see valuable use cases for both.

src/util/rebind.js
```diff
@@ -24,4 +24,30 @@
...
+        exclusions.forEach(function(property) {
+            if (!source.hasOwnProperty(property)) {
+                throw new Error('The method ' + property + ' does not exist on the source object');
+            }
+        });
+
+        var bindings = {};
+        for (var property in source) {
+            if (source.hasOwnProperty(property) && exclusions.indexOf(property) === -1) {
+                bindings[prefix + capitalizeFirstLetter(property)] = property;
```
chrisprice: Would it easier to just call `d3.rebind` here?

src/chart/cartesianChart.js
ColinEberhardt: Yes it would, although I think I've got a better idea for how to handle axis orientation, I'll just try out this alternative first.

src/util/rebind.js
ColinEberhardt: Nice idea - done.

src/util/rebind.js
ColinEberhardt: I don't think that will work, these bindings need to be 'aliased' to a different property name.

ColinEberhardt:
@chrisprice I've made quite a big update to this code:

 + Axes are not `fc.series.axis` allowing the baseline to be configured
 + The chart layout has been completely changed and simplified. It is simple a single container (plot area) with a configurable margin. Axes are now on the chart boundaries (as you suggested ;-)
 + Both axes support 'orient'

This feels like a much more elegant approach.

src/chart/cartesianChart.js
```diff
@@ -0,0 +1,211 @@
...
+        xScale = xScale || d3.scale.linear();
+        yScale = yScale || d3.scale.linear();
+
+        var plotAreaMargin = {
+                top: 20,
+                bottom: 30,
+                left: 0,
+                right: 40
+            },
+            yAxisLabel = 'y-axis',
```
chrisprice: I can't see anyone leaving these labels as-is, is it worth picking defaults that are more likely to be applicable? E.g. `''`?

src/chart/cartesianChart.js
```diff
@@ -0,0 +1,211 @@
...
+                    });
+
+                var yAxisLayout = {
+                    position: 'absolute',
+                    top: margin.top,
+                    bottom: margin.bottom,
+                    alignItems: 'center',
+                    flexDirection: 'row'
+                };
+                if (yAxis.orient() === 'right') {
```
chrisprice: `yAxisLayout[yAxis.orient] = 0;`

src/chart/cartesianChart.js
```diff
@@ -0,0 +1,211 @@
...
+            return cartesianChart;
+        };
+        cartesianChart.plotArea = function(x) {
+            if (!arguments.length) {
+                return plotArea;
+            }
+            plotArea = x;
+            return cartesianChart;
+        };
+        cartesianChart.xAxisLabel = function(x) {
```
chrisprice: `xLabel`

src/chart/cartesianChart.js
```diff
@@ -0,0 +1,211 @@
...
+            return cartesianChart;
+        };
+        cartesianChart.plotAreaMargin = function(x) {
+            if (!arguments.length) {
+                return plotAreaMargin;
+            }
+            plotAreaMargin = x;
+            return cartesianChart;
+        };
+        cartesianChart.yAxisLabel = function(x) {
```
chrisprice: `yLabel`

src/chart/cartesianChart.js
```diff
@@ -0,0 +1,211 @@
...
+            return cartesianChart;
+        };
+        cartesianChart.xAxisLabel = function(x) {
+            if (!arguments.length) {
+                return xAxisLabel;
+            }
+            xAxisLabel = x;
+            return cartesianChart;
+        };
+        cartesianChart.plotAreaMargin = function(x) {
```
chrisprice: Just `margin`?

src/chart/cartesianChart.js
```diff
@@ -0,0 +1,211 @@
...
+        var cartesianChart = function(selection) {
+
+            selection.each(function(data, index) {
+
+                var container = d3.select(this);
+
+                var svg = containerDataJoin(container, [data]);
+                svg.enter().html(
+                    '<g class="title"> \
+                        <g layout-css="height: 0; width: 0"> \
```
chrisprice: I don't think this guarantees that labels are wholly within the bounding box?

src/chart/cartesianChart.js
ColinEberhardt: done

src/chart/cartesianChart.js
ColinEberhardt: done

src/chart/cartesianChart.js
ColinEberhardt: done

src/chart/cartesianChart.js
ColinEberhardt: done

ColinEberhardt:
@chrisprice - all review comments addressed, I think

chrisprice:
:+1: from me, if you feel inclined you could make that change but otherwise feel free to create an issue and merge.

ColinEberhardt:
Yay, time for a merge beer

chrisprice:
Going to push this out as 1.1.0, there's been a couple of things added
recently which are useful.

ColinEberhardt:
Cool, I'll add the docs page for the website shortly.

On Fri, Sep 11, 2015 at 5:48 pm, ScottLogic/d3fc
<reply@reply.github.com>
wrote:
Going to push this out as 1.1.0, there's been a couple of things added
recently which are useful.

On Fri, 11 Sep 2015 at 17:47 Colin Eberhardt <notifications@github.com>
wrote:

> Yay, time for a merge beer
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ScottLogic/d3fc/pull/518#issuecomment-139595903>.
>
—
[https://github.com/ScottLogic/d3fc/pull/518#issuecomment-139596209] .[https://github.com/notifications/beacon/ABDBfkNOZ4AEmktSf8PZ8MdkPm_UlaNvks5owv1ngaJpZM4FyU_b.gif]

PR #523, opened 2015-08-28
Bugfix for macd rerendering causes duplicates

### LINKED ISSUE #522: Re-rendering macd causes duplicates
Same issue as #493.

Again thanks to @djmiley for spotting it.

djmiley:
Resolves #522 .


chrisprice:
Thanks @djmiley

PR #525, opened 2015-08-30
Add simple resize handles to navigator extent in Low Barrel example

### LINKED ISSUE #341: slider handles
The advanced example looks great. 

http://scottlogic.github.io/d3-financial-components/example.html

I think it would be even better if the slider section had handles as is common with other stock charts. 

image

jleft:
Resolves #341.

resize-handles

### DIFF
--- a/site/src/examples/low-barrel/index.js
+++ b/site/src/examples/low-barrel/index.js
@@ -206,6 +206,22 @@
                     ]);
                 }
                 return data;
+            })
+            .decorate(function(sel) {
+                var height = Math.abs(chart.yScale().range()[0] - chart.yScale().range()[1]);
+                sel.enter()
+                    .selectAll('.resize.e>rect, .resize.w>rect')
+                    .style('visibility', 'visible')
+                    .attr('y', height / 4)
+                    .attr('rx', 4)
+                    .attr('ry', 4);
+
+                // As a y scale is set on the brush (multi does this),
+                // the brush component resets the height of the rect on every redraw,
+                // as such it has to be overridden within the update selection,
+                // rather than the enter selection
+                sel.selectAll('.resize.e>rect, .resize.w>rect')
+                    .attr('height', height / 2);
             });
 
         chart.plotArea(multi);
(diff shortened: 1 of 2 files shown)


site/src/examples/low-barrel/index.js
```diff
@@ -206,6 +206,18 @@
...
+            .decorate(function(sel) {
+                var height = Math.abs(chart.yScale().range()[0] - chart.yScale().range()[1]);
+                sel.enter()
+                    .selectAll('.resize.e>rect, .resize.w>rect')
+                    .style('visibility', 'visible')
+                    .attr('y', height / 4)
+                    .attr('rx', 4)
+                    .attr('ry', 4);
+
+                sel.selectAll('.resize.e>rect, .resize.w>rect')
```
ColinEberhardt: Is there any reason why the height is set within the update selection, whereas the other properties of the `rect` are set in the enter selection?

ColinEberhardt:
Another good decorate example - just one question (above).

site/src/examples/low-barrel/index.js
jleft: As a [y scale is set](https://github.com/mbostock/d3/blob/5b981a18db32938206b3579248c47205ecc94123/src/svg/brush.js#L81-L85) on the brush ([multi does this](https://github.com/ScottLogic/d3fc/blob/6cafee1565501c7eab8a254a8f53d4f2ddd0b327/src/series/multi.js#L51)), D3's brush component [resets the height of the `rect`](https://github.com/mbostock/d3/blob/5b981a18db32938206b3579248c47205ecc94123/src/svg/brush.js#L142) on every redraw, so it has to overridden within the update selection, rather than the enter selection.

ColinEberhardt: Thanks, that makes sense. Probably worth adding a comments as it isn't obvious behaviour?

jleft: No problem. Agreed, definitely a good idea - updated with a comment.

ColinEberhardt:
Thanks @jleft

PR #526, opened 2015-09-01
Updated to use the public API for css-layout

ColinEberhardt:
The `css-layout` project now has a properly defined public API, so we don't need to reference `src\Layout.js` directly any more. See this PR:

https://github.com/facebook/css-layout/pull/108

The public API performs a 'fill nodes' step to add the required `layout` property with default values.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -122,25 +122,7 @@ module.exports = function (grunt) {
                     {
                         expand: true,
                         cwd: 'visual-tests/src/site/assets/',
-                        src: ['**/*.css', '**/*.js'],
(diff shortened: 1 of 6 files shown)


Gruntfile.js
```diff
@@ -133,8 +133,8 @@ module.exports = function (grunt) {
                     },
                     {
                         expand: true,
-                        cwd: 'node_modules/css-layout/src/',
-                        src: ['Layout.js'],
+                        cwd: 'node_modules/css-layout/dist/',
```
chrisprice: I know this is a pre-existing problem but can the d3, css-layout and d3fc dependencies be collapsed into one block?

PR #529, opened 2015-09-03
dataJoin inserts elements in an order consistent with the bound data

### LINKED ISSUE #528: Order of elements created by data join should match the order of their corresponding datums
The elements in data join's enter selection are appended, so their order may not correspond to the order of the data.

Rather than [appending](https://github.com/mbostock/d3/wiki/Selections#append) the elements, they could be [inserted](https://github.com/mbostock/d3/wiki/Selections#insert) (using `selection.insert()`), to maintain order.

> For enter selections, the before selector may be omitted, in which case entering elements will be inserted immediately before the next following sibling in the update selection, if any. This allows you to insert elements into the DOM in an order consistent with bound data. Note, however, the slower selection.order may still be required if updating elements change order.

Source: https://github.com/mbostock/d3/wiki/Selections#insert

---

This would be particularly useful when the z-index is important, for example the ordering of series in multi.

jleft:
Resolves #528

### DIFF
--- a/src/util/dataJoin.js
+++ b/src/util/dataJoin.js
@@ -45,8 +45,14 @@
 
             // enter
             // when container is a transition, entering elements fade in (from transparent to opaque)
+            // N.B. insert() is used to create new elements, rather than append(). insert() behaves in a special manner
+            // on enter selections - entering elements will be inserted immediately before the next following sibling
+            // in the update selection, if any.
+            // This helps order the elements in an order consistent with the data, but doesn't guarantee the ordering;
+            // if the updating elements change order then selection.order() would be required to update the order.
+            // (#528)
             var enterSelection = updateSelection.enter()
-                .append(element) // <<<--- this is the secret sauce of this whole file
+                .insert(element) // <<<--- this is the secret sauce of this whole file
                 .attr(attr)
                 .style('opacity', effectivelyZero);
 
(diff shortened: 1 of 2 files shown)


src/util/dataJoin.js
```diff
@@ -46,7 +46,7 @@
             // enter
             // when container is a transition, entering elements fade in (from transparent to opaque)
             var enterSelection = updateSelection.enter()
-                .append(element) // <<<--- this is the secret sauce of this whole file
+                .insert(element) // <<<--- this is the secret sauce of this whole file (#528)
```
jleft: Is it sufficient referencing 528 here, or would it be better to add a comment with explanation?

ColinEberhardt: Personally I'd prefer a comment.

ColinEberhardt:
Thanks for this and thumbs up for unit tests!

I had to read the selection.insert documentation quite closely to see how this change worked. The special behaviour of insert in the case of enter selections is far from obvious!

src/util/dataJoin.js
chrisprice: I'd explicitly say we're using the special enter-only insert rather than append to help with ordering

jleft: Updated with a comment - hopefully the explanation is clear enough?

jleft:
> Thanks for this and thumbs up for unit tests!

No problem, thank you for looking over it :smiley:

chrisprice:
:+1:

PR #532, opened 2015-09-04
Use point series in crosshair

### LINKED ISSUE #491: Use point series in crosshair for trackball
It'll tidy things up a little I think.

chrisprice:
Fixes #491


ColinEberhardt:
Neat :-)

PR #533, opened 2015-09-04
Ensures that the version numbers for cdnjs are the most recent

ColinEberhardt:
Assemble was missing the string manipulation functionality I required, so I added `slice` to handlebars helpers:

https://github.com/assemble/handlebars-helpers/pull/194

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "grunt-contrib-watch": "^0.6.1",
     "grunt-jscs": "^1.8.0",
     "grunt-version": "^1.0.0",
+    "handlebars-helpers": "git://github.com/ColinEberhardt/handlebars-helpers.git",
     "jquery": "^1.11.2",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -35,6 +35,7 @@
     "grunt-contrib-watch": "^0.6.1",
     "grunt-jscs": "^1.8.0",
     "grunt-version": "^1.0.0",
+    "handlebars-helpers": "git://github.com/ColinEberhardt/handlebars-helpers.git",
```
chrisprice: I think you can shorten this to just ColinEberhardt/handlebars-helpers

chrisprice:
One comment but otherwise :+1:

package.json
ColinEberhardt: > I think you can shorten this to just ColinEberhardt/handlebars-helpers

Yep - that worked.

Issue #520, opened 2015-08-27
No decorate property on indicator renderers


jleft:
Many of the components expose a decorate property, but the indicator renderers don't. I think it makes sense for the renderers to have this functionality too.

chrisprice:
Agreed, the renderers are essentially specialised series so it definitely makes sense. I assume it was just an oversight.

chrisprice:
#534 allows decoration but would someone be able to provide some useful examples of decoration?

chrisprice:
Assuming this now works.

PR #534, opened 2015-09-04
Allow decoration of indicator renderers

chrisprice:
Adds the decorate functionality asked for in #520 but I'd like to see some examples before that issue is closed.


ColinEberhardt:
I'll try this one out first before merging it - I've not used a multi-series decorator before so want to see how that works!

ColinEberhardt:
Well, I've tried decorating the indicators and it definitely works - couldn't find anything terribly useful to do via decoration, but it certainly works!

PR #536, opened 2015-09-07
Update order of cdnjs links in site Getting Started page

jleft:
Just a minor update to list d3fc's dependencies before d3fc itself (I think it should be this way around!).


ColinEberhardt:
Thanks - well spotted.

PR #542, opened 2015-09-11
Column series

ColinEberhardt:
Adds a column series

I did try to make a bar series with an 'orient' property, but the resulting API and code was a mess!

Note, before this is merged we need to swap `bar` <=> `column`, although I think it would be easier to review this change before renaming.


src/series/column.js
```diff
@@ -0,0 +1,120 @@
...
+            yValue = function(d, i) { return d.date; },
+            x0Value = d3.functor(0),
+            columnWidth = fc.util.fractionalBarWidth(0.75);
+
+        var dataJoin = fc.util.dataJoin()
+            .selector('g.column')
+            .element('g')
+            .attr('class', 'column');
+
+        var yValueScaled = function(d, i) { return yScale(yValue(d, i)); };
```
chrisprice: This seems to duplicate `y`, just use `y` in its place?

src/series/column.js
```diff
@@ -0,0 +1,120 @@
...
+                var y = function(d, i) { return yValueScaled(d, i); },
+                    x1 = function(d, i) { return xScale(x1Value(d, i)); },
+                    x0 = function(d, i) { return xScale(x0Value(d, i)); };
+
+                g.enter()
+                    .attr('transform', function(d, i) {
+                        return 'translate(' + x0(d, i) + ', ' + y(d, i) + ')';
+                    })
+                    .append('path')
+                    .attr('d', function(d) { return pathGenerator([d]); });
```
chrisprice: What do you think of adding a `pathGenerator.single`/`pathGenerator.one`/`pathGenerator.once` method (as a separate improvement)?

src/series/column.js
```diff
@@ -0,0 +1,120 @@
...
+
+                var filteredData = data.filter(function(d, i) {
+                    return x0Value(d, i) !== undefined &&
+                        x1Value(d, i) !== undefined &&
+                        yValue(d, i) !== undefined;
+                });
+
+                var g = dataJoin(this, filteredData);
+
+                var height = columnWidth(filteredData.map(yValueScaled));
```
chrisprice: I'm really confused about the orientation in this line, are the variables named appropriately?

src/series/column.js
```diff
@@ -0,0 +1,120 @@
...
+
+                g.enter()
+                    .attr('transform', function(d, i) {
+                        return 'translate(' + x0(d, i) + ', ' + y(d, i) + ')';
+                    })
+                    .append('path')
+                    .attr('d', function(d) { return pathGenerator([d]); });
+
+                g.each(function(d, i) {
+                    pathGenerator.width(x0(d, i) - x1(d, i));
```
chrisprice: Could the `g.each` be removed? I think `x0(d, i) - x1(d, i)` can be replaced with a callback, same for `'translate(' + x1(d, i) + ', ' + y(d, i) + ')'`.

src/series/column.js
```diff
@@ -0,0 +1,120 @@
...
+        };
+        column.columnWidth = function(x) {
+            if (!arguments.length) {
+                return columnWidth;
+            }
+            columnWidth = d3.functor(x);
+            return column;
+        };
+
+        d3.rebind(column, dataJoin, 'key');
```
chrisprice: Is there a use-case for rebinding the `fc.svg.bar` align properties?

ColinEberhardt:
> Is there a use-case for rebinding the fc.svg.bar align properties?

Along one of the axes there might be a use-case, I've rebound and aliased simply as `align`, so for a bar series you can align the bars left, right or center. Might be of use to someone!

> Could the g.each be removed?

Good call - I found that quite a few things could be simplified, including the removal of the transition.

> I'm really confused about the orientation in this line, are the variables named appropriately?

That's a tricky one! It all depends on whether you consider the width and height of columns and bars to be transposed. Personally I'd still use the word 'width' to describe the 'thickness' of columns - i.e their size along the cross-axis.

src/series/column.js
ColinEberhardt: Sounds good to me - we're using it like this all over the place.

chrisprice: Done #549 although I've just spotted that these methods are all tagged on the enter selection, should they be on the update instead?

ColinEberhardt: They are on both because when the bar is rendered via the enter selection it has a zero height. This means that if it is used with a transition it will grow upwards from y0 to y1.

It's a little unusual in that our other series types do not consider transitions - but I think it is a neat idea. Creating decent transitions with D3 takes a bit of thought.

Might be worth adding a quick comment as this behaviour probably isn't that obvious from the code.

chrisprice: Of course, I think I was suffering from a bit of email diff blindness...
I'll have a proper look tomorrow

chrisprice:
I've taken another look and it looks good to me. 

For the naming, a quick hunt around suggests people normally leave it at width so happy with that.

Were you going to add a comment or do you want to just merge?

ColinEberhardt:
I'll add a comment before merging, although there was one last thing to do ...

We have bar and column the wrong way round. So a rename is required (and a breaking change).

ColinEberhardt:
I was about to launch into a rename, but am having second thoughts. From looking around, it would appear that a [great many people consider bar charts to be vertical](https://www.google.co.uk/search?q=bar+chart&client=safari&rls=en&source=lnms&tbm=isch&sa=X&ved=0CAcQ_AUoAWoVChMIjqOD8bb3xwIVgr0aCh27XQUT&biw=1366&bih=706).

A few options:
1. Go ahead with the bar <=> column rename, which will be a breaking change
2. Come up with a different name for the new series type, `horizontalBar` perhaps?
3. Merge the two into `fc.series.bar` adding an `orient` property. I like this as a concept, but it does require a bit of juggling with the x & y properties. 

@chrisprice any thoughts?

chrisprice:
I like 3 the most but do share your concerns about the complexity. Would it be possible to do both 2 & 3, where 3 is just an adapter?

ColinEberhardt:
> I like 3 the most but do share your concerns about the complexity. Would it be possible to do both 2 & 3, where 3 is just an adapter?

Well (3) actually has a couple of sub-options:

3.1. When the `orient` property is flipped, x becomes y and vice-versa.
3.2. The series has properties of `x`, `y`, `x0`, `y0`, `y1`, `x1` - and which are 'active' depends upon the orientation.

I think 3.2 could be the most elegant solution (and is no a breaking change :-) ). When switching from column to bar you will have to change your scales regardless, so it doesn't add much extra effort to flip the relevant accessors.

They should throw errors if an accessor is set that isn't relevant for a given orientation. Although this would only be possible at the point where the series is rendered.

chrisprice:
What about `x`/`y` aliased as `x1`/`y1` (in the existing fashion) and `x0`/`y0`? `orient` would then set these to appropriate defaults (e.g. 'horizontal' sets `x0 = xScale.domain()[0], x1 = (d) -> d.date, y0 = y1 = (d) -> d.close`) which you could then override.

ColinEberhardt:
@chrisprice - I've combined bar and column into a bar series with horizontal / vertical orientation. I think this works quite nicely (once I managed to wrap my head around which properties to transpose).

src/series/bar.js
```diff
@@ -86,25 +145,32 @@ export default function() {
...
         return bar;
     };
-    bar.y0Value = function(x) {
+    bar.x0Value = function(x) {
         if (!arguments.length) {
-            return y0Value;
+            return x0Value;
         }
-        y0Value = d3.functor(x);
+        x0Value = x;
```
chrisprice: Any reason `y0Value` is a functor and `x0value`/`yValue`/`xValue` aren't?

src/series/bar.js
```diff
@@ -4,62 +4,121 @@ import fractionalBarWidth from '../util/fractionalBarWidth';
 import {noop} from '../util/fn';
 import svgBar from '../svg/bar';
 
+// the bar series renders a vertical (column) or horizontal (bar) series. In order
```
chrisprice: Nitpick: capital letter?

src/series/bar.js
ColinEberhardt: > Any reason y0Value is a functor and x0value/yValue/xValue aren't?

Because my code snippet for properties doesn't add functors ;-)

Good spot. I've made `x0Value` a functor property in line with the current bar implementation. I can 't see any reason for using a functor for `xValue` or `yValue`.

chrisprice:
Looks good to me, just need to add some drop shadows and linear gradients... :wink:

ColinEberhardt:
> just need to add some drop shadows and linear gradients...

I'll raise an issue for adding d3fc themes ...

PR #548, opened 2015-09-13
Removed the `orient` property exclusion from the axis rebinding

### LINKED ISSUE #547: Yahoo FInance Chart example is broken
http://d3fc.io/examples/yahoo-finance-chart/

ColinEberhardt:
Fixes #547 

This bug was introduced here:

https://github.com/ScottLogic/d3fc/pull/518/files#diff-6f455d17dcde9ea880da5741db76089eL110

I hadn't spotted that linearTimeSeries was rebinding `orient`. It didn't really feel logic to be rebinding that property, but clearly I was wrong as I was using it myself ;-)

### DIFF
--- a/src/chart/linearTimeSeries.js
+++ b/src/chart/linearTimeSeries.js
@@ -99,12 +99,10 @@
             yNice: 'nice'
         });
 
-        var axisExclusions = [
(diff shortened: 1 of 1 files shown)


chrisprice:
:+1:

Issue #415, opened 2015-06-15
Make build faster


ColinEberhardt:
This should make the build a bit faster!

chrisprice:
Make sure we fix #370 as part of this.

chrisprice:
```
Problem: Grunt
Solution: Gulp
Problem: Gulp
```

ColinEberhardt:
Fair point. The issue should really just be 'make the build faster' - I don't really care which took it uses!

chrisprice:
Done :)

ColinEberhardt:
Concrete things to consider:

 + The CSS is built / minified before JS unit tests are run, which means that the build is slow to report breaking changes
 + Do the visual tests need to use the minified / bundled output?
 + BrowserSync?

jleft:
I expect that [time-grunt](https://github.com/sindresorhus/time-grunt) would be useful.

This is the output when running `grunt ci`:

![grunt-ci-time](https://cloud.githubusercontent.com/assets/2017615/9809893/195b2e5a-5864-11e5-9dcf-0f06747d01b6.PNG)

chrisprice:
I've had a go at running the jasmine tests using jsdom rather than phantom (jsdom plays nice with Windows since node 4) and that seems to make a big difference to the `jasmine:test` task. I've also spotted we're `jshint`/`jscs`-ing twice and there's no need to run `uglify` for dev builds.

That adds up to about 50% of the time taken, I'd suggest we start there.

chrisprice:
I also tried to use `grunt-newer` to only lint changed files but I couldn't get it to work. It looks like it's been abandoned as a project though so I only mention it so that no one else has the same problem.

ColinEberhardt:
The build is certainly a lot faster at reporting unit test failures, which is good :-)

ColinEberhardt:
I'm happy with the build speed now :-)

PR #550, opened 2015-09-14
Run tests using jsdom and tidy up gruntfile

chrisprice:
Helps with #415


ColinEberhardt:
I like this simplified grunt tasks - nice job.

Gruntfile.js
```diff
@@ -309,12 +317,12 @@ module.exports = function(grunt) {
...
 
         umd: {
-            dist: {
+            components: {
                 options: {
                     src: 'dist/d3fc.js',
                     objectToExport: 'fc',
                     deps: {
-                        'default': ['d3', 'cssLayout'],
+                        'default': ['d3', 'computeLayout'],
```
ColinEberhardt: well spotted - I guess this was wrong previously?

chrisprice: Yea, my bad, I didn't test the UMD plugin version.

PR #551, opened 2015-09-14
Esnext modules

chrisprice:
Definitely not just trying to get this in before your bar/column changes. No sir. Definitely didn't cross my mind. What changes?

This is on top of #550, which failed to build so I'll fix that then rebase this one again.

### DIFF
--- a/src/data/data.js
+++ b/src/data/data.js
@@ -1,0 +1,7 @@
+import feed from './feed/feed';
+import random from './random/random';
+
+export default {
+    feed: feed,
+    random: random
+};
--- a/.jscsrc
+++ b/.jscsrc
@@ -1,5 +1,6 @@
 {
     "preset": "google",
+    "esnext": true,
 
     "disallowKeywordsOnNewLine": ["else"],
     "disallowMultipleLineBreaks": null,
--- a/.jshintrc
+++ b/.jshintrc
@@ -7,7 +7,6 @@
     "latedef": true,
     "newcap": true,
     "noarg": true,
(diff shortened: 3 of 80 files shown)


chrisprice:
Fixed.

ColinEberhardt:
> What changes?

Just one or two!

One minor thing (probably not worth changing):

    import _dataJoin from '../util/dataJoin';

If I was importing a module and found it collided with the name I wanted to use for a variable, I'd probably pick a different name for the variable rather than change the name I used for the imported module.

src/data/data.js
```diff
@@ -0,0 +1,13 @@
+import coinbase from './feed/coinbase';
```
ColinEberhardt: Is it worth splitting this into `feed.js` and `random.js` under their respective folders? My first thought was 'why is this file structured differently?'

chrisprice: Laziness... fixed!

chrisprice:
On the naming issue, I'd like to see us have a standard for it e.g. -
```
import {util_dataJoin} from '../util/dataJoin';

var dataJoin = util_dataJoin()
  //...
```

I don't like to have to think about it and it would be good to have some context when looking at whether something is a factory or an instance. 

That said, I don't think it's necessary to do in this PR. We could just have a rule to fix it up any time a file is touched?

ColinEberhardt:
> That said, I don't think it's necessary to do in this PR. We could just have a rule to fix it up any time a file is touched?

Agreed - it's not that important right now.

PR #556, opened 2015-09-15
Added cartesian chart documentation

ColinEberhardt:
Also, tweaked the CSS, y-axis labels are rotated so require a 'hanging' baseline regadless of whether they are left or right aligned.
We also need to juggle around the overflow behaviour of SVG a little!

### DIFF
--- a/site/src/components/chart/cartesianChart.md
+++ b/site/src/components/chart/cartesianChart.md
@@ -1,0 +1,114 @@
+---
+layout: component
+title: Cartesian Chart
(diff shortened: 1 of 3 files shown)


chrisprice:
:+1: 

I think the branch protection feature must prevent non-ff merges to master though, so this needs updating before it can be merged.

ColinEberhardt:
> I think the branch protection feature must prevent non-ff merges to master though

Yes, fast forwarding the branch allowed me to merge.

PR #561, opened 2015-09-16
Use exponentialMovingAverage calculator

### LINKED ISSUE #546: fc.indidcator.algorithm.exponentialMovingAverage uses d3.mean
https://github.com/ScottLogic/d3fc/blob/master/src/indicator/algorithm/exponentialMovingAverage.js#L6

I would have expected it to use `fc.indicator.algorithm.calculator.exponentialMovingAverage`?

chrisprice:
Fix #546.

Also demonstrates my idea for a module alias convention.

### DIFF
--- a/src/indicator/algorithm/exponentialMovingAverage.js
+++ b/src/indicator/algorithm/exponentialMovingAverage.js
@@ -1,14 +1,13 @@
-import calculator from './calculator/slidingWindow';
 import d3 from 'd3';
-import merge from './merge';
+import indicator_algorithm_calculator_exponentialMovingAverage from './calculator/exponentialMovingAverage';
+import indicator_algorithm_merge from './merge';
 
 export default function() {
 
-    var ema = calculator()
-            .accumulator(d3.mean)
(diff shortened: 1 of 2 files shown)


src/indicator/algorithm/exponentialMovingAverage.js
```diff
@@ -1,14 +1,13 @@
...
 import d3 from 'd3';
-import merge from './merge';
+import indicator_algorithm_calculator_exponentialMovingAverage from './calculator/exponentialMovingAverage';
+import indicator_algorithm_merge from './merge';
 
 export default function() {
 
-    var ema = calculator()
-            .accumulator(d3.mean)
+    var ema = indicator_algorithm_calculator_exponentialMovingAverage()
```
ColinEberhardt: This does feel a little verbose, the value of `indicator_algorithm_calculator_exponentialMovingAverage` versus simply `exponentialMovingAverage` is that it would allow you to locate the source of the module a little more quickly, but it's not much effort to just locate the module via the import statement.

ColinEberhardt:
Concise, yet requires no thought - I like it :-)

PR #562, opened 2015-09-18
Extends cartesian chart to support ordinal scale

ColinEberhardt:


src/chart/cartesianChart.js
```diff
@@ -23,28 +23,43 @@ export default function(xScale, yScale) {
         plotArea = line(),
         decorate = noop;
 
+    // The axes have a baseline that is defined in the coordinate system of the
+    // opposing axis. If the opposing axis is linear, the baseline is expressed
```
chrisprice: I think the comment might be misleading, does `linear` just mean `non-ordinal`?

src/chart/cartesianChart.js
```diff
@@ -23,28 +23,43 @@ export default function(xScale, yScale) {
...
 
-    var axesSeries = multi()
-        .series([xAxis, yAxis]);
-
     var containerDataJoin = dataJoin()
         .selector('svg.cartesian-chart')
         .element('svg')
         .attr({'class': 'cartesian-chart', 'layout-css': 'flex: 1'});
 
+    // Ordinal and linear scales have different methods for setting the range. This
```
chrisprice: Same as above

src/chart/cartesianChart.js
```diff
@@ -136,14 +154,21 @@ export default function(xScale, yScale) {
...
+            setScaleRange(xScale, [0, plotAreaContainer.layout('width')]);
+            setScaleRange(yScale, [plotAreaContainer.layout('height'), 0]);
 
             // render the axes
-            var axesContainer = svg.select('.axes-container');
-            axesSeries.xScale(xScale)
-                .yScale(yScale);
-            axesContainer.call(axesSeries);
+            xAxis.xScale(xScale)
+                .yScale(isOrdinal(yScale) ? d3.scale.identity() : yScale);
```
chrisprice: Would it be possible to always make the axes accept an identity axis in the cross direction?

src/util/rebind.js
```diff
@@ -49,16 +49,35 @@ export function rebindAll(target, source, prefix, exclusions) {
         exclusions = args.slice(3);
     }
 
-    exclusions.forEach(function(property) {
+    var stringExclusions = exclusions.filter(function(exclusion) {
```
chrisprice: I was going to make a few suggestions here then got carried away and [created a PR instead](https://github.com/ColinEberhardt/d3-financial-components/pull/1)

visual-tests/src/test-fixtures/chart/cartesianChart.js
```diff
@@ -34,10 +35,16 @@
 
         row.append('td')
             .append('input')
-            .attr({'type': 'text'})
+            .attr('type', function(d) { return d.type || 'text'; })
             .on('blur', function(d) {
-                d.value = this.value;
-                render();
+                d.value = this.type === 'checkbox' ? this.checked : this.value;
```
chrisprice: Minor: extract function and re-use it for both `blur` & `click`?

visual-tests/src/test-fixtures/chart/cartesianChart.js
```diff
@@ -73,11 +93,11 @@
         }
 
         // Create the line and area series
```
chrisprice: Delete?

ColinEberhardt:
I've had a go at the changes we discussed, using an identity cross-scale for the axis series. I think it makes the code a just a little bit cleaner, in that it removes a couple of the 'if ordinal' checks. It also means that you can set an axis baseline where the cross-axis is ordinal.

src/chart/cartesianChart.js
```diff
@@ -19,32 +19,53 @@ export default function(xScale, yScale) {
         },
         yLabel = '',
         xLabel = '',
+        xBaseline = noop,
```
chrisprice: Are you using `noop` rather than null because it's being run through `d3.functor`? If so, it's probably worth adding a comment explaining that (it took me a little while to work that out).

chrisprice:
Just the one small comment otherwise :+1: from me.

src/chart/cartesianChart.js
chrisprice: Actually thinking again about it, `null` seems a more consistent (with the rest of d3) value. Maybe it could be special cased in the property instead?

ColinEberhardt: > Are you using noop rather than null because it's being run through d3.functor

Yes, that was my original thinking, but I've just realised, the default value isn't functor-ed, so `null` works just fine.

ColinEberhardt:
Thanks for the review. I've just merged this in.

PR #566, opened 2015-09-25
Cartesian chart transitions

### LINKED ISSUE #563: Support transitions with the chart components
Both our chart components cannot be called on a transition because they use various selection methods which are not available on transitions (`attr`, `html`, ...). Which makes sense.

However, as the various series / axis components support transitions it's a shame that you cannot make use of this feature when using the components in conjunction with charts.

I quickly hacked together a modified chart component that optionally converts the selections used for the axes / series into transitions:

https://github.com/ColinEberhardt/d3-financial-components/commit/15595130b70b3628a5ad7c4ba96986db085f7230

The char component exposes a `transitionDuration` property which is used to determine whether to use transitions or selections.

ColinEberhardt:
Fixes #563 

After a bit of digging I found the underlying problem, our `layout` component is exposed on selections but not transitions. This means that if you call a component that uses `layout` from a transition, it fails.

This change adds `layout` to transitions. Also, transitions do not provide a 'getter' for the `attr` method, which is why this `layout` now drops down to the underlying DOM API.

### DIFF
--- a/src/layout/layout.js
+++ b/src/layout/layout.js
@@ -99,7 +99,7 @@ function _layout() {
     return layout;
 }
 
-d3.selection.prototype.layout = function(name, value) {
+function layoutAdapter(name, value) {
     var layout = _layout();
     var n = arguments.length;
     if (n === 2) {
@@ -109,7 +109,7 @@ d3.selection.prototype.layout = function(name, value) {
             this.call(layout);
         } else {
             // layout(name, value) - sets a layout- attribute
-            this.attr('layout-css', name + ':' + value);
+            this.node().setAttribute('layout-css', name + ':' + value);
         }
     } else if (n === 1) {
         if (typeof name !== 'string') {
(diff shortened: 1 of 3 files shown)


chrisprice:
Good sleuthing. Do you know why attr isn't exposed on transition selections? 

That's bot a leading question, I'm just curious what the split is.

ColinEberhardt:
Yeah, it was fun, watching things blow up in transitions timer functions and
much head scratching!
Transitions support the attr setter but not getter. I presume because the value
returned by the getter would change over time as the value is tweened which
would be quite a confusing API.

On Sat, Sep 26, 2015 at 9:53 a.m., ScottLogic/d3fc
<reply@reply.github.com>
wrote:
Good sleuthing. Do you know why attr isn't exposed on transition selections?

That's bot a leading question, I'm just curious what the split is.

—
[https://github.com/ScottLogic/d3fc/pull/566#issuecomment-143412586] .[https://github.com/notifications/beacon/ABDBfpIhiodhZgctTgEy9NhP7VrvGrqsks5o1lSLgaJpZM4GD92b.gif]

chrisprice:
Thanks, makes a lot of sense.

PR #575, opened 2015-09-30
[WIP] Stacked-bar refactor and offset-bar

ColinEberhardt:
This is a Work In Progress, but thought it was worth sharing my current thoughts and notes at this stage.

Open questions:
 - is `transposeCsv` a good name for the component which takes `d3.csv` and converts it into series (plural!)
 - transposeCsv adds the name for each series as an array property. This feels a bit ugly,
 but does retain a 2D array structure making it easier to datajoin.
 - The `yDomain` is a bit messy:

```
fc.util.extent(series, [function(d) { return 0; }, function(d) { return d.y; }])
```

This makes use of `fc.util.extent` to operate on arrays of arrays. It also ensures the domain starts from zero.
A couple of options spring to mind:

1) Create `fc.util.max` / `min` that also operate on arrays of arrays resulting in:

    [0, fc.util.max(series, function(d) { return d.y; }])]

2) Use a `d3.functor` to make it easier to provide the zero domain minimum.

    fc.util.extent(series, [0, function(d) { return d.y; }])

I quite like option (2).

Also, in keeping with the `rebindAll` var-arg API, this should also be possible

fc.util.extent(series, 0, function(d) { return d.y; });

TODO:
 - [x] Offset bar, the offset is currently applied via decorate, it should really be applied as an adapter around the xScale to keep decorate 'clean'.
 - [ ] Tests for transposeCsv
 - [ ] Look into the stackedBar decorate function, I'd prefer an easier way of colouring each series


chrisprice:
Random thought, why do you need transpose?

ColinEberhardt:
That's a very good question, it is currently used so that we can render each
data 'attribute' or column using a single series.
For stacked bar we could render each 'stack' independently, however it wouldn't
work so well for line or area series where there is some natural connection
between each row of data in the original CSV.
Thinking about it, I'm not sure how we would employ d3's stacked layout without
performing a transpose operation.

On Wed, Sep 30, 2015 at 4:04 p.m., ScottLogic/d3fc
<reply@reply.github.com>
wrote:
Random thought, why do you need transpose?

—
[https://github.com/ScottLogic/d3fc/pull/575#issuecomment-144439523] .[https://github.com/notifications/beacon/ABDBfiOv1_U5Rv5UELQxzsADfqp2TGYFks5o2_GZgaJpZM4GGUAy.gif]

ColinEberhardt:
Mike bostocks stacked line performs a transpose when mapping the data before
passing it to the stack layout:
http://bl.ocks.org/mbostock/3885211
However his stacked bar does not transpose for the reasons I mentioned earlier.
On Wed, Sep 30, 2015 at 4:04 p.m., ScottLogic/d3fc
<reply@reply.github.com>
wrote:
Random thought, why do you need transpose?

—
[https://github.com/ScottLogic/d3fc/pull/575#issuecomment-144439523] .[https://github.com/notifications/beacon/ABDBfiOv1_U5Rv5UELQxzsADfqp2TGYFks5o2_GZgaJpZM4GGUAy.gif]

ColinEberhardt:
Replaced by #600

PR #576, opened 2015-09-30
Layout now reflects the way that d3 selection methods work - using .each

### LINKED ISSUE #568: d3fc.io - jumbotron chart not scrolling
It's broken :-(

ColinEberhardt:
Fixes #568


src/layout/layout.js
```diff
@@ -99,44 +98,47 @@ function _layout() {
...
 }
 
-function layoutAdapter(name, value) {
-    // Quick bodge to fix #568
-    if (this.node() == null) {
-        return this;
+function layoutSelection(name, value) {
+    var argsLength = arguments.length;
+
+    // For layout(string), return the lyout value for the first node
```
chrisprice: Edge case but would this blow up if you do `x.enter().append('g').layout('width')`?

ColinEberhardt: No it doesn't - it follows the same pattern as D3's `attr`:

https://github.com/mbostock/d3/blob/master/src/selection/attr.js#L9

Just to be sure, I added an additional test

chrisprice:
Nice one

PR #579, opened 2015-10-01
Build performance

ColinEberhardt:
The build times are bugging me ...

Currently on master it takes ~13.5 seconds for a full visual-test + components build. This PR has a couple of things that reduce this:

 1. JIT task loader, this improves performance to ~11secs
 2. concurrently running assemble alongside component build, improves to ~7secs

Note that for (1) I've had to turn off jscs on the gruntfile as it doesn't appear to support `renameTask`. I've raised an issue with the plugin just to check.


ColinEberhardt:
3. Moved from jshint + jscs to eslint, now builds in ~5 secs

Although, if we do want to move to eslint there are a few rules like `indent` and `eqeqeq` that we might want to turn on and fix the related issues.

chrisprice:
See also #581

ColinEberhardt:
@chrisprice - I've updated this PR after merging your changes to remove assemble from the visual tests

chrisprice:
I'm not keen on the eqeqeq change (30eba17), as it changes the functionality of the code. Could we go for the "allow-null" option instead?

http://eslint.org/docs/rules/eqeqeq.html

ColinEberhardt:
> Could we go for the "allow-null" option instead?

Good point, I reviewed the cases where eqeqeq had been applied and I think you're right in those cases we could benefit from the flexibility.

I've updated the PR.

chrisprice:
That's a whole lot of whitespace in that diff (thank god for ?w=1).

PR #582, opened 2015-10-04
Update multi to always order series containers in the same order as specified in the series array

### LINKED ISSUE #537: Should multi series always order elements in the same order as specified in the series array?
It can be confusing that the ordering of the series (specified in an array passed to multi's series method) doesn't always match the order of the series' associated groups in the DOM. Should multi always guarantee this ordering?

Current behaviour:

```js
var mutli = fc.series.multi()
    .series([a, b, c]);

// call multi on selection. output:
// <g><!-- series a --></g>
// <g><!-- series b --></g>
// <g><!-- series c --></g>

multi.series([a, c, b]);

// call multi on selection
// B and C remain in the same order
// output:
// <g><!-- series a --></g>
// <g><!-- series b --></g>
// <g><!-- series c --></g>
```

Suggested change to behaviour:

```js
var mutli = fc.series.multi()
    .series([a, b, c]);

// call multi on selection. output:
// <g><!-- series a --></g>
// <g><!-- series b --></g>
// <g><!-- series c --></g>

multi.series([a, c, b]);

// call multi on selection
// B and C have swapped order - matching the order in the series array
// output:
// <g><!-- series a --></g>
// <g><!-- series c --></g>
// <g><!-- series b --></g>
```

If not, I think this behaviour should be documented.

I'm happy to either update multi and write some tests, and/or update the docs - which ever is preferable.

jleft:
Resolves #537

### DIFF
--- a/src/series/multi.js
+++ b/src/series/multi.js
@@ -56,6 +56,8 @@ export default function() {
                     .call(series);
             });
 
+            g.order();
+
             decorate(g);
         });
     };
--- a/tests/series/multiSpec.js
+++ b/tests/series/multiSpec.js
@@ -109,4 +109,96 @@ describe('multi', function() {
         expect(seriesA.calls.mostRecent().args[0].node()).toBe(seriesAContainer);
         expect(seriesB.calls.mostRecent().args[0].node()).toBe(seriesBContainer);
     });
+
+    describe('ordering of series containers', function() {
+
+        var seriesA, seriesB, seriesC;
+
+        function verifyContainers(element, series) {
+            expect(element.childNodes.length).toBe(series.length);
+            series.forEach(function(series, i) {
+                var seriesContainer = series.calls.mostRecent().args[0].node();
+                expect(element.childNodes[i]).toBe(seriesContainer);
+            });
+        }
+
+        beforeEach(function() {
+            seriesA = createSeriesSpy();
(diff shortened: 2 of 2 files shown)


chrisprice:
Thanks for the fix. Can't see any problems with the implementation but maybe a custom matcher would simplify the tests?

ColinEberhardt:
Thanks, I like the comprehensive unit tests too.

On Sun, Oct 4, 2015 at 3:47 p.m., ScottLogic/d3fc
<reply@reply.github.com>
wrote:
Thanks for the fix. Can't see any problems with the implementation but maybe a
custom matcher would simplify the tests?

—
[https://github.com/ScottLogic/d3fc/pull/582#issuecomment-145355015] .[https://github.com/notifications/beacon/ABDBfiIiPXfQXBi6Z264UIgC9lFITkbnks5o4TOLgaJpZM4GIlqY.gif]

jleft:
No problem, thanks.

>  Can't see any problems with the implementation but maybe a custom matcher would simplify the tests?

Sure, I'll update with a custom matcher.

jleft:
> Sure, I'll update with a custom matcher.

Done. (Not strictly using a custom matcher, but I think the tests are much simpler than before.)

ColinEberhardt:
Thanks @jleft - this looks really good.

PR #583, opened 2015-10-05
Allow fc.util.extent to support varargs

### LINKED ISSUE #571: fc.util.extent should support var-args
as per rebindAll: https://github.com/ScottLogic/d3fc/blob/master/src/util/rebind.js#L36

chrisprice:
Fixes #571


ColinEberhardt:
:+1:

PR #586, opened 2015-10-05
WIP: Base series

chrisprice:
Based on our discussions earlier, any thoughts?

It seems to save a significant amount of code and feels quite elegant (even in bar!).

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -109,7 +109,6 @@
         "func-style": 0,
         "indent": 2,
         "key-spacing": 2,
-        "linebreak-style": 2,
         "new-cap": 2,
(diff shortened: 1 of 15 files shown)


ColinEberhardt:
I agree, this looks really good - with bar it allows you to clearly see the bar rendering logic without the generic series boiler-plate.

ColinEberhardt:
I agree, this looks really good - with bar it allows you to clearly see the bar rendering logic without the generic series boiler-plate.

chrisprice:
Hm run into a few problems with ohlc/candlestick

chrisprice:
Superseded by #608

PR #592, opened 2015-10-08
Removed the Math.sqrt

jdunkerley:
No reason to run the Math.sqrt as looking just for the minimum distance

### DIFF
--- a/src/util/snap.js
+++ b/src/util/snap.js
@@ -22,17 +22,17 @@ export function noSnap(xScale, yScale) {
     };
 }
 
-export function pointSnap(xScale, yScale, xValue, yValue, data, pointDistance) {
+export function pointSnap(xScale, yScale, xValue, yValue, data, objectiveFunction) {
     // a default function that computes the distance between two points
-    pointDistance = pointDistance || function(x, y, cx, cy) {
+    objectiveFunction = objectiveFunction || function(x, y, cx, cy) {
         var dx = x - cx,
             dy = y - cy;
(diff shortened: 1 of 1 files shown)


ColinEberhardt:
Thanks @jdunkerley - as discussed, yes `Math.sqrt` isn't necessary. However, now that I have seen the diff I am less keen. The variable is named `pointDistance` and is documented as computing the distance between points. Removing `Math.sqrt` makes the code somewhat contradictory. 

Someone might revisit this code an re-add `Math.sqrt` in the belief that the omission was a bug!

Unless we have a real need to make this optimisation, I'd vote to keep the behaviour as-is for clarity.

Happy to get feedback from others @ScottLogic/d3fc

chrisprice:
Whilst it is currently called `pointDistance`, the result is not exposed so I don't see a problem with renaming it to something more appropriate. Exactly what that would be is harder to say but once that's done I don't see a problem with taking this PR.

Out of curiosity, was this something gleamed from profiling?

jdunkerley:
I was digging through how cross hairs and snap worked together as trying to create one which followed two charts simultaneously. 

Was getting slightly latent behaviour on a large time series so went nosing.

Doubt the square root makes any significant difference but seemed unnecessary.

Call it objectiveFunction?

chrisprice:
Name sounds fine to me, happy to see it merged with that.

jdunkerley:
updated the branch with the new name.

chrisprice:
Thanks!

PR #595, opened 2015-10-08
WIP Quandl feed

jdunkerley:
First implementation of feed from Quandl.

Seems to be working but a little bit more testing on different databases from Quandl needed.


src/data/feed/quandl.js
```diff
@@ -0,0 +1,146 @@
...
+        }
+        collapse = x;
+        return quandl;
+    };
+    // Function Used to Normalise the Quandl Column Name To Field Name, Return Null To Skip Field
+    quandl.columnNameMap = function(x) {
+        if (!arguments.length) {
+            return columnNameMap;
+        }
+        columnNameMap = x || _columnNameMap;
```
ColinEberhardt: I'm not keen on the following behaviour:

```
foo.setBar('baz') !== foo.getBar();
```

I think it is confusing for the user. Can you fall-back to your default implementation in the component function?

src/data/feed/quandl.js
```diff
@@ -0,0 +1,146 @@
...
+
+            data = data.dataset_data;
+            var colNames = data.column_names.map(columnNameMap);
+
+            data = data.data.map(function(d) {
+                var output = {};
+
+                output[colNames[0]] = new Date(d[0]);
+                for (var i = 1; i < colNames.length; i++) {
+                    if (colNames[i] != null) {
```
ColinEberhardt: How about using `Array.filter` when you create `colNames` removing the need for this repeated check.

src/data/feed/quandl.js
```diff
@@ -0,0 +1,146 @@
...
+        }
+        if (!descending) {
+            params.push('order=asc');
+        }
+        if (collapse != null) {
+            params.push('collapse=' + collapse);
+        }
+
+        var url = 'https://www.quandl.com/api/v3/datasets';
+        url += ('/' + database + '/' + dataset + '/').replace(/\/\//g, '');
```
ColinEberhardt: Just wondering what this string replacement is for?

ColinEberhardt:
A few minor comments, but other than that looks really good to me. Would be good to have a few unit tests to test (and illustrate) the purpose of the column name mapping regex.

src/data/feed/quandl.js
jdunkerley: Yep will amend to that. 

Was thinking of still setting it initially to the default implementation as then you can get access to it from outside easily.

src/data/feed/quandl.js
jdunkerley: No real need for it, meant to trim off leading and trailing '/' in the database or dataset.

chrisprice: I'm not sure it'll have the expected behaviour if the database has a
trailing slash. I'd just leave it up to the consumer to use the right values

jdunkerley: Will remove better to leave to consumer to get correct

src/data/feed/quandl.js
```diff
@@ -0,0 +1,146 @@
...
+        d3.json(url, function(error, data) {
+            if (error) {
+                cb(error);
+                return;
+            }
+
+            data = data.dataset_data;
+            var colNames = data.column_names.map(columnNameMap);
+
+            data = data.data.map(function(d) {
```
chrisprice: Could we add some more variables in to make this a little clearer? E.g.

```js
var mappedData = dataset_data.data.map(function(d) {
```

src/data/feed/quandl.js
```diff
@@ -0,0 +1,146 @@
+import d3 from 'd3';
+
+//  https://www.quandl.com/docs/api#datasets
+export default function() {
+
+    var _columnNameMap = function(colName) {
+        colName = colName.replace(/[\s()-.]/g, '');
```
chrisprice: I would leave this to the consumer as well, it might be that removing these characters leads to duplicate keys

src/data/feed/quandl.js
```diff
@@ -0,0 +1,146 @@
...
+        return colName.substr(0, 1).toLowerCase() + colName.substr(1);
```
chrisprice: Minor style niggle -
```js
return colName[0].toLowerCase() + colName.substr(1);
```

src/data/feed/quandl.js
jdunkerley: Sure, the default mapping would still sort Date to date and all fields for an OHLC data set so consumer would probably be doing post processing anyway.

jdunkerley:
Closing this PR will open a new one once unit tests etc done

PR #599, opened 2015-10-09
Fix Windows build

tlsim:
Update rollup to get Luke's Windows fixes.
ESLint complains about CRLF, but we use .gitattributes.


chrisprice:
Thanks

PR #600, opened 2015-10-11
Stacked and grouped bars

ColinEberhardt:
Replaces #575 

Changes since the previous PR:

 + transposeCsv is not spreadCsv and does not incorporate stacking
 + offsetBar has become groupedBar. Also, the implementation has been updated so that the offset for each bar is no longer applied via decoration (which feels a bit cleaner).


src/util/rebind.js
```diff
@@ -34,6 +34,8 @@ function capitalizeFirstLetter(str) {
  * be rebound.
  */
 export function rebindAll(target, source, prefix, exclusions) {
+    prefix = typeof prefix !== 'undefined' ? prefix : '';
```
ColinEberhardt: This doesn't look right - I don't think I have touched rebind. Might be a bad merge.

I'll look into this.

src/data/spreadCsv.js
```diff
@@ -0,0 +1,49 @@
+
+// the D3 CSV loader / parser converts each row into an object with property names
+// derived from the headings in the CSV. The spreadCsv component converts this into an
+// array of series, one per column
+export default function spreadCsv() {
+
+    var xValueKey = '';
+    var yValue = function(row, key) {
+        // D3 CSV returns all keys as strings, this converts them to numbers
```
chrisprice: keys = values?

src/data/spreadCsv.js
```diff
@@ -0,0 +1,49 @@
...
+                return key !== xValueKey;
+            })
+            .map(function(key) {
+                var series = data.map(function(row) {
+                    return {
+                        x: row[xValueKey],
+                        y: yValue(row, key)
+                    };
+                });
+                series.name = key;
```
chrisprice: `series.key`?

src/data/spreadCsv.js
```diff
@@ -0,0 +1,49 @@
+
+// the D3 CSV loader / parser converts each row into an object with property names
+// derived from the headings in the CSV. The spreadCsv component converts this into an
+// array of series, one per column
+export default function spreadCsv() {
```
chrisprice: Just call it `spread`?

src/series/groupedBar.js
```diff
@@ -0,0 +1,72 @@
+import _bar from './bar';
```
chrisprice: `barSeries`

src/series/groupedBar.js
```diff
@@ -0,0 +1,72 @@
...
+import d3 from 'd3';
+import _dataJoin from '../util/dataJoin';
```
chrisprice: `dataJoinUtil`

src/util/rebind.js
ColinEberhardt: No, not a bad merge, just poor memory. This allows properties to be re-bound without a prefix, as tested here:

https://github.com/ScottLogic/d3fc/pull/600/files#diff-f6d6160b813bdc9d086071b5af5c3e5eR73

ColinEberhardt:
@chrisprice - thanks for the feedback. I've made those few naming tweaks and rebased onto master.

PR #606, opened 2015-10-13
Move low-barrel to cartesian chart and fix cartesian chart in IE

chrisprice:


Gruntfile.js
```diff
@@ -61,8 +61,9 @@ module.exports = function(grunt) {
             },
             site: {
                 src: [
-                    'node_modules/d3/d3.js',
                     'node_modules/css-layout/dist/css-layout.js',
+                    'node_modules/d3/d3.js',
+                    'node_modules/svg-innerhtml/svg-innerhtml.js',
```
ColinEberhardt: That's a useful shim!

I do wonder whether we should include css-layout and svg-innerhtml in our bundle? While the current approach is good for browserify / npm users, it is not do good for people who do not have a build.

svg-innerhtml isn't on cdnjs.

site/src/examples/low-barrel/index.md
```diff
@@ -2,34 +2,32 @@
...
-  </tr>
-  <tr class="navigator-row">
-    <td>
-      <svg class="navigator"></svg>
-    </td>
-    <td></td>
-  </tr>
-</table>
-
+<svg id="low-barrel">
```
ColinEberhardt: Good to see SVG flexbox being used here :-)

ColinEberhardt:
LGTM - main concern is about dependencies.

chrisprice:
I'm open to the idea of creating a bundle alongside the existing one?
However, I think we're only supporting that mode for convenience, it is
(/should) not be the default bundle

ColinEberhardt:
> I'm open to the idea of creating a bundle alongside the existing one?
However, I think we're only supporting that mode for convenience, it is
(/should) not be the default bundle

Yep, agreed, the bundled version should be in addition to our current distribution builds. It would be very useful for codepen / jsbin and just trying out the library.

It does mean that we need that bundled build and some documentation updates before this code is released.

chrisprice:
And here's me thinking I could get some work done on my talk this morning...

ColinEberhardt:
@chrisprice - it looks like the new bundle file has to be added here in order to automatically add it to cdnjs each time it is published to npm:

https://github.com/ColinEberhardt/cdnjs/blob/d7116c0d3523e88593f4e475d9ef38abaab5e79c/ajax/libs/d3fc/package.json#L28

chrisprice:
I've created the PR https://github.com/cdnjs/cdnjs/pull/5923

ColinEberhardt:
:dancers:

PR #608, opened 2015-10-13
Introduce xyBase in applicable series

chrisprice:
Also tidies up imports for all series.

### DIFF
--- a/src/series/area.js
+++ b/src/series/area.js
@@ -1,30 +1,21 @@
 import d3 from 'd3';
-import _dataJoin from '../util/dataJoin';
+import dataJoinUtil from '../util/dataJoin';
 import {noop} from '../util/fn';
+import xyBase from './xyBase';
 
 export default function() {
 
-    var decorate = noop,
-        xScale = d3.time.scale(),
-        yScale = d3.scale.linear(),
-        y0Value = d3.functor(0),
-        y1Value = function(d, i) { return d.close; },
-        xValue = function(d, i) { return d.date; };
+    var decorate = noop;
 
-    // convenience functions that return the x & y screen coords for a given point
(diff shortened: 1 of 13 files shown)


tests/series/areaSpec.js
```diff
@@ -18,12 +18,12 @@ describe('area', function() {
...
             .call(area);
 
-        expect(xValueSpy.calls.count()).toEqual(data.length);
-        this.utils.verifyAccessorCalls(xValueSpy, data);
-
-        // the defined call also invokes the y value accessors,
+        // the defined call also invokes the value accessors,
         // therefore they are invoked twice for each data point
 
+        expect(xValueSpy.calls.count()).toEqual(data.length * 2);
```
ColinEberhardt: just wondering why the xValue accessor is now invoked twice for each point?

src/series/area.js
```diff
@@ -1,30 +1,21 @@
...
 
 export default function() {
 
-    var decorate = noop,
-        xScale = d3.time.scale(),
-        yScale = d3.scale.linear(),
-        y0Value = d3.functor(0),
-        y1Value = function(d, i) { return d.close; },
-        xValue = function(d, i) { return d.date; };
+    var decorate = noop;
```
ColinEberhardt: Any reason not to add `decorate` to the base component?

ColinEberhardt:
A couple of minor comments, but otherwise LGTM

tests/series/areaSpec.js
chrisprice: As per the comment I changed defined to also check x/x0. It's a change to what we had but it seems the logical behaviour.

src/series/area.js
chrisprice: It didn't feel like it belonged. I thought about putting it on its own component but that seemed overkill.

Issue #607, opened 2015-10-13
DateTime Axis with DiscontinuityProvider doesn't cope with empty domain


jdunkerley:
If you have an empty data array the fc.util.extent function returns `[undefined, undefined]` passing this into a dateScale with a discontinuityProvider causes an error.

```javascript
var dateScale = fc.scale.dateTime()
    .domain(fc.util.extent([], 'date'))
    .discontinuityProvider(fc.scale.discontinuity.skipWeekends())
    .range([0, width]);
```

ColinEberhardt:
Thanks for the report, what behaviour would you expect in this context?

On Tue, Oct 13, 2015 at 1:33 PM, James Dunkerley <notifications@github.com>
wrote:

> If you have an empty data array the fc.util.extent function returns [undefined,
> undefined] passing this into a dateScale with a discontinuityProvider
> causes an error.
>
> var dateScale = fc.scale.dateTime()
>     .domain(fc.util.extent([], 'date'))
>     .discontinuityProvider(fc.scale.discontinuity.skipWeekends())
>     .range([0, width]);
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ScottLogic/d3fc/issues/607>.
>



-- 
Regards,
Colin E.

jdunkerley:
I think best to leave as undefined. 

Small change I think to the skipWeekends clampUp and clampDown functions.

So `skipWeekends.clampUp(undefined)` would return undefined.

ColinEberhardt:
Ah, I thought you were proposing a change of behaviour for d3.util.extent!

Your proposed change makes sense and makes the code more robust.

jdunkerley:
k will implement and PR

chrisprice:
Fixed by #609

PR #609, opened 2015-10-13
Skipweekend undefined fix

jdunkerley:
Small change to skipWeekends to fix issue #607


chrisprice:
Looks like a merge commit has snuck in there, please can you rebase that
out. Otherwise LGTM.

Issue #539, opened 2015-09-10
The website examples should more prominently link back to component pages


chrisprice:
Lots of links point to the examples but it's not obvious that each component page also contains examples.

chrisprice:
I think this is now in a much better state.

PR #622, opened 2015-10-15
Added better links to documentation examples from examples on the site.

bjedrzejewski:
As this was reported as issue here: https://github.com/ScottLogic/d3fc/issues/539 I decided to try fixing it. This is how it looks if you accept pull request:

http://pasteboard.co/1oGhQWgY.png

### DIFF
--- a/site/src/_layouts/example.hbs
+++ b/site/src/_layouts/example.hbs
@@ -53,6 +53,11 @@ layout: default
                     </li>
                 {{/is}}
               {{/each}}
+                <li class="{{#if this.isCurrentPage}}active{{/if}}">
+                    <a href="../components/introduction/1-getting-started.html">
+                        ...more in the Documentation
+                    </a>
+                </li>
             </ul>
           </li>
         {{/each}}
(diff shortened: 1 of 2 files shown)


chrisprice:
Sorry I don't think I made myself very clear in the issue. I was thinking
that the example descriptions should be reworded to be brief walkthroughs
of how to create the example referencing the documentation examples
(components) more thoroughly. However taking inspiration from your changes
maybe we could just trail the template with something like -

> For more usage examples see the [component documentation](example.com).

ColinEberhardt:
I'd like to have a more broad conversation about the structure of the site, create more of a long term vision ...

bjedrzejewski:
I agree that some broad conversation would be good. I intended this pull request to be a quick fix / conversation starter.

jdunkerley:
Think that would be a great idea. Walk throughs sound nice but good to get a better feel of how we want site to be

bjedrzejewski:
Shall I close this one then? Not to clutter the PRs?

chrisprice:
Yes I think we should but I'm not sure we actually came up with that  clear vision for the site, maybe we move that discussion onto slack.

PR #625, opened 2015-10-16
Stack bar fix

### LINKED ISSUE #623: Stacked bar broken on website
.... gives me something to do on the flight home ...

ColinEberhardt:
Fixes #623 

The issue was with `xyBase` which changed the 'defined' rules to be strictly numeric.

I've also re-instated 0.5 opacity for bollinger fill, however rather than making all area series 0.5 opacity, this is now just applied to bollinger.

### DIFF
--- a/src/series/xyBase.js
+++ b/src/series/xyBase.js
@@ -24,8 +24,8 @@ export default function() {
         return yScale(yValue(d, i));
     };
     base.defined = function(d, i) {
-        return x0Value(d, i) !== undefined && y0Value(d, i) !== undefined &&
-            xValue(d, i) !== undefined && yValue(d, i) !== undefined;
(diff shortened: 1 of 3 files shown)


src/series/xyBase.js
```diff
@@ -24,8 +24,8 @@ export default function() {
         return yScale(yValue(d, i));
     };
     base.defined = function(d, i) {
-        return !isNaN(x0Value(d, i)) && !isNaN(y0Value(d, i)) &&
```
chrisprice: Worth using `==`?

ColinEberhardt: Yeah - fair point. Changed to `!= null` to allow for undefined and null for 'undefined' values.

chrisprice:
:+1:

PR #626, opened 2015-10-16
Webdriver local - executed by 'grunt webdriver'

bjedrzejewski:
This is the first version of the webdriver local runner.

If you want to use it you need to edit:

user: 'x', //Username from https://www.browserstack.com/accounts/automate
key: 'x', //Access Key from https://www.browserstack.com/accounts/automate

in wdio.conf.js

Also, you need to be running BrowserStackLocal.exe with your Access Key as a parameter. At the moment, the test doesn't check anything other than trying to visit every page on the site.

I want to pull request so that others can try and help working / testing the test suite and CI. Next commits would be as @chrisprice suggested - checking for no console errors on selected browsers.


ColinEberhardt:
Thanks @bjedrzejewski - I'll take a proper look at this later. Out of interest, have you tried the browserstack-runner?

https://github.com/browserstack/browserstack-runner

Perhaps this removes the need to manually start-up the local browserstack client.

bjedrzejewski:
The runner looks interesting, I was thinking of something to avoid running the local client manually, this could be it. I need to fix the CI errors now.

bjedrzejewski:
Added a test that will break on any console logs. I can limit that to only 'severe', but as far as I know- no logs are permitted.

bjedrzejewski:
@chrisprice @ColinEberhardt I believe that this is ready to be merged into the main d3fc code. I have automatic testing on browserstack as a part of CI build. The only thing that may be necessary to be done by you is to encrypt BROWSERSTACK_USER and BROWSERSTACK_KEY as the encryption may not work through pull request.

chrisprice:
Just been trying to get this running on my local machine and I've realised it's not integrated into the grunt build, only the travis one. I have a strong preference for keeping the travis config as simple as possible so that you can reproduce the build locally. Therefore please can you remove as many modifications from the `.travis.yml` file as possible and instead add them to the `grunt ci` task.

bjedrzejewski:
This is quite tough to run from local machine as it needs internet connection and the site being hosted while this is running. It also can take significant amount of time on local machine. I don't think this should be run as a part of every developer build. You can still run it as a grunt webdriver if you have your machine set up correctly.

bjedrzejewski:
Also keeping the keys encrypted and working with CI is far from trivial- if we don't want to use Travic CI encryption help. Not sure how we would do it.

chrisprice:
Completely agree that this shouldn't be part of every developer build (in fact I can't think of much worse!). However, the the `grunt ci` task is only intended to be run by travis not by developers so we should be safe adding it to that. 

As for the user and access key, they should be configured using environmental variables. Locally you can set them in the console or terminal config, and travis will create them from the encrypted credentials (as it is currently configured to do).

bjedrzejewski:
The complications are still:
```
-wget https://www.browserstack.com/browserstack-local/BrowserStackLocal-linux-x64.zip
-unzip BrowserStackLocal-linux-x64.zip
-/BrowserStackLocal -v -onlyAutomate -forcelocal $BROWSERSTACK_KEY &
-sleep 3
```
This will also be needed in the grunt- not sure how to easily add that (at the moment I manually start the exe).

Second issue that I don't really know how to handle in grunt is:
```
- grunt site:serve &
- sleep 10
```
We need to serve the site but only for the duration of the test.

I thought that we are at the point where this works well in Travis CI and we could start using it and others could improve if the two problems mentioned here are fixed.

chrisprice:
For the tunnel [this project](https://www.npmjs.com/package/browserstacktunnel-wrapper) or one of its dependents is probably worth a look.

For the hosting the site in the background bit, it looks like that's the default behaviour of the `grunt-contrib-connect` plugin we're using -
> Note that this server only runs as long as grunt is running. Once grunt's tasks have completed, the web server stops. This behavior can be changed with the keepalive option, and can be enabled ad-hoc by running the task like grunt connect:keepalive.

> This task was designed to be used in conjunction with another task that is run immediately afterwards, like the grunt-contrib-qunit plugin qunit task.

bjedrzejewski:
Ok, I will try with that. Where would you advise users to keep their key and username - we will need these in the wrapper once added and later in the wdio.conf.js ideally via process.env.BROWSERSTACK_USER and process.env.BROWSERSTACK_KEY, so that no code change would be needed for Travis CI.

bjedrzejewski:
Actually when trying to run:
```
grunt.registerTask('ci', ['components', 'uglify:components', 'site', 'uglify:site', 'site:serve', 'webdriver']);
```
Grunt just waits after ```site:serve```. This is the original reason why I had to split the grunt task (I think ```site:serve``` needs its own process because of the ```watch```. If we don't have ```watch:site``` we have errors in the console.

chrisprice:
I think environment variables are the most appropriate place to store them.

chrisprice:
If I run -

`grunt site connect:site [another task]` 

The `site` task builds the site, the `connect:site` starts serving it (and keeps the server running) and whichever task I add at the end runs (with the server still running in the background). Would that work?

bjedrzejewski:
Problem is the watch:site. Without it the console is full of errors. This is the blocking one. We could explicitly ignore these error (liverunner) as one option.

chrisprice:
Yea of course. Is it easy to exclude these errors?

The other options is to have some way to conditionally include the livereload script but then we'd a flag to turn that on and off.

bjedrzejewski:
Lets start with excluding the errors and then we may think about conditional liveload. I will do that tomorrow.

bjedrzejewski:
@chrisprice I think you will like this latest solution. All you need to do is to set BROWSERSTACK_USER and BROWSERSTACK_KEY in your environment and Travis CI and then just grunt ci!

.travis.yml
```diff
@@ -14,6 +14,12 @@ env:
...
+  - secure: "AIpFlu67qB9vCjqaF6CgJtEzkPiGzxGyBv3d+8sDdT2Nr7VRE4jcIjfw5mczLQLs8oITljVFURjq4excy29eZGMjxYFwOVju41lCUBZAInE7tVyvcGr43hrf2RgTWmO3rvMm2Kp5/X3UVDAKT12hn+U6yCSbW1P+MUjn8bx8PSPLdpsq+UE38kyeuwpaU6xwFgLB4mG8XYwpitJ0a6RQx3aGTiQM0kqHqjDPezKW9kNSw3+PYBtRSO5//6QVXMax64NP0PLAF/AHOtfEcxlv23dAr1ZcuAYL3wcw63/zWrlZIbRwXywnXQ7AIigITguIOeJUDQvvKUITbfpuow3n+6LlWDD2SlH5RKwxcwRgPC+WJFCkmvJoQMeRFKQDNGWJ2yJgGRlQoG/phnNmY2LPCJ8LgPb8k7+Fga3ZieSh0IrA9VYsC43764Lq9vK4mxpb+vX6jhqXdkz7cqMcYtunYPJOe/bJ4HdQXSiCkVgFn8OL/3DslzCJl/KdoTJFVlzI06jh8cIcikKCFwcKcnWshTC5JQ7+ZJ6Ggv1MVzCewd3MPyAS9LreBdVrXWvI4J/hoCWgqLKnoTbYMXICHZwd5FVJH8n1/kTyiRK3oZZlTSBvIszkm81FMkkR8x5xPPPMiFy+R04OTYVzOxoOIK3vktRx3I45zp+fhD+HvG0UuXE"
+
+script:
```
chrisprice: I think it will run this by default, it certainly is running this currently e.g. https://travis-ci.org/ScottLogic/d3fc/builds/86428877 (although travis seems to be down at the minute...)

bjedrzejewski: You mean, that 'grunt ci' should be removed? With it there, it doesn't run it twice, but it looks more explicit... Your call @chrisprice

Gruntfile.js
```diff
@@ -253,6 +273,9 @@ module.exports = function(grunt) {
 
     require('jit-grunt')(grunt);
 
+    grunt.loadNpmTasks('grunt-selenium-webdriver');
```
chrisprice: [jit-grunt](https://www.npmjs.com/package/jit-grunt) should load these automatically.

bjedrzejewski: Will test without and remove.

tests/wdio.conf.js
```diff
@@ -0,0 +1,156 @@
+exports.config = {
```
chrisprice: Can we trim this file down to just the bits we need (leave defaults as defaults etc.)? If we've got lots of comments, they're very likely to be ignored and unmaintained.

tests/webdriver/siteConsoleSpecWeb.js
```diff
@@ -0,0 +1,38 @@
...
+        }
+
+        return browser
+            .url('/components/introduction/1-getting-started.html')
+            .getAttribute('.nav-stacked a', 'href').then(function (attr) {
+                urls = attr; //iterate over it using the thing im in
+                return testNextPage();
+            });
+    });
+});
```
chrisprice: Files should have a newline at the end

.travis.yml
chrisprice: I personally don't like having setting default values to default values, so please can you remove it.

tests/wdio.conf.js
bjedrzejewski: I can remove all the comments. This is a standard file generated by the configuration generator. I thought the comments are useful and make it easier to understand.

chrisprice: I'd rather the file was 10 lines long :)

bjedrzejewski:
@chrisprice I think all your comments were addressed and commits cleaned up

chrisprice:
Cherry-picked into master as 84f06e423f0183d0e9ce0e9025a4370b910cae5f so that I could get the credentials right. Thanks again @bjedrzejewski!

PR #636, opened 2015-10-19
Performance improvements to low-barrel example

chrisprice:
Includes new functionality for suspending layouts.

### DIFF
--- a/site/src/examples/low-barrel/index.js
+++ b/site/src/examples/low-barrel/index.js
@@ -63,6 +63,51 @@
 }(d3, fc));
 
 (function (d3, fc) {
+    fc.candlestickSeries = function() {
+
+        var base = fc.series.ohlcBase();
+
+        var candlestick = fc.svg.candlestick()
+            .x(function(d) { return d.x; })
+            .open(function(d) { return d.yOpen; })
+            .high(function(d) { return d.yHigh; })
+            .low(function(d) { return d.yLow; })
+            .close(function(d) { return d.yClose; });
+
+        var fractionalBarWidth = fc.util.fractionalBarWidth(0.75);
+
+        var nest = d3.nest()
+            .key(function(d) { return d.direction; });
+
+        var dataJoin = fc.util.dataJoin()
+            .selector('path')
+            .element('path');
+
+        function candlestickSeries(selection) {
+            selection.each(function(data) {
+                data = data.filter(base.defined);
+
+                candlestick.width(base.width(data));
+
+                dataJoin(this, nest.entries(data.map(base.values)))
+                    .attr({
+                        'd': function(d) {
+                            return candlestick(d.values);
+                        },
+                        'class': function(d) {
+                            return 'candlestick ' + d.key;
+                        }
+                    });
+            });
+        }
+
+        d3.rebind(candlestickSeries, base, 'xScale', 'yScale', 'xValue', 'yValue');
+
+        return candlestickSeries;
+    };
+}(d3, fc));
+
+(function (d3, fc) {
     fc.mainChart = function() {
 
         var event = d3.dispatch('crosshair', 'zoom');
@@ -70,7 +115,7 @@
         var gridlines = fc.annotation.gridline()
             .yTicks(3);
 
-        var candlestick = fc.series.candlestick();
+        var candlestick = fc.candlestickSeries();
 
         var tooltip = fc.tooltip();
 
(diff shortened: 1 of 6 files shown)


site/src/examples/low-barrel/index.js
```diff
@@ -62,19 +62,10 @@
...
-(function(d3, fc) {
-    'use strict';
-
-    var dataGenerator = fc.data.random.financial()
-        .startDate(new Date(2014, 1, 1));
-
-    var container = d3.select('#low-barrel')
-        .datum(dataGenerator(250))
-        .layout();
+(function (d3, fc) {
```
ColinEberhardt: Do you need two wrap the components in functions? I am guessing this was left over from when they were in separate files?

site/src/examples/low-barrel/index.js
```diff
@@ -62,19 +62,10 @@
...
+    fc.mainChart = function() {
```
ColinEberhardt: Same again with adding components to the `fc` namepsace

site/src/examples/low-barrel/index.js
```diff
@@ -279,8 +339,20 @@
 
         container.select('svg.navigator')
             .call(navigatorChart);
+
+        container.layoutSuspended(true);
+    }
+
+    var rafId = null;
```
ColinEberhardt: Is it worth wrapping this ugliness up into a mini component?

ColinEberhardt:
Looks good to me - although it is a shame we need suspend layout :-(

I think this example illustrates a couple of important points:

 1. Using components to build more complex examples, I like the way that each of the charts is an independent component.
 2. Eventing - you'd probably expect each chart to handle the events exposed by the others, but instead the react / flux style architecture means that events always bubble up to the top-level code which re-renders everything.

It does feel like the tooltip should be rendered via the legend component. I had a quick go at that here:

https://github.com/ColinEberhardt/d3-financial-components/commit/a3a04611ddaaaa95612720b4302c0a6d27328bea

One issue I had was that the legend needs to be rendered via data-join, whereas in yours, as it decorate the crosshair, it's 'lifecycle' is already managed.

I do wonder whether we need an SVG legend as well?

chrisprice:
Thanks for the feedback, I've also just spotted that my last minute refactoring has broken it.... I'll convert this to WIP for now.

ColinEberhardt:
Forgot to mention, I think the point snapping is wrong - should snap on y-only

On Tue, Oct 20, 2015 at 9:07 a.m., ScottLogic/d3fc
<reply@reply.github.com>
wrote:
Thanks for the feedback, I've also just spotted that my last minute refactoring
has broken it.... I'll convert this to WIP for now.

—
[https://github.com/ScottLogic/d3fc/pull/636#issuecomment-149470090] .[https://github.com/notifications/beacon/ABDBfvYePhAhLDC02N4HS1POQDF85gzrks5o9e2tgaJpZM4GRjTo.gif]

chrisprice:
Agreed, will fix.

chrisprice:
* [ ] Investigate legend rendering
* [ ] Create rAF component
* [ ] Create new namespace
* [ ] Fix layout suspension

ColinEberhardt:
I should update the yahoo example to follow these patterns at some point as well
...

On Tue, Oct 20, 2015 at 9:10 a.m., ScottLogic/d3fc
<reply@reply.github.com>
wrote:
 * Investigate legend rendering
 * Create rAF component
 * Create new namespace
 * Fix layout suspension

—
[https://github.com/ScottLogic/d3fc/pull/636#issuecomment-149471033] .[https://github.com/notifications/beacon/ABDBfv-P24zhxjM1aruTk08Oynw2y6sQks5o9e5tgaJpZM4GRjTo.gif]

chrisprice:
I've done everything apart from the legend rendering. The current legend implementation which uses an HTML table doesn't easily fit into the structure, I think an SVG implementation would be a good idea but out of the scope of this PR.

chrisprice:
WIP removed ready for review.

ColinEberhardt:
Great, I'll have a look on the train home ...

On Tue, Oct 20, 2015 at 12:14 p.m., ScottLogic/d3fc
<reply@reply.github.com>
wrote:
I've done everything apart from the legend rendering. The current legend
implementation which uses an HTML table doesn't easily fit into the structure, I
think an SVG implementation would be a good idea but out of the scope of this
PR.

—
[https://github.com/ScottLogic/d3fc/pull/636#issuecomment-149530586] .[https://github.com/notifications/beacon/ABDBfhgeU7sRs4KGgmi7TAgUnPjXb4Yiks5o9hmmgaJpZM4GRjTo.gif]

chrisprice:
Thanks, one final thing I've (just) done is to wrap everything in a root component `example.lowBarrel`.

ColinEberhardt:
:+1:

PR #639, opened 2015-10-20
Added mapping description to multi series page on website

WPFerg:


chrisprice:
Thanks @WPFerg. As this is a documentation fix, I won't be releasing a new version.

PR #640, opened 2015-10-20
Multiple colours within multi series via nth index

jdunkerley:
Using css nth index with the multi class to change the stroke and fill styles for each item.


src/series/series.css
```diff
@@ -18,21 +18,47 @@
...
 }
 
-path.line {
-  fill: none;
+svg {
   stroke: #06c;
+  fill: #9cf;
 }
 
-path.area {
```
chrisprice: Just `text` maybe?

src/series/series.css
```diff
@@ -18,21 +18,47 @@
...
-  fill: #9cf;
+svg text {
+  stroke: none;
+  fill: #000;
 }
 
-.point>circle {
```
chrisprice: This is exactly what I was thinking.

src/series/series.css
```diff
@@ -18,21 +18,47 @@
   stroke: #6c0;
```
chrisprice: For all the ones above, could we drop the `>path`?

chrisprice:
Looking good! 

Are you going to add some `*-opacity` styles for `area`, `line`, `point`, etc. or are you going to do that with default decorators? Maybe a good first stab would be to do it all in CSS?

src/series/series.css
jdunkerley: yep thats a good change

jdunkerley:
on this PR, I had put inline styles for line chart, will change to CSS.

I still wonder as to whether a class on all series so we could apply more targeted (e.g. make .series.area, .series.bar) could then .series.up and .series.down for the candlestick and ohlc colour but guess that is a different issue.

chrisprice:
Sorry I missed that in the diff, I've been talked around to it being in the CSS :smile: 

I'm still to be convinced of `.series`, can you think of another usage of it?

As part of this, do you think it would make sense to centralise the CSS into one file?

jdunkerley:
We have about 5 CSS files at the moment. Think it could be nice to have them in one place as it would make it easier to come up with a normalised pattern.

jdunkerley:
Main thing I was looking to gain from the series class was not affecting the whole of svg. This could be done via the `cartesianChart` and `linearTimeSeries`, adjusting the CSS for their classes to set the default. That way it wouldn't affect all svg elements.

src/series/series.css
```diff
@@ -1,38 +1,72 @@
 .candlestick>path {
```
chrisprice: Do we need the `>path` here?

src/indicator/indicator.css
```diff
@@ -26,4 +26,10 @@
 
 .bollinger path.area {
```
chrisprice: Do we need the `.path`?

src/series/series.css
jdunkerley: nope just blind as a bat

src/series/series.css
```diff
@@ -1,38 +1,72 @@
...
+  fill: #f9f;
+}
+
+.area {
+  stroke-opacity: 0.0;
+}
+
+.line {
+  fill-opacity: 0.0;
+}
```
chrisprice: Newline required

src/indicator/indicator.css
```diff
@@ -1,29 +1,30 @@
-.divergence .bar>path {
+.macd-divergence .bar {
```
chrisprice: I'd be tempted to drop all of these rules?

chrisprice: (for .macd-divergence .bar)

chrisprice:
> We have about 5 CSS files at the moment. Think it could be nice to have them in one place as it would make it easier to come up with a normalised pattern and tidy up a little. For example, I think we have a chart class defined in fc.css not used anywhere.

:+1: Move it all into fc.css?

> Main thing I was looking to gain from the series class was not affecting the whole of svg. This could be done via the cartesianChart and linearTimeSeries, adjusting the CSS for their classes to set the default. That way it wouldn't affect all svg elements.

I don't think we need to worry about affecting the whole svg, the CSS should only be considered a starting point for anyone looking to use the library. In fact, maybe we should add some documentation to the top of the file stating as much?

jdunkerley:
Spark lines are not rendering quite right at the moment. 
image

jdunkerley:
Spark lines are not rendering correctly.

image

ColinEberhardt:
I'm very happy with this change :-)

On thing I didn't spot in the diff, have you removed all the logic in the visual tests where colours are 'manually' applied to stacked / grouped series?

Once merged, this will also require some website update too ;-)

src/fc.css
```diff
@@ -12,3 +12,214 @@
...
+.domain,
+.tick line,
+.tick path {
+  fill: none;
+  stroke-width: 1.0;
+  stroke: #bbb;
+}
+
+.tick text {
+  font: 10px sans-serif;
```
chrisprice: Kill it

jdunkerley: Removed, have likewise removed the one on annotations text

jdunkerley:
Yes @ColinEberhardt have removed all the decorate stuff for stacked and grouped.

All the visual tests are working, still have a little more work on examples to do.

jdunkerley:
I have gone over all the examples and documentation on the website and tweaked where needed (not very much).

A fix for the CrossHairs style so back to pale blue. 
All visualTests working (including another patch to crossHairs tests).

Think this is basically ready now.

site/src/examples/low-barrel/index.js
```diff
@@ -298,7 +302,7 @@
             });
 
         var multi = fc.series.multi()
-            .series([gridlines, line, area, brush])
```
chrisprice: Do we not need a line here?

jdunkerley: used a decorate to reveal the stroke on the area. Seemed more efficient

chrisprice: Oh cool, if it gives the same effect I've no problem with that

ColinEberhardt: One thing to bear in mind is that a stroking an area will cause it to have an outline on all of its edges, whereas a combination of line & area gives the impression that only the 'top' edge of the area has been stroked. Although looking at the visuals I can't see that it makes any difference here.

ColinEberhardt:
Great job - thanks.

Issue #643, opened 2015-10-20
layout is now fixed after initial run


chrisprice:
The behaviour of `layout` is to check for `layout-width` & `layout-height` attributes on the node, if they exist then use those values and if they don't then measure the node using `innerDimensions`. 

This works the first time `layout` is called because there are no attributes on the node. The final part of `layout` is to set the attributes on the current node AND the sub-tree. Therefore, the second time `layout` is called the old dimensions stored in the attributes are used.

This is caused by a logical flaw in the algorithm as it is now implemented (previously there was no check of the attributes). This was changed to prevent forced reflows being triggered (by innerDimensions) when `layout` was called on sub-trees which had already been laid out.

This is causing problems for a project which depends on us, we've basically broken their layout.

I propose fixing this by changing the algorithm to only set the `layout-width` and `layout-height` attributes on the sub-tree AND NOT on the current node. 

Does anyone have any thoughts?

ColinEberhardt:
I am guessing that the reflowing of layout is a performance issue rather be than a functional one? Out of interest, what is the expensive part of the layout algorithm, reading the dom state, computing layout or updating the dom?

If reading the state is quick, then can we cache the output of the layout algorithm?

chrisprice:
In summary, all are causing problems :smile: 

The algorithmic part of the layout we run takes a significant amount of time when you're aiming for 60fps, therefore we need some way of controlling it. This was fixed with the addition of `selection.layoutSuspended(<bool>)` (64e52e69268c7c8940ebd21ae5719344c6ef0f56).

The problem this issue references is the interlacing of layout-affecting DOM property reads and writes. The new approach avoids this by only reading from the DOM once at the top level.

All of this needs better documentation!

ColinEberhardt:
> All of this needs better documentation!

That would be good :-) Is it worth adding a brief technical discussion to the end of the layout component page on the website? 

I think documentation elsewhere would get lost,.

PR #644, opened 2015-10-20
Don't add layout-width/height attributes to root node

chrisprice:
See #643 for description of the bug. Also changes a couple of other things -

* Remove `fc.layout` component, I don't think it really needs to be exposed anymore.
* Force the return value of `.layoutSuspended()` to Boolean


chrisprice:
Missed unused reference to `fc.layout` in Yahoo example (bug), I've amended the commit to remove it.

PR #652, opened 2015-10-21
Add layout-x/y and enhance documentation

### LINKED ISSUE #648: Nested layout calls are resetting the x/y positioning of their root node
Similarly to #643, we should not set the x/y position on the root node. It will always be 0, 0 in the non-nested case and in the nested case it should remain where it is.

chrisprice:
Fixes #648

### DIFF
--- a/site/src/components/layout/layout.md
+++ b/site/src/components/layout/layout.md
@@ -42,9 +42,9 @@ example-code: |
     );
 ---
 
-The layout component allows you to structure your SVG using CSS Flexbox. This removes much of the manual layout calculations that are typically required for D3 charts, [as described in this blog post](http://blog.scottlogic.com/2015/02/02/svg-layout-flexbox.html).
+The layout functionality allows you to structure your SVG using CSS Flexbox. This removes much of the manual layout calculations that are typically required for D3 charts, [as described in this blog post](http://blog.scottlogic.com/2015/02/02/svg-layout-flexbox.html).
 
-In order to use this layout approach, add the `layout-style` attribute to the elements within your `SVG` container. This component uses Facebook's css-layout project, which re-implements flexbox in JavaScript, [refer to their project page](https://github.com/facebook/css-layout) for supported properties.
+In order to use this layout approach, add the `layout-style` attribute to the elements within your `SVG` container (note the camel-cased property names). This component uses Facebook's css-layout project, which re-implements flexbox in JavaScript, [refer to their project page](https://github.com/facebook/css-layout) for supported properties.
 
 Here's an example SVG with group elements that have a layout suitable for chart construction (axes, legend, etc ...):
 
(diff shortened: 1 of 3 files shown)


site/src/components/layout/layout.md
```diff
@@ -58,7 +58,7 @@ The following code performs the required layout:
 {{{example-code}}}
 ```
 
-The flexbox layout will be computed, resulting in translate transforms being applied to each element. Group elements do not have width or height properties, so this information is added via `layout-width` and `layout-height` properties. The layout component also provides a convenience accessor for these values, e.g. `selection.layout('width')`.
+Nodes are positioned and sized appropriately using a combination of `x`, `y`, `width`, `height` and `transform` attributes depending on the whether the node is an `svg`, `rect` or other. Additionally `layout-width`, `layout-height`, `layout-x` and `layout-y` attributes are added which can be accessed using e.g. `selection.layout('width')`.
```
ColinEberhardt: Currently it adds `x` and `y` properties without the `layout-` prefix, which looks like a bug.

site/src/components/layout/layout.md
```diff
@@ -81,3 +81,17 @@ svg.append('g')
       'flexDirection': 'row'
   });
 ```
+
+When building high performance charts e.g. when targeting 60fps, layout can be a source of performance problems. So in order to get the best performance, it's worth understanding some of its internals.
+
+At a high level, when a node is laid-out -
```
ColinEberhardt: You probably need to make it clear that you are referring to node layout within the context of our SVG / flexbox implementation.

src/layout/layout.js
```diff
@@ -80,6 +82,17 @@ function computeDimensions(node) {
     }
 }
 
+function computePosition(node) {
+    if (node.hasAttribute('layout-x') && node.hasAttribute('layout-y')) {
```
ColinEberhardt: Ah - ok, ignore the previous comment!

site/src/components/layout/layout.md
```diff
@@ -81,3 +81,17 @@ svg.append('g')
...
+
+* if it has not previously been laid-out it, its inner dimensions are measured. These are then combined with its `layout-style` and those of its descendants. After running these through the `css-layout` algorithm, the results are then applied node-by-node to the subtree.
+
+* if it has previously been laid-out, its inner dimensions are not be measured. Instead the `layout-width`, `layout-height`, `layout-x` and `layout-y` attributes set on it by the previous layout will be used as the basis for the algorithm. These are combined with the `layout-style` attributes of the node and its descendants then applied to the subtree as above.
```
ColinEberhardt: How does this effect re-sizing? if layout is invoked, then the browser resized, the inner dimensions could potentially change, however, according to this rule, the inner dimensions will not be re-measured.

ColinEberhardt:
Really good documentation, thanks @chrisprice - just one point that I think requires a little clarification, and one question on the behaviour.

ColinEberhardt:
LGTM ::+1::

chrisprice:
Just waiting for the build to be fixed then!

PR #657, opened 2015-10-22
Two ways to print env variables - delete soon

bjedrzejewski:
Delete that and the travis logs it will make.


bjedrzejewski:
@chrisprice this can be deleted from the history i think

PR #659, opened 2015-10-22
Only run webdriverTests if browserstack credentials available

### LINKED ISSUE #655: Grunt CI should not run webdriver when no key is set
I added that to mark that I will work on it.

chrisprice:
Fixes #655


bjedrzejewski:
Let's merge this one ASAP :+1:

PR #663, opened 2015-10-22
Fix for Quandl API in WebTests

jdunkerley:


ColinEberhardt:
Thanks, one quick thought, rather than matching the specific quandl API method, why not make it more generic and match for any quandl URL?

Also, could do with updating the comment!

jdunkerley:
Updated sorry was being lazy to get to build on my branch

ColinEberhardt:
Thanks, let's see if this works ;-)

PR #669, opened 2015-10-23
[WIP] Adjusting indicators and multi to remove colour controls from indicators

jdunkerley:
An idea from the work on the multi class allowing indicators (or anything else) to exclude themselves from the multi colours by setting the class name used by the multi series.

Normalised and unified all the indicator styles so common across the current set of indicators.

Goal was to make it so if we had two Bollinger bands then they would colour sequence. Also means don't need to fix their colour within the CSS file.

### DIFF
--- a/src/series/multi.js
+++ b/src/series/multi.js
@@ -9,7 +9,7 @@ import dataJoinUtil from '../util/dataJoin';
 // overriding where the series value is stored on the node (__series__) and
 // forcing the node datum (__data__) to be the user supplied data (via mapping).
 
-export default function() {
+export default function(className) {
 
     var xScale = d3.time.scale(),
         yScale = d3.scale.linear(),
@@ -18,10 +18,12 @@ export default function() {
         key = index,
         decorate = noop;
 
+    className = className || 'multi';
+
     var dataJoin = dataJoinUtil()
-        .selector('g.multi')
+        .selector('g.' + className)
         .children(true)
-        .attr('class', 'multi')
+        .attr('class', className)
         .element('g')
         .key(function(d, i) {
             // This function is invoked twice, the first pass is to pull the key
--- a/src/fc.css
+++ b/src/fc.css
@@ -171,35 +171,20 @@ text {
 }
 
 /* Indicators */
-.macd-divergence .bar {
-  stroke: grey;
+.indicator .bar {
   fill: grey;
-  fill-opacity: 1.0;
 }
 
-.macd-signal .line {
-  stroke: red;
+.signal .line {
+  stroke-width: 2px;
 }
 
-.macd .line {
-  stroke: #06c;
+.indicator .line {
+  stroke-dasharray: 2, 2;
 }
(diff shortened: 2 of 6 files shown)


src/series/multi.js
```diff
@@ -9,7 +9,7 @@ import dataJoinUtil from '../util/dataJoin';
 // overriding where the series value is stored on the node (__series__) and
 // forcing the node datum (__data__) to be the user supplied data (via mapping).
 
-export default function() {
+export default function(className) {
```
ColinEberhardt: Is there any reason for having classname as an immutable property of the multi component? If not, I'd just use the standard property pattern.

jdunkerley: I did it as immutable as if the dataJoin had created the objects would be hard not to leave legacy bits. First pass was a property

jdunkerley:
still would need to work through the site to normalise the effect of the change if people felt useful.

chrisprice:
Sorry to beat the decorate drum again but... is this not something best solved by decorate?

This diff seems to sum up my argument nicely https://github.com/ScottLogic/d3fc/pull/669/files#diff-051607ee3462a19057ed739a1a120251L41

jdunkerley:
All of the indicators had a decorate function attached to the multi already. But then with the multi series CSS changes, they need additional decorators (or very controlled styles in the CSS e.g. Bollinger Bands) to control the colours of each part of the render, as they will be a different colour by default. 

The changes to the decorators in each of the indicators is to remove the multi and use standard classes across all the indicators (e.g. signal and indicator class).

ColinEberhardt:
Sorry, only just had the chance to look at this PR again. In its present form it 'breaks' a number of the indicator renderers:

screen shot 2015-10-27 at 17 55 40

Although this could be due to changes on master since you created this PR.

However, putting this to one side for now, I don't think I agree with the underlying concept of making the indicator CSS more generic. I don't think that reflects reality, various indicators have generally accepted styles, e.g. MACD signal being rendered as a red line.

Although, one thing I do like in this PR is the change to remove the `multi` class when it is redundant, replacing it with `bollinger` for example. 

Furthermore, this could be 'pushed down' into datajoin, so that the pattern we see repeated throughout the code:

    var dataJoin = _dataJoin()
        .selector('g.annotation')
        .element('g')
        .attr('class', 'annotation');

simply becomes 

     var dataJoin = dataJoinUtil('annotation');

Also, it would remove the need for this code in small multiples:

https://github.com/ScottLogic/d3fc/blob/master/src/chart/smallMultiples.js#L33

So in summary:

 1. Sorry, don't fully agree with the motivations of this change :-(
 2. I like the improvements to multi series
 3. I'd happily receive a PR that improves on this further, pushing the concept into dataJoin removing the repeated boiler-plat code required to create a dataJoin that is 'scoped' by a CSS class.

jdunkerley:
Will take a look at pushing down to data join and do a new PR soon

PR #671, opened 2015-10-23
Axes come after data series in the DOM for cartesianChart

lukessmith:
The 'axes-container' SVG element is moved to come after 'plot-area', ensuring axes and their labels appear on top of the chart.

Without this the axes labels are easily obscured by area series.

I can't think of any cases where axes below plot data would be desirable, but please feel free to prove me wrong.

### DIFF
--- a/src/chart/cartesian.js
+++ b/src/chart/cartesian.js
@@ -93,13 +93,13 @@ export default function(xScale, yScale) {
                 <g class="plot-area-container"> \
                     <rect class="background" \
                         layout-style="position: absolute; top: 0; bottom: 0; left: 0; right: 0"/> \
+                    <svg class="plot-area" \
+                        layout-style="position: absolute; top: 0; bottom: 0; left: 0; right: 0"/> \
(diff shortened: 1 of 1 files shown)


chrisprice:
Unfortunately there's now a merge conflict because
`fc.chart.cartesianChart` has been renamed `fc.chart.cartesian`.

lukessmith:
I've rebased the changes.

ColinEberhardt:
I'm not so sure about this, what about when the axes are not on the border of the plot area, when they have a baseline set, I'd expect the series to be rendered over them not under

chrisprice:
I went looking for examples, here's what MatLab does - 
image
image

ColinEberhardt:
@lukessmith could you perhaps illustrates / describe the problem you are having? we might be able to come up with an alternative solution.

ColinEberhardt:
@lukessmith - I'm closing this for now, but as mentioned, would be happy to discuss alternatives

Issue #646, opened 2015-10-21
Expand FC.Util.Extent to a component


jdunkerley:
Thought it would be nice to have a little more flexibility with the Extent function.

If we created it as component could do something like:
```javascript
fc.util.extent()
    .includeZero()
    .scale(1.1)(data, ['high', low']);
```

chrisprice:
Like it!

ColinEberhardt:
I've wanted something similar ... on a few occasions I've found myself doing something like this:

```
fc.util.extent(series, function(d) { return 0; }, 'y')
```

(Notice you do not need to pass an array due to https://github.com/ScottLogic/d3fc/issues/468)

What if extent supported the following:

```
fc.util.extent(series, 0, 'y')
```

i.e. if a number is passed, convert to `function() { return number; }`

jdunkerley:
passing a number would certainly cover the 0 case, guess felt we got a little more flexibility if we had a component model.

Could then do thing like roundTo(...)

chrisprice:
I prefer the idea of a component over heavily overloaded method arguments. Maybe you could take a stab at the API for the component?

jdunkerley:
yep happy to

ColinEberhardt:
An API that allows for the addition of range padding would be useful.

jdunkerley:
Think @owennw is going to have a go at creating this as first component for him.

owennw:
I've implemented a function on extent that accepts a number and ensures that is it in the range, making the feature more flexible than just including 0. At the moment it's called "include".

What other features might we want on the extent? For example:

Scale(number): scales the max value up by the input, and the min value down by the same proportion? This wouldn't do anything to 0 though.

jdunkerley:
I think scale should magnify the range around the average point.

owennw:
How do we want `include` and `scale` to operate together?

For example:
Calculated range is `[10, 20]`, and `30` is included.
The scale is `0.8`.

Do we expect the final range to be `[11, 30]` or `[12, 28]` (or even anything else)?

chrisprice:
I'd be careful using the word scale, it's quite a loaded term in d3. Is
there a good use case for scaling with a value < 1? Does that therefore
suggest a different name?

ColinEberhardt:
I'd opt for something along the lines of padding or margin, basically a factor that increase the max and min by a given amount

owennw:
[...]

PR #674, opened 2015-10-23
Adjusted fc.util.extent to be component #646

owennw:
Also includes the `include` and `pad` methods. Note: `pad` will only scale number ranges, for the moment. If we want to implement scaling other types of range (e.g. `date`) then I see this as a separate issue.

Post review TODO list:
- [x] remove `d3.functior`
- [x] make use of `include`
- [x] try supporting `Date`, if it is easy to do

### DIFF
--- a/site/src/components/series/grouped-bar.md
+++ b/site/src/components/series/grouped-bar.md
@@ -26,7 +26,7 @@ example-code: |
       .rangePoints([0, width], 1);
 
   var y = d3.scale.linear()
-    .domain(fc.util.extent(series, function(d) { return 0; }, 'y'))
+    .domain(fc.util.extent()(series, function(d) { return 0; }, 'y'))
     .range([height, 0]);
 
   // create the grouped bar series
--- a/site/src/components/series/grouped-bar.md
+++ b/site/src/components/series/grouped-bar.md
@@ -26,7 +26,7 @@ example-code: |
       .rangePoints([0, width], 1);
 
   var y = d3.scale.linear()
-    .domain(fc.util.extent()(series, function(d) { return 0; }, 'y'))
+    .domain(fc.util.extent().include(0)(series, 'y'))
     .range([height, 0]);
 
   // create the grouped bar series
--- a/site/src/components/series/stacked.md
+++ b/site/src/components/series/stacked.md
@@ -30,7 +30,7 @@ example-code: |
       .rangePoints([0, width], 1);
 
   var y = d3.scale.linear()
-    .domain(fc.util.extent(series, function(d) { return 0; },
+    .domain(fc.util.extent()(series, function(d) { return 0; },
                             function(d) { return d.y + d.y0; }))
     .range([height, 0]);
 
--- a/site/src/components/series/stacked.md
+++ b/site/src/components/series/stacked.md
@@ -30,8 +30,7 @@ example-code: |
       .rangePoints([0, width], 1);
 
   var y = d3.scale.linear()
-    .domain(fc.util.extent()(series, function(d) { return 0; },
-                            function(d) { return d.y + d.y0; }))
+    .domain(fc.util.extent().include(0)(series, function(d) { return d.y + d.y0; }))
     .range([height, 0]);
 
   // create the stacked bar series (this could also be line or area)
--- a/src/util/extent.js
+++ b/src/util/extent.js
@@ -4,54 +4,62 @@ import d3 from 'd3';
  * The extent function enhances the functionality of the equivalent D3 extent function, allowing
  * you to pass an array of fields, or accessors, which will be used to derive the extent of the supplied array. For
  * example, if you have an array of items with properties of 'high' and 'low', you
- * can use <code>fc.util.extent(data, ['high', 'low'])</code> to compute the extent of your data.
+ * can use <code>fc.util.extent()(data, ['high', 'low'])</code> to compute the extent of your data.
  *
  * @memberof fc.util
- * @param {array} data an array of data points, or an array of arrays of data points
- * @param {array} fields the names of object properties that represent field values, or accessor functions.
  */
-export default function(data, fields) {
+export default function() {
 
-    // we need an array of arrays if we don't have one already
-    if (!Array.isArray(data[0])) {
-        data = [data];
-    }
-    if (arguments.length === 2) {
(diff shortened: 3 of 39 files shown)


src/util/extent.js
```diff
@@ -1,57 +1,105 @@
...
-        data = [data];
-    }
-    if (arguments.length === 2) {
-        // the fields parameter must be an array of field names, but we can pass non-array types in
-        if (!Array.isArray(fields)) {
-            fields = [fields];
+export default function() {
+
+    var include = noop,
+        pad = d3.functor(0);
```
ColinEberhardt: Neither `include` or `pad` need to be functions as you never pass any arguments to them when invoked.

We use functions elsewhere for things like property accessors, and use `d3.functor` to allow someone to supply a constant, e.g.

```
foo.xValue(function(d, i) { return 0; })
```

Can be simplified to:

```
foo.xValue(0)
```

thanks to `d3.functor`

src/util/extent.js
```diff
@@ -1,57 +1,105 @@
...
+        // Scale the range for the given padding
+        if (typeof dataMin === 'number' && typeof dataMax === 'number') {
+            var rangeDelta = pad() * (dataMax - dataMin) / 2;
+
+            min -= rangeDelta;
+            max += rangeDelta;
+        }
+
+        // Include the specified point in the range
+        var extraPoint = include();
```
ColinEberhardt: using the fact that `>` and `<` comparisons with `undefined` are always `false` feels a bit cryptic to me!

tests/util/extentsSpec.js
```diff
@@ -28,29 +28,102 @@ describe('fc.util.extent', function() {
...
+            .pad(1)(data, 'high');
+        expect(extents).toEqual([0, 25]);
+
+        extents = fc.util.extent()
+            .include(30)
+            .pad(1)(data, 'high');
+        expect(extents).toEqual([5, 30]);
+    });
+
+    it('should only pad numbers', function() {
```
ColinEberhardt: why not support dates too? as we have a lot of financial examples this would be useful. It would also be pretty trivial to implement.

site/src/components/series/grouped-bar.md
```diff
@@ -26,7 +26,7 @@ example-code: |
       .rangePoints([0, width], 1);
 
   var y = d3.scale.linear()
-    .domain(fc.util.extent(series, function(d) { return 0; }, 'y'))
+    .domain(fc.util.extent()(series, function(d) { return 0; }, 'y'))
```
ColinEberhardt: use `include(0)`

site/src/components/series/stacked.md
```diff
@@ -30,7 +30,7 @@ example-code: |
       .rangePoints([0, width], 1);
 
   var y = d3.scale.linear()
-    .domain(fc.util.extent(series, function(d) { return 0; },
+    .domain(fc.util.extent()(series, function(d) { return 0; },
```
ColinEberhardt: use 'include(0)'

visual-tests/series/grouped-bar.js
```diff
@@ -21,7 +21,7 @@
                 d3.scale.ordinal(),
                 d3.scale.linear())
             .xDomain(data.map(function(d) { return d.State; }))
-            .yDomain(fc.util.extent(series, function(d) { return 0; }, 'y'))
+            .yDomain(fc.util.extent()(series, function(d) { return 0; }, 'y'))
```
ColinEberhardt: use 'include(0)'

visual-tests/series/stacked-bar.js
```diff
@@ -39,7 +39,7 @@
                 d3.scale.ordinal(),
                 d3.scale.linear())
             .xDomain(data.map(function(d) { return d.State; }))
-            .yDomain(fc.util.extent(series, function(d) { return 0; }, function(d) { return d.y + d.y0; }))
+            .yDomain(fc.util.extent()(series, function(d) { return 0; }, function(d) { return d.y + d.y0; }))
```
ColinEberhardt: use 'include(0)'

ColinEberhardt:
- [ ] remove `d3.functior`
- [ ] make use of `include`
- [ ] try supporting `Date`, if it is easy to do

ColinEberhardt:
Nice job @owennw

chrisprice:
Sorry, I know I'm commenting late on this but I've only just spotted it when browsing the site (I'm really liking all the new documentation). Did you consider making fields part of the configuration of the component?

ColinEberhardt:
That crossed my mind too, but I couldn't decide whether the fields were effectively part of the input data to the component, rather than component state.

Other than juggling the bracket locations, can you see any other benefits to making the fields a property of the component?

chrisprice:
Static versus dynamic configuration (most of the time)
Removal of unnamed parameter
Unlikely to want an include which affects both dimensions e.g. date and
price

ColinEberhardt:
> Removal of an unnamed parameter

That's a very good point, moving from this:

    fc.util.extent()(data, ['high', 'low'])

to this:

    fc.util.extent().fields('high', 'low')(data)

certainly improves the API. I'd also simplify the API so that the `fields` property only supports a variable argument list.

owennw:
>Unlikely to want an include which affects both dimensions e.g. date and price

I'm unsure what this implies. An extent is for a single dimension of the data -- the extent for the other dimension could use an `include` too. How does changing the API affect that?

ColinEberhardt:
>>Unlikely to want an include which affects both dimensions e.g. date and price
>I'm unsure what this implies. An extent is for a single dimension of the data -- the extent for the other dimension could use an include too. How does changing the API affect that?

I think what Chris means is that currently you can currently construct an extent as follows:

    var extentWithInclude = fc.util.extent().include(20)

Then you can apply it to multiple dimensions

    var xDomain = extentWithInclude(data, 'x');
    var yDomain = extentWithInclude(data, 'y');

However, how realistic is this as an example? How often are you going to want to have an extent with the same value included for multiple dimensions?

I don't think there is a single example of this in our current codebase.

chrisprice:
Might be worth considering not making it varargs. I think that's the d3
standard (e.g. scale range/domain) to permit returning the same thing from
the "getter"

owennw:
So something like `fc.util.extent.fields(['high', 'low'])(data)`?

PR #682, opened 2015-10-26
Add documentation page for extent

owennw:


site/src/components/util/extent.md
```diff
@@ -0,0 +1,29 @@
+---
+layout: component
+title: Extent
+component: util.extent.js
+namespace: util
+
+example-code: |
```
ColinEberhardt: elsewhere we use `example-code` front-matter variables for a couple of reasons:

 1. To ensure that the code being executed is identical to the code displayed to the user
 2. So that we can re-use code, e.g. showing an example on the front page.

Neither apply in this case, so feel free to add the code snippets directly in to the body of the post.

I'd also recommend breaking this block of examples up and interspersing it with the text.

site/src/components/util/extent.md
```diff
@@ -0,0 +1,29 @@
...
+  fc.util.extent()
+      .pad(0.5)(data, fields); // [2.5, 47.5]
+  fc.util.extent()
+      .pad(0.5)
+      .include(0)(data, fields); // [0, 47.5]
+---
+
+ The extent function enhances the functionality of the equivalent D3 extent function, allowing you to pass an array of fields, or accessors, which will be used to derive the extent of the supplied array. For example, if you have an array of items with properties of 'high' and 'low', you can use `fc.util.extent()(data, ['high', 'low'])` to compute the extent of your data.
+
+ The range can be modified by using the `pad` property. Pad will scale the calculated range by the given input, e.g. if the calculated range is `[10, 20]` then calling `pad(1)` will modify the range to be `[5, 25]`. `pad(0)` is an identity.
```
ColinEberhardt: "by the given input" - "by the given amount"?

ColinEberhardt: if the calculated range is `[10, 20]` then the result of padding by '1' will be ...

site/src/components/util/extent.md
```diff
@@ -0,0 +1,29 @@
...
+
+ The range can be extended to include a fixed value, using the `include` property. The range is extended after padding, so the order of setting the properties is not important. If the included value was already within the range there will be no change.
+
+```js
+{{{example-code}}}
+```
```
ColinEberhardt: You can also use extent on an array of arrays, would be good to include an example of that too.

ColinEberhardt:
Thanks @owennw - a few suggestions, but otherwise looking good.

bjedrzejewski:
@owennw I think you mixed up your eslint stuff into this PR

ColinEberhardt:
Thanks :+1:

PR #683, opened 2015-10-26
Added browserstack to visualTests. Closes #668.

### LINKED ISSUE #668: Add browserstack testing for visualTests
We need to also test visualTests with browserstack. This needs some work with grunt, as visualTests can run at the same time as the website.

bjedrzejewski:
This is testing the visualTests now and running visualTests on the port 9000.

The testPagesForConsoleOutput function is duplicated, but I couldn't get the browserstack to correctly use it as a beforeEach or before in wdio.conf.js.

If anyone can fix it, please add a commit here, or fix in a later commit.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -48,6 +48,11 @@ module.exports = function(grunt) {
                     name: 'localhost',
                     port: 8000,
                     sslFlag: 0
+                },
+                {
+                    name: 'localhost',
+                    port: 9000,
+                    sslFlag: 0
                 }],
                 forcelocal: true,
                 onlyAutomate: true,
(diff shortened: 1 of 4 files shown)


owennw:
The README.md should also be updated to reflect the port number change.

bjedrzejewski:
@owennw thanks for spotting that. It's fixed now,

bjedrzejewski:
@ColinEberhardt @chrisprice this is ready for merge I think. It will also run visualTests though the browserstack and it switches visualTests to port 9000.

ColinEberhardt:
Thanks @bjedrzejewski :-)

PR #691, opened 2015-10-27
Integrated Playground

jdunkerley:
Got a lot of work to do with Assemble but have merged in a prototype playground.

Live online at http://jdunkerley.github.io/d3fc/


ColinEberhardt:
This is shaping up quite nicely. It would be good to discuss further how you see this integrating into the site - will it replace, or extend examples?

One observation on your code, are you aware that the following pattern:

    component.property = function(x) {
        if (!arguments.length) {
            return property;
        }
        property = x;
        return component;
    };

is quite specific to the D3 component pattern? it only really makes sense if you want an object construction pattern where you configure a factory which then creates one or more instances of a component based on that configuration.

jdunkerley:
Have tidied up the playground.js not to have the properties
Playground URL now at http://jdunkerley.github.io/d3fc/playground/playground.html

Still have index page to build for it
http://jdunkerley.github.io/d3fc/playground/playground.html?example=components/series/bar

Only currently works for components using the example-fixture

jdunkerley:
Far too much work with Assemble but have integrated links working now

site/src/_layoutsPlayground/component.hbs
```diff
@@ -0,0 +1,45 @@
...
+<head>
+    <meta charset="UTF-8">
+    <title>{{default title 'd3fc'}}</title>
+    <link href="https://cdnjs.cloudflare.com/ajax/libs/d3fc/{{ package.version }}/d3fc.min.css" rel="stylesheet" />
+</head>
+
+<body>
+    <svg id="chart" width="100%" height="300px"></svg>
+
+    <!-- <script src="https://cdnjs.cloudflare.com/ajax/libs/d3/{{ slice package.dependencies.d3 1 }}/d3.js"></script> -->
```
ColinEberhardt: Any reason for this commented out code?

site/src/_layoutsPlayground/component.hbs
```diff
@@ -0,0 +1,45 @@
...
+    <!-- <script src="https://cdnjs.cloudflare.com/ajax/libs/css-layout/{{ slice package.dependencies.css-layout 1 }}/css-layout.js"></script> -->
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/d3fc/{{ package.version }}/d3fc.bundle.min.js"></script>
+    <script data-src="{{basename}}.js">
+var width = document.getElementById('chart').getBoundingClientRect().width || 600,
```
ColinEberhardt: Do we need the `|| 600` code here?

ColinEberhardt: Actually, why not use fc.util.innerDimensions()?

https://github.com/ScottLogic/d3fc/blob/master/src/util/innerDimensions.js

site/src/_layoutsPlayground/default.hbs
```diff
@@ -0,0 +1,19 @@
...
+<html lang="en">
+
+<head>
+    <meta charset="UTF-8">
+    <title>{{default title 'd3fc'}}</title>
+    <link href="https://cdnjs.cloudflare.com/ajax/libs/d3fc/{{ package.version }}/d3fc.min.css" rel="stylesheet" />
+</head>
+
+<body>{{> body}}
+    <!-- <script src="https://cdnjs.cloudflare.com/ajax/libs/d3/{{ slice package.dependencies.d3 1 }}/d3.js"></script> -->
```
ColinEberhardt: commented out code?

site/src/playground/examples/horizon.md
```diff
@@ -0,0 +1,71 @@
+---
```
ColinEberhardt: Can we omit the horizon example for the time-being? I really like the code / concept and want to see it in the core of d3fc. I'd like to keep our website, and where we place examples, as simple as possible.

site/src/playground/index.md
```diff
@@ -0,0 +1,33 @@
+---
+layout: default
+title: Playground Gallery
+---
+<div class="previews container">
```
ColinEberhardt: Can you add a title and a few lines of text to this page to describe what it is?

site/src/_layoutsPlayground/component.hbs
jdunkerley: The playground uses this to switch to local d3fc.js files when ?local is added
Otherwise gets trimmed out at load

site/src/playground/examples/horizon.md
jdunkerley: sure will delete it and change the default to a bar chart

site/src/_layoutsPlayground/component.hbs
jdunkerley: didn't know that function and was just changing the createFixture to something clearer. Will switch

site/src/_layoutsPlayground/default.hbs
jdunkerley: Same as above.

site/src/playground/index.md
jdunkerley: yep will do

chrisprice:
Given the additional functionality this gives us I think we should merge this now. There are still some minor issues that we've discussed (e.g. removing the additional assemble pass and misc. styling tweaks) but I think we should just polish those up subsequently. Thanks @jdunkerley (Chief ASSemble Wrangler)!

PR #692, opened 2015-10-28
Envelopes indicator

### LINKED ISSUE #678: Add envelopes options to moving averages and EMA
Envelopes are popular with moving averages and exponential moving averages. Example can be found here:

http://stockcharts.com/school/doku.php?id=chart_school:technical_indicators:moving_average_envelopes

I think it should be easy to add support for drawing them on the chart.

bjedrzejewski:
I added envelopes indicator with tests, visual tests and documentation.
In the documentation you can find usage of envelopes with ema and candlesticks.

closes #678

### DIFF
--- a/site/src/components/indicator/envelope.md
+++ b/site/src/components/indicator/envelope.md
@@ -1,0 +1,108 @@
+---
+layout: component
+title: Envelope
+component: indicator/algorithm/envelope.js
+tags:
+  - frontpage
+namespace: indicator
+
+example-code: |
+  // Create the line series
+  var line = fc.series.line()
+      .xScale(xScale)
+      .yScale(yScale);
+
+  container.append('g')
+      .datum(data)
+      .call(line);
+
+  // Create and apply the envelopes algorithm
+  var envelopeAlgorithm = fc.indicator.algorithm.envelope()
+      .channel(0.01)
+      .midValue(function (d){return d.close});
+  envelopeAlgorithm(data);
+
+  // Create the renderer
+  var envelope = fc.indicator.renderer.envelope()
+      .xScale(xScale)
+      .yScale(yScale);
+
+  // Add it to the container
+  container.append('g')
+      .datum(data)
+      .call(envelope);
+      
+example-code-2: |
+  // Create the candlestick series
+  var candlestick  = fc.series.candlestick ()
+      .xScale(xScale)
+      .yScale(yScale);
+
+  container.append('g')
+      .datum(data)
+      .call(candlestick );
+      
+  // Create and apply the EMA
+    var movingAverage = fc.indicator.algorithm.exponentialMovingAverage()
+    movingAverage(data);
+  
+    // Create a line that renders the result
+    var ema = fc.series.line()
+        .yValue(function(d) { return d.exponentialMovingAverage; })
+        .xScale(xScale)
+        .yScale(yScale);
+  
+    // Add it to the container
+    container.append('g')
+        .datum(data)
+        .call(ema);
+
+  // Create and apply the envelopes algorithm to the exponential moving average
+  var envelopeAlgorithm = fc.indicator.algorithm.envelope()
+        .channel(0.01)
+        .midValue(function (d){return d.exponentialMovingAverage});
+  envelopeAlgorithm(data);
+
+  // Create the renderer
+  var envelope = fc.indicator.renderer.envelope()
+      .xScale(xScale)
+      .yScale(yScale);
+
+  // Add it to the container
+  container.append('g')
+      .datum(data)
+      .call(envelope);
+---
+
+Envelope is an indicator that defines upper and lower `channel` for a given time series. It can be used on any
+value series by specifying the `midValue` function.  The `channel` size can also be easily specified.
+
+The example below creates a line series and an envelope indicator:
+
+```js
+{{{example-code}}}
+```
+
+{{>example-fixture}}
+
+Envelopes are most often used with [Exponential Moving Averages](./exponentialMovingAverage) as breaking out
+of the price from such a channel defines either buy or sell signal. The example below creates a 
+[candlestick chart](../series/candlestick) with [exponential moving average](./exponentialMovingAverage) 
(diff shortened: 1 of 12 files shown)


ColinEberhardt:
As a general note, I just wanted to point out that we have a set of design guidelines for indicators:

https://github.com/ScottLogic/d3fc/blob/master/src/indicator/design-guidelines.md

It looks like you have followed almost all of these, but might have missed one or two.

src/indicator/algorithm/calculator/envelope.js
```diff
@@ -0,0 +1,36 @@
+import d3 from 'd3';
+
+export default function() {
+
+    var channel = 0.1;
+    var midValue = function(d) {
+        return d.value;
+    };
+    var envelope = function(data) {
+        data.forEach(function(s) {
```
ColinEberhardt: as per the design guidelines, calculators should output a new array of data rather than mutating the input data.

src/indicator/algorithm/calculator/envelope.js
```diff
@@ -0,0 +1,36 @@
...
+            if (midValue(s) !== undefined) {
```
ColinEberhardt: In order that calculators do not have to deal with leading 'undefined' values we have an adapter that takes care of that, resulting in simpler calculator implementations. See the following:

https://github.com/ScottLogic/d3fc/blob/915b2ff2d79096c8aa12cbac634038f62debcf4a/src/indicator/algorithm/calculator/macd.js#L18

src/indicator/algorithm/calculator/envelope.js
```diff
@@ -0,0 +1,36 @@
...
+                s.lowerEnvelope = midValue(s) * (1.0 - channel);
+                s.upperEnvelope = midValue(s) * (1.0 + channel);
+            }
+        });
+        return data;
+    };
+
+    envelope.channel = function(x) {
```
ColinEberhardt: is 'channel' the conventional name for this? I'd have thought that 'factor' or 'multiplier' might make more sense, although I haven't done any research on envelope indicators!

src/indicator/algorithm/calculator/envelope.js
```diff
@@ -0,0 +1,36 @@
+import d3 from 'd3';
+
+export default function() {
+
+    var channel = 0.1;
+    var midValue = function(d) {
```
ColinEberhardt: the value accessor should default to the identity

ColinEberhardt: I'd also simply call it 'value'

src/indicator/algorithm/calculator/envelope.js
```diff
@@ -0,0 +1,36 @@
...
+        }
+        channel = x;
+        return envelope;
+    };
+
+    envelope.midValue = function(x) {
+        if (!arguments.length) {
+            return midValue;
+        }
+        midValue = d3.functor(x);
```
ColinEberhardt: I wouldn't functor this property, I don't think it makes sense. Why would you ever supply a constant? In the past we've been a bit too liberal with our functors!

src/indicator/renderer/envelope.js
```diff
@@ -0,0 +1,99 @@
...
+    var envelope = function(selection) {
+
+        var multi = _multi()
+            .xScale(xScale)
+            .yScale(yScale)
+            .series([area, upperLine, lowerLine])
+            .decorate(function(g, data, index) {
+                g.enter()
+                    .attr('class', function(d, i) {
+                        return 'multi envelope ' + ['area', 'upperEnvelope', 'lowerEnvelope'][i];
```
ColinEberhardt: I'd simply use 'upper' and 'lower', the element is already classed 'envelope'

site/src/components/indicator/envelope.md
```diff
@@ -0,0 +1,108 @@
...
+      .call(envelope);
+      
+example-code-2: |
+  // Create the candlestick series
+  var candlestick  = fc.series.candlestick ()
+      .xScale(xScale)
+      .yScale(yScale);
+        
+  // Create and apply the EMA
+    var movingAverage = fc.indicator.algorithm.exponentialMovingAverage()
```
ColinEberhardt: inconsistent indentation

bjedrzejewski:
@ColinEberhardt I think I have addressed all the issues mentioned

src/indicator/algorithm/calculator/envelope.js
```diff
@@ -0,0 +1,40 @@
...
+    var envelopeFactor = function(valueData) {
+        return valueData.map(function(s) {
+            return {
+                lowerEnvelope: s * (1.0 - factor),
+                upperEnvelope: s * (1.0 + factor)
+            };
+        });
+    };
+    var envelope = function(data) {
+        return undefinedInputAdapter().undefinedValue({}).algorithm(envelopeFactor)(
```
ColinEberhardt: the undefined adapter should be applied within the algorithm, rather than the calculator. As mentioned, calculators do not have to deal with undefined values.

ColinEberhardt:
@ScottLogic/d3fc the envelope currently uses `channel` as the property for scaling the envelope size. I'm not so keen on this. How about `scale` or `factor`. What do people think?

ColinEberhardt:
> the envelope currently uses channel as the property 

ignore me, I was looking at the wrong commit!

src/indicator/algorithm/calculator/envelope.js
bjedrzejewski: I think I was mislead by the MACD example as it uses the undefinedInputAdapter in the calculator.

bjedrzejewski: @ColinEberhardt I am not sure if this is very clean, as the algorithm will have to use value function from the calculator to decide if something is undefined or not. Is there a way around it, or is it ok?

jdunkerley:
Couple of changes to doc. Rest LGTM.

ColinEberhardt:
replaced by #706

PR #693, opened 2015-10-28
Added missing functors to svg elements. Closes #631

### LINKED ISSUE #631: SVG path generators should functor-ise x/y values
To [match d3](https://github.com/mbostock/d3/blob/master/src/svg/area.js#L27).

bjedrzejewski:
@chrisprice I think this is what you referred to in #631 . I have looked in the svg directory and added functors where there was expectation of a function, but no functor applied.

If this is broader issue that you had in mind, let me know and I can go through the rest of it.

### DIFF
--- a/src/svg/candlestick.js
+++ b/src/svg/candlestick.js
@@ -56,35 +56,35 @@ export default function() {
         if (!arguments.length) {
             return x;
         }
-        x = _x;
+        x = d3.functor(_x);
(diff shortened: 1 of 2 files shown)


chrisprice:
Yes that's exactly what I had in mind. I can't remember when I stumbled across the problem but I think we should functor anything which could conceivably have a data item passed to it.

PR #694, opened 2015-10-28
Move fields to be a property on extent

owennw:

PR #696, opened 2015-10-28
Move fields to be a property on extent 

ColinEberhardt:
replaces #694


jdunkerley:
Nice improvement. LGTM

ColinEberhardt:
Thanks @jdunkerley

PR #699, opened 2015-10-29
Documentation how to make 'step series' closes #698

### LINKED ISSUE #698: Add step series chart
In the d3fc vision there is a mention of step series. It looks like that:

http://www.highcharts.com/stock/demo/step-line

It is getting more popular and I have seen it in some of our designs. I think it should be very similar to line series, just rendered differently, so maybe a different renderer is what is needed.

bjedrzejewski:
Thanks to @jleft it turns out that we can already do a step series pretty easy, the documentation just wasnt there. I updated the documentation then.

### DIFF
--- a/site/src/components/series/line.md
+++ b/site/src/components/series/line.md
@@ -13,6 +13,27 @@ example-code: |
   container.append('g')
       .datum(data)
       .call(line);
+      
+example-code-2: |
+  var line = fc.series.line()
+      .interpolate('basis')
+      .tension(0.3)
+      .xScale(xScale)
+      .yScale(yScale);
+
+  container.append('g')
+      .datum(data)
+      .call(line);
+      
+example-code-3: |
+  var line = fc.series.line()
+      .interpolate('step')
+      .xScale(xScale)
+      .yScale(yScale);
+
+  container.append('g')
+      .datum(data)
+      .call(line);
 ---
 
 The line series renders the given data as a line, constructed from an SVG `path`:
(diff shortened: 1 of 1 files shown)


ColinEberhardt:
Thanks for this - I think it is very helpful for our users to inform them of some of the D3 features that we support (via rebind etc ...) but they might not be aware of.

However, we shouldn't reproduce large chunks of D3 documentation, what happens if this changes in future? How do we maintain our copy?

I'd instead link to the d3 interpolate documentation page where people can read about the interpolation modes.

Other than that, looks good to me :-)

site/src/components/series/line.md
```diff
@@ -27,3 +48,53 @@ Which gives the following:
 
 This series has the same properties for specifying the x and y value accessors, and decorate as the [point series](#point).
 
+`Line` also exposes the `interpolate` property from the d3 library, which accepts 11 different interpolations:
```
jleft: I think there are actually 13 types of interpolation - `linear-closed` and `step` are missing from this list.

https://github.com/mbostock/d3/blob/v3.5.6/src/svg/line.js#L83-L98

bjedrzejewski:
Will link the documentation and just mention couple of them for the examples.

bjedrzejewski:
@jleft  @ColinEberhardt updated and rebased

jleft:
:+1: Looks good :)

jdunkerley:
Looks good to me. Another one for the multi example playground...

ColinEberhardt:
Thanks @bjedrzejewski - just spotted something, the `basis` interpolation mode does not support tension. Rather than provide an exhaustive set of examples I've simplified just to highlight the step interpolation mode.

PR #706, opened 2015-10-29
Envelope Indicator

ColinEberhardt:
Replaces #692 

I've made the following changes:

+ Making use of fc.util.identity
+ changed from d.envelope.upperEnvelope to d.envelope.upper etc ...
+ envelope algorithm now adapts the envelope computer to accomodate undefined values
+ changed undefinedInputAdapter to use the algorithm's value accessor
+ slightly more complex visual test
+ a few documentation tweaks

### DIFF
--- a/site/src/components/indicator/envelope.md
+++ b/site/src/components/indicator/envelope.md
@@ -1,0 +1,108 @@
+---
+layout: component
+title: Envelope
+component: indicator/algorithm/envelope.js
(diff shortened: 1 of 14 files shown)


bjedrzejewski:
Thanks Colin for all the help with this one. I am surprised that only such a small tweak for `undefinedInputAdapter` was required.

PR #708, opened 2015-10-29
Add class name control to Multi Series and Data Join 

jdunkerley:
Added ability to set class name on multi series
Added ability to set class name and element name on data join util
Normalised import on _multi and _dataJoin to multiSeries and dataJoinUtil

### DIFF
--- a/src/indicator/renderer/relativeStrengthIndex.js
+++ b/src/indicator/renderer/relativeStrengthIndex.js
@@ -1,6 +1,6 @@
 import d3 from 'd3';
 import annotationLine from '../../annotation/line';
-import _multi from '../../series/multi';
+import multiSeries from '../../series/multi';
 import {noop} from '../../util/fn';
 import seriesLine from '../../series/line';
 
@@ -10,7 +10,7 @@ export default function() {
         yScale = d3.scale.linear(),
         upperValue = 70,
         lowerValue = 30,
-        multiSeries = _multi(),
+        multi = multiSeries('rsi'),
         decorate = noop;
 
     var annotations = annotationLine();
@@ -19,9 +19,17 @@ export default function() {
 
     var rsi = function(selection) {
 
-        multiSeries.xScale(xScale)
+        multi
+            .xScale(xScale)
             .yScale(yScale)
-            .series([rsiLine, annotations])
+            .series([annotations, rsiLine])
+            .decorate(function(g, data, index) {
+                g.enter()
+                    .attr('class', function(d, i) {
+                        return 'rsi ' + ['', 'signal'][i];
+                    });
+                decorate(g, data, index);
+            })
             .mapping(function(series) {
                 if (series === annotations) {
                     return [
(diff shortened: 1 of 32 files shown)


src/indicator/renderer/relativeStrengthIndex.js
```diff
@@ -19,9 +19,17 @@ export default function() {
 
     var rsi = function(selection) {
 
-        multiSeries.xScale(xScale)
+        multi
+            .xScale(xScale)
             .yScale(yScale)
-            .series([rsiLine, annotations])
+            .series([annotations, rsiLine])
+            .decorate(function(g, data, index) {
```
ColinEberhardt: any reason why you've changed the order from `series().mapping().decorate()` to `series().decorate().mapping()`?

ColinEberhardt:
One quick question, other than that, looks really good :-)

src/indicator/renderer/relativeStrengthIndex.js
jdunkerley: consistently with all the other indicators (all the others with annotations are annotations then indicators)

ColinEberhardt: I don't think my question was clear (and probably on the wrong line!) - your commit changes the order in which the methods on the D3 selection are called. Was this required in order to fix some issue?

jdunkerley: No was just bringing into same sequence as the stochastic just for consistency.

Force indicator added after this one is the same was as RSI was so happy to revert, was only going to make them all look as similar as possible

ColinEberhardt: No problem, just wanted to make sure it wasn't changed in order to address some underlying issue.

ColinEberhardt:
Just discussed with @chrisprice and we both like the simplification that this provides by supplying the classname. However by adding element name it results in a more procedural-style syntax and lacks clarity:

```
fc.util.dataJoin('up', 'path')
```

On reading the above, what do the two string arguments mean?

Here's a suggestion:

a. Change dataJoin so that it does not select on element type, so select on `.foo` rather than `g.foo`. This would allow you to change element name on the component directly rather than supplying as an argument to the factory function.
b. Rather than changing dataJoin directly, effectively overloading its functionality, create a new component, a classedDataJoin (needs a good name!) that adapts dataJoin, selectively rebinding its properties. In this case just rebind `element`, `key` and `children`.

Expanding on point (2) ... from an API perspective it would look something like this:

```
var dataJoin = fc.util.classedDataJoin().classed('annotation');
```

This has a few advantages:

1. We no longer need to have variables passed to the constructor function.
2. By suppressing the `attr` and `selector` properties you can make `classed` a proper mutable property.

This might mean that (a) above is not needed.

Going forwards it might make sense to prototype this and show how the API is working out before creating a fully rolled-out pull request.

jdunkerley:
closing PR for now will do a new version based on last comments and come back

Issue #672, opened 2015-10-23
SVG Legend Component


bjedrzejewski:
Current legend component is a simple table in a div. Proper SVG legend component would be preferred.

PR #709, opened 2015-10-30
Added a couple of improved examples

ColinEberhardt:
Here are a couple of examples that illustrate the sort of level we should be pitching at, simple charts composed form a small number of components to illustrate common use-cases

screen shot 2015-10-30 at 08 05 16
screen shot 2015-10-30 at 08 05 47

Once we have a few of these I'll tidy up the examples landing page.

Also, would love an SVG legend :-) #672


ColinEberhardt:
@owennw one thing I did discover, on the bubble chart I'd have liked to pad the x-axis extent however it is a log scale which makes things a bit complicated, for two reasons, you have to pad each end logarithmically (of course!), but also you cannot pad to create a minimum <=0

chrisprice:
Conversely where do you stop with supporting specialised series?

owennw:
I don't think it would be much trouble to support padding logarithmically, but as I mentioned yesterday about the scope of d3fc (toolkit vs charting library) I don't know where that would belong..

chrisprice:
These look good to me, we probably should discuss how these type of examples combine with the playground though. Maybe we just drop the playground link and make all the examples work with the playground? One for the meeting on Monday.

PR #713, opened 2015-10-30
Refactor bar and improve documentation

owennw:
Small tweak to the bar class ahead of implementing variable widths, and update to the documentation to showcase the existing functionality.

### DIFF
--- a/site/src/components/series/bar.md
+++ b/site/src/components/series/bar.md
@@ -13,6 +13,16 @@ example-code: |
   container.append('g')
       .datum(data)
       .call(bar);
+
+example-code-2: |
+  var bar = fc.series.bar()
+      .xScale(xScale)
+      .yScale(yScale)
+      .orient('horizontal');
+
+  container.append('g')
+      .datum(data)
+      .call(bar);
 ---
 
 The bar series renders the given data as a series of vertical bars:
@@ -27,3 +37,23 @@ Which gives the following:
 
 This series has the same `xValue`, `yValue` and `decorate` properties as the [point series](./point). You can also specify the width of the bars via the `barWidth` property.
 
+The bar series supports horizonal display, using the `orient` property.
+
+```js
+{{{example-code-2}}}
+```
+
+<div id="series_bar_horizontal" class="chart"> </div>
+<script type="text/javascript">
+(function() {
+    var desiredWidth = $('#series_bar_horizontal').width(),
+        desiredHeight = desiredWidth / 2.4; //keeps the width-height ratio at 600-250 (defaults for createFixture)
+    var f = createFixture('#series_bar_horizontal', desiredWidth, desiredHeight, null, function() { return true; });
+    var container = f.container, data = f.data,
+        yScale = f.xScale,
+        xScale = d3.scale.linear()
+            .domain(fc.util.extent().fields(['high', 'low'])(data))
+            .range([0, desiredHeight]);
+    {{{example-code-2}}}
+}());
+</script>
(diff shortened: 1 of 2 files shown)


site/src/components/series/bar.md
```diff
@@ -5,7 +5,7 @@ component: series/bar.js
 tags:
 namespace: series
 
-example-code: |
+example-code-1: |
```
jdunkerley: no need to change from example-code and example-fixture

jdunkerley:
looks good, but would switch back `example-code-1` to `example-code` and keep the `example-fixture` as no need to change and forthcoming playground integration based on this at moment.

jdunkerley:
:+1:

ColinEberhardt:
I like the extra documentation, it's a good idea that we demonstrate the capabilities of this library more clearly. 

However, I'm not so keen on the refactor to the bar series component. I don't think it makes the code any clearer or more efficient, it's just different!

Also, it is a little inconsistent in that it adds an `isVertical` function but not a corresponding `isHorizontal`.

I'd be happy to merge the documentation update, but keep the code as-is. Sorry!

owennw:
Adding an `isHorizontal` is trivial, but the main cause for the refactor is to remove duplication, and only specify a piece of logic once. If we wanted to change the string that specified the orientation we now only have to do that in one place, not several. (The same is currently true for horizontal, which is why I left it there).

I think in general not repeating yourself and expressing a concept once is both clearer, and from a developer's point-of-view, more efficient.

ColinEberhardt:
Unfortunately code quality is subjective and what constitutes clearer code
depends on opinion. While I do agree with the DRY principles in their most
general sense, my feeling is that they lose value when applied to such small
units of code.
Furthermore, the features of a particular language have an influence on the
conventions that developers adopt. You'll find that JavaScript developers are
much more comfortable with repeated string literals than a C# or Java developer
might be.
For guidance take a look at the d3 source code.

On Fri, Oct 30, 2015 at 1:45 p.m., ScottLogic/d3fc
<reply@reply.github.com>
wrote:
Adding an isHorizontal is trivial, but the main cause for the refactor is to remove duplication, and
only specify a piece of logic once. If we wanted to change the string that
specified the orientation we now only have to do that in one place, not several.
(The same is currently true for horizontal, which is why I left it there).

I think in general not repeating yourself and expressing a concept once is both
clearer, and from a developer's point-of-view, more efficient.

—
[https://github.com/ScottLogic/d3fc/pull/713#issuecomment-152529881] .[https://github.com/notifications/beacon/ABDBfmTX_sWQkaD6eWN2DmXTlyz4m9P3ks5pA2vhgaJpZM4GY5d8.gif]

owennw:
I agree that aspects of code quality can be subjective, but I haven't seen a reason for why this particular refactor doesn't have, or loses, value.

>If we wanted to change the string that specified the orientation we now only have to do that in one place, not several.

While JS devs might be more comfortable with repeated logic, I don't see why that is sufficient reason to ignore the benefit pointed out above.

jdunkerley:
Could we do something taking this a bit further:

``` JavaScript
orientations = {
    vertical {
        containerTranslation: function(d, i) { return 'translate(' + base.x1(d, i) + ', ' + base.y0(d, i) + ')'; }
        barHeight: function(d, i) { return base.y1(d, i) - base.y0(d, i); },
        valueAxisDimension ...
    }, 
    horizontal {
        ...
    }
}

var bar = function(selection) {
    selection.each(function(data, index) {

        if (orient !== 'vertical' && orient !== 'horizontal') {
            throw new Error('The bar series does not support an orientation of ' + orient);
        }
```

PR #714, opened 2015-10-30
Site and visualTests can work together now. Fixes #711

### LINKED ISSUE #711: Support running the site and visual tests at the same time through grunt
Despite the two pages running on different ports they both use the 'watch' component which I believe uses the same port, meaning the two cannot be run in parallel.

bjedrzejewski:
@owennw specially for you


jdunkerley:
LGTM :)

chrisprice:
Nice!

PR #717, opened 2015-11-01
Fix point series to update symbol path data, rather than just set it on enter

### LINKED ISSUE #716: Point series does not update symbol size or type
I believe this is a recent regression (https://github.com/ScottLogic/d3fc/commit/70fb0e264da82a4d37936c47760f363542d7bffe#diff-5ecbd1f1e05aef955dc413ecd921ef1a). The path data of the  symbol on the point series is only set on _enter_ and not on _update_.

In the following example, the symbol would not change from circle to cross, which I'd expect. This affects both the `size` and `type` of the symbol.

```js
var point = fc.series.point()
    .type('circle');

container.datum(data)
    .call(point);
    
point.type('cross');
    
container.call(point);
```

jleft:
Fixes #716

### DIFF
--- a/src/series/point.js
+++ b/src/series/point.js
@@ -28,10 +28,11 @@ export default function() {
             var g = dataJoin(this, filteredData);
             g.enter()
                 .attr('transform', containerTransform)
-                .append('path')
-                .attr('d', symbol);
+                .append('path');
 
-            g.attr('transform', containerTransform);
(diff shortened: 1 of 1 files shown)


jdunkerley:
This looks good to me

ColinEberhardt:
Good job, thanks @jleft

PR #722, opened 2015-11-02
Restore line chart and remove stroke

### LINKED ISSUE #720: low barrel example navigator chart is showing area outline
It's subtle on this screen but I've noticed it worse on other monitors -

image

I think it was [this change](https://github.com/ScottLogic/d3fc/commit/c492751085a4edd926682d6a281439008f627aa8#diff-19b5e9ef58b8f9e9f1bd0aec6d0f9847) which caused the regression.

jdunkerley:
Fixes #720

### DIFF
--- a/site/src/examples/low-barrel/index.js
+++ b/site/src/examples/low-barrel/index.js
@@ -284,11 +284,13 @@
             .yTicks(0);
 
         var area = fc.series.area()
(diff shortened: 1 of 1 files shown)


ColinEberhardt:
Thanks @jdunkerley

PR #724, opened 2015-11-03
svg legend

### LINKED ISSUE #672: SVG Legend Component

ColinEberhardt:
Fixes #672 

This integrates the really cool SVG legend from this project:

http://d3-legend.susielu.com

It is exposed via the cartesian component via the `legend`property, an additional property `legendLayout` sets the flexbox layout of the container allowing you to position the legend.

I like the way this significantly simplifies the scatter and bubble series examples.

It does mean that we now have our own HTML legend and an SVG legend. We actually have two different concepts here:

 1.  A legend (i.e. this PR) that is 'bound' to a scale
 2.  A tooltip (our current HTML legend) which is bound (via datajoin) to the data

In the long term I'd like to replace our current HTML legend with an SVG equivalent and rename it to tooltip. It should integrate easily with our crosshair.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -109,6 +109,7 @@ module.exports = function(grunt) {
                 src: [
                     'node_modules/d3/d3.js',
                     'node_modules/css-layout/dist/css-layout.js',
+                    'node_modules/d3-svg-legend/d3-legend.js',
                     'node_modules/svg-innerhtml/svg-innerhtml.js',
                     'dist/d3fc.js'
                 ],
@@ -155,6 +156,7 @@ module.exports = function(grunt) {
                     {
                         src: [
                             'node_modules/css-layout/dist/css-layout.js',
+                            'node_modules/d3-svg-legend/d3-legend.js',
                             'dist/d3fc.js',
                             'dist/d3fc.css',
                             'node_modules/jquery/dist/jquery.js',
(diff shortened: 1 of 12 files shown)


jdunkerley:
Looks great to me @ColinEberhardt. Really like the new component.

One little thing, I see you fixing the grouped bar visual test. It is not on the index.html list for the visualTests at the moment,

ColinEberhardt:
well spotted - I forgot that they are no longer automatically added to the index page. I'll add it.

chrisprice:
I really like legend functionality provided by the library but I'm not sure
about it's inclusion in cartesian chart.

As opposed to the axis labels which have non trivial positioning logic, the
legend container seems too simple. There's also nothing which makes it
specifically a legend container. The only bespoke logic seems to be to
toggle the legend container if the legend property is toggled.

I suppose it feels to me like we're making a monolith rather than having a
workable pattern of composable components.

ColinEberhardt:
@chrisprice the legend decorates-in quite easily:

https://github.com/ScottLogic/d3fc/pull/724/files#diff-1410861ad9f50ac17b1225532e3765faR47

I agree, this is a more flexible way to add a legend. @jleft have you considered this approach for adding the year label in your 'Health & Wealth' example?

chrisprice:
Cool, glad to hear it. Is it worth a comment to explain why you need to
layout after adding the element and why you're doing that on the update
selection?

ColinEberhardt:
> Is it worth a comment to explain why you need to
layout after adding the element and why you're doing that on the update
selection?

It probably is, there are two things of note, firstly layout is being performed from the parent SVG, second, I added it to the update selection because it felt like the correct thing to do, if the chart were re-sized it would need to be laid-out again.

However, in order to avoid making this example over-complicated, I think I'll move it to the enter selection for now. I'll re-visit when we start making more of an effort to support resizing throughout the website.

jleft:
> @jleft have you considered this approach for adding the year label in your 'Health & Wealth' example?

I haven't but I'll look into it - thanks for the idea @ColinEberhardt :)

Issue #690, opened 2015-10-27
Add Elder Ray Indicator


bjedrzejewski:
Elder Ray is an indicator that features in Trading for a Living by Alexander Elder (very popular book) and is defined as two separate indicators:

Bull Power = Daily High - n-period EMA
Bear Power = Daily Low - n-period EMA 

With default 13 days period (customisable) and has an interesting visualization as two bar series:

http://www.dailyfx.com/forex_forum/attachments/coding-strategy-advisors-indicators-functions/65921d1283304973-elder-ray-bear-bull-power-20100831-m12-194kb.jpg

ColinEberhardt:
This would be a useful addition - thanks.

PR #727, opened 2015-11-03
Add Elder Ray Indicator - #690

owennw:


src/indicator/renderer/elderRay.js
```diff
@@ -0,0 +1,86 @@
...
+
+        bullBarTop
+            .xValue(xValue)
+            .yValue(function(d, i) {
+                return root(d).bullTop ? root(d).bullPower : undefined; });
+
+        bearBarTop
+            .xValue(xValue)
+            .yValue(function(d, i) {
+                return root(d).bearTop ? root(d).bearPower : undefined; });
```
bjedrzejewski: A comment why 4 series are used would be good.

bjedrzejewski:
Looks good to me :+1:

site/src/components/indicator/elderRay.md
```diff
@@ -0,0 +1,49 @@
...
+namespace: indicator
+
+example-code: |
+  // Create and apply the elder ray algorithm
+  var elderRayAlgorithm = fc.indicator.algorithm.elderRay().period(6);
+  elderRayAlgorithm(data);
+
+  // the elder ray is rendered on its own scale
+  var yScale = d3.scale.linear()
+      .domain([-5, 5])
```
ColinEberhardt: is a domain of -5 to -5 always going to be appropriate? Why not use fc.util.extent?

Actually, there's something else `fc.util.extent` should support, symmetrical domains! In this case you probably want to render a chart with a domain centred on '0', but where the maximum is computed from the data.

I'm happy to leave this example at [-5, 5] for now, but would love to see fc.util.extent extended further.

src/indicator/algorithm/calculator/elderRay.js
```diff
@@ -0,0 +1,73 @@
...
+    var ema = exponentialMovingAverage()
+        .windowSize(13);
+
+    var elderRay = function(data) {
+
+        ema.value(value);
+        var x = ema(data);
+
+        var elderRay = d3.zip(data, x)
+            .map(function(d) {
```
ColinEberhardt: There's a bit of a problem with this calculator, it doesn't accomodate the fact that the ema will have 'leading' undefined values. Look at where undefinedInputAdapter is used, in MACD and Envelope, this should hopefully allow you to ignore the undefined values in your calculator.

ColinEberhardt:
Thanks for this @owennw - looking really good.

Other than my comments in the code there are a couple of other things:

 1. I'm not sure about the renderer and what the `bullTop` property is doing? From googling it seems that quite a few people render the two components as separate series (which would not need a dedicated renderer, barSeries would be just fine), or they render both with one series above (z-index) the other (e.g. https://sites.google.com/site/jayeshbheda/ADX-Elders-Ray-RSI-on-NIFTY.jpg). I'd have thought a renderer that simply combines two barSeries without any further logic would be sufficient.
2. Can you add a visual test too? I know it adds a bit of duplication, however these are quite useful. You can make them even more useful by making them configurable (see cartesian as an example). Ideally they let us test a range of values.

If you want to discuss any of the above - give me a shout.

src/indicator/algorithm/calculator/elderRay.js
owennw: Yes I had a think about the leading undefined values and opted to leave them in, not knowing if there was a preference. I'll remove them though.

site/src/components/indicator/elderRay.md
owennw: Not always appropriate, but appropriate here. I didn't use extent as the range data is inside the 'elderRay' object on the data, and I didn't want to pollute the example. Do you think that should be included?

As for extent I'll raise an issue.

src/indicator/algorithm/calculator/elderRay.js
ColinEberhardt: Having leading undefined values makes it it easier to implement series / renderers. If you look at point / bar / area etc ,, they all have common functionality that filters out undefined values.

ColinEberhardt: It also means that for indicator renderers, where you do `root(d)` you can always assume that this will return an object rather than undefined. I think all of this is captured here:

https://github.com/ScottLogic/d3fc/blob/master/src/indicator/design-guidelines.md

ColinEberhardt:
This is looking really good - thanks for the updates. There is still one remaining issue, if you look at your visual tests you can see that there are errors due to incorrect path data:

screen shot 2015-11-04 at 12 40 23

You can see that in the line above I have logged the data that is being used by your renderer. Notice that the `elderRay` properties are NaN. Our series use `undefined` or `null` to indicate missing data:

https://github.com/ScottLogic/d3fc/blob/master/src/series/xyBase.js#L27

As a result these `NaN` values end up in the SVG path data.

I know I've been encouraging you to use undefinedInputAdapter. From playing around with the code it looks like it isn't suitable for your needs here. I've instead fixed the problem directly in your calculator:

https://github.com/ColinEberhardt/d3-financial-components/commit/f3df2c889d36d60f0ebde272a4309facc592a19d

Can you please cherry pick the commit above onto your branch so that it is added to this PR? I'd be happy to merge after that.

Many Thanks

PR #729, opened 2015-11-04
Error bar series added to the site and visualTests. Closes #700

### LINKED ISSUE #700: Add error bar series
As described in the vision for d3fc, error bar series would be useful.  We can draw then in couple ways, so feedback is welcome. Here are examples:

https://en.wikipedia.org/wiki/Error_bar
http://scienceblogs.com/cognitivedaily/2008/07/31/most-researchers-dont-understa-1/

Or should we go as far as to make these:

https://en.wikipedia.org/wiki/Box_plot

bjedrzejewski:
This is a the first version. Please check the site and the visualTests example.

### DIFF
--- a/site/src/components/series/errorBar.md
+++ b/site/src/components/series/errorBar.md
@@ -1,0 +1,46 @@
+---
+layout: component
+title: Error Bar Series
+component: series/errorBar.js
+tags:
+  - frontpage
+namespace: series
+
+example-code: |
+  var point = fc.series.point()
+      .xScale(xScale)
+      .yScale(yScale);
+  
+  container.append('g')
+      .datum(data)
+      .call(point);
+  
+  //Generating the error information for the data
+  data.forEach(function(d) {
+      d.yUpError = Math.random() * 2;
+      d.yDownError = Math.random() * 2;
+  });
+
+  var eBar = fc.series.errorBar()
+      .xScale(xScale)
+      .yScale(yScale)
+      .yLow(function(d,i) {return d.close - d.yDownError;})
+      .yHigh(function(d,i) {return d.close + d.yUpError;})
+      .yValue(function(d,i) {return d.close;});
+
+  container.append('g')
+      .datum(data)
+      .call(eBar);
+---
+
+An [error bar series](http://en.wikipedia.org/wiki/Error_bar) renders the error or uncertainty area around given
+values. These can be in x or y dimension. To get the bars rendered, `Low` or `High` values should be supplied using
+`xLow`, `xHigh`, `yLow`, `yHigh` properties of the series. 
+
+```js
+{{{example-code}}}
+```
+
+Which gives the following:
+
+{{>example-fixture}}
--- a/site/src/components/series/errorBar.md
+++ b/site/src/components/series/errorBar.md
@@ -3,7 +3,7 @@ layout: component
 title: Error Bar Series
 component: series/errorBar.js
 tags:
-  - frontpage
+  - playground
 namespace: series
 
 example-code: |
@@ -24,18 +24,17 @@ example-code: |
   var eBar = fc.series.errorBar()
       .xScale(xScale)
       .yScale(yScale)
-      .yLow(function(d,i) {return d.close - d.yDownError;})
-      .yHigh(function(d,i) {return d.close + d.yUpError;})
-      .yValue(function(d,i) {return d.close;});
-
+      .errorLow(function(d,i) {return d.close - d.yDownError;})
+      .errorHigh(function(d,i) {return d.close + d.yUpError;});
+      
   container.append('g')
       .datum(data)
       .call(eBar);
 ---
 
 An [error bar series](http://en.wikipedia.org/wiki/Error_bar) renders the error or uncertainty area around given
-values. These can be in x or y dimension. To get the bars rendered, `Low` or `High` values should be supplied using
-`xLow`, `xHigh`, `yLow`, `yHigh` properties of the series. 
+values. These can be in x or y dimension depending on the `orient` property. To get the bars rendered, `lowError` and 
+`highError` properties need to be set correctly.
 
 ```js
 {{{example-code}}}
--- a/src/series/errorBase.js
+++ b/src/series/errorBase.js
@@ -1,0 +1,99 @@
+import d3 from 'd3';
+import fractionalBarWidth from '../util/fractionalBarWidth';
+
+export default function() {
+
+    var xScale = d3.time.scale(),
+        yScale = d3.scale.linear(),
+        yHigh = d3.functor(0),
+        xHigh = function(d, i) { return d.date; },
+        yLow = d3.functor(0),
+        xLow = function(d, i) { return d.date; },
+        xValue = function(d, i) { return d.date; },
+        yValue = function(d, i) { return d.y; },
+        barWidth = fractionalBarWidth(0.75),
+        xValueScaled = function(d, i) {
+            return xScale(xValue(d, i));
+        };
+
+    function base() { }
+
+    base.width = function(data) {
+        return barWidth(data.map(xValueScaled));
+    };
+
+    base.values = function(d, i) {
+        return {
+            x: xValueScaled(d, i),
+            y: yScale(yValue(d, i)),
+            yLow: yScale(yLow(d, i)),
+            yHigh: yScale(yHigh(d, i)),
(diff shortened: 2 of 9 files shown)


bjedrzejewski:
I need to fix a browserstack issue first here.

bjedrzejewski:
@ColinEberhardt this should work, passes my travis. Can we restart the travis build to see if this is one of these instabilities?

site/src/components/series/errorBar.md
```diff
@@ -0,0 +1,46 @@
+---
+layout: component
+title: Error Bar Series
+component: series/errorBar.js
+tags:
+  - frontpage
```
owennw: Does this want to be on the frontpage?

site/src/components/series/errorBar.md
```diff
@@ -0,0 +1,46 @@
...
+      d.yUpError = Math.random() * 2;
+      d.yDownError = Math.random() * 2;
+  });
+
+  var eBar = fc.series.errorBar()
+      .xScale(xScale)
+      .yScale(yScale)
+      .yLow(function(d,i) {return d.close - d.yDownError;})
+      .yHigh(function(d,i) {return d.close + d.yUpError;})
+      .yValue(function(d,i) {return d.close;});
```
owennw: I think the yLow etc. lines wouldn't pass eslint due to the whitespace with the `{`, but these aren't linted because they're in the md. In a more general sense is it possible to lint these files? And should we try to make these sections eslint compliant?

owennw: Maybe it would be worth using the multiseries here?

src/series/errorBase.js
```diff
@@ -0,0 +1,99 @@
...
+            xLow: xScale(xLow(d, i)),
+            xHigh: xScale(xHigh(d, i))
+        };
+    };
+    base.defined = function(d, i) {
+        return yLow(d, i) != null && yHigh(d, i) != null &&
+            xValue(d, i) != null && yValue(d, i) != null
+            && xLow(d, i) != null && xHigh(d, i) != null;
+    };
+
```
owennw: I think we should be consistent with the line breaks between properties -- unless there's a logical reason to split them here?

src/svg/errorBar.js
```diff
@@ -0,0 +1,97 @@
...
+    var errorBar = function(data) {
+
+        return data.map(function(d, i) {
+            var halfWidth = width(d, i) / 2,
+                yTotal = yHigh(d, i) - yLow(d, i),
+                yBottom = y(d, i) - yLow(d, i),
+                yUpper = yHigh(d, i) - y(d, i),
+                xTotal = xHigh(d, i) - xLow(d, i),
+                xBottom = x(d, i) - xLow(d, i),
+                xUpper = xHigh(d, i) - x(d, i);
```
owennw: Small thing, I would expect to either have "yBottom/yTop" or "yLower/yHigher".

visual-tests/series/errorBar.js
```diff
@@ -0,0 +1,50 @@
...
+        d.yUpError = Math.random();
+        d.yDownError = Math.random();
+    });
+
+    var errorBar = fc.series.errorBar()
+        .xScale(dateScale)
+        .yScale(priceScale)
+        .xHigh(function(d, i) {
+            var newD = new Date(d.date.getTime() + (0.5 * 24 * 60 * 60 * 1000)); //add half a day
+            return newD;
```
owennw: You can just return the date straight away.

src/series/errorBase.js
```diff
@@ -0,0 +1,99 @@
...
+        xHigh = function(d, i) { return d.date; },
+        yLow = d3.functor(0),
+        xLow = function(d, i) { return d.date; },
+        xValue = function(d, i) { return d.date; },
+        yValue = function(d, i) { return d.y; },
+        barWidth = fractionalBarWidth(0.75),
+        xValueScaled = function(d, i) {
+            return xScale(xValue(d, i));
+        };
+
```
owennw: Might be worth including a yValueScaled too, or commenting on why there isn't one.

owennw:
Looks good to me! Would be good to see that browserstack error go, and besides that only small changes from me.

site/src/components/series/errorBar.md
ColinEberhardt: 'no' to front page, and 'yes' to playground!

site/src/components/series/errorBar.md
ColinEberhardt: Yeah, we should be linting this stuff, https://github.com/ScottLogic/d3fc/issues/730

src/series/errorBase.js
ColinEberhardt: I'm pretty relaxed about line breaks, as long as it is readable and doesn't fail a linting rule, I'm happy!

visual-tests/series/errorBar.js
ColinEberhardt: You can also remove the milliseconds calculations by using d3's time intervals

https://github.com/mbostock/d3/wiki/Time-Intervals

ColinEberhardt:
Nice job, it's good to see you have followed the patterns, creating an svg renderer, rendering each datapoint within a 'g' to allow decoration etc ... 

The documentation looks good too.

I have spotted a few issues:

 1. What happens if someone wants to render vertical error bars with a dataset that has points with x and y properties? They would set the xValue accessor of errorBar series to xValue(function(d) { return d.x; }), however xHigh and xLow would still be trying to access a date property which does not exist?
 2. barWidth is not exposed
 3. I don't think you really need errorBase, we only have 'base' components when there is functionality that can be shared across series components. You could probably make use of xyBase (selectively rebinding the relevant properties).

One way to tackle (1) and make your series implementation simpler is to just have an error bar series that renders errors in a single direction. You could then expose an 'orient' property, which reflects the similar property on bar series (and svg renderer) I think this would allow you to remove the check that causes the problem and simplify your code.

bjedrzejewski:
@ColinEberhardt 

1. In this case they will have to set xHigh and xLow as well. Other properties can be zeroed simply.
2. It is exposed as .width from the svg
3. I thought that I may not need it, I was considering renaming it to 'box base' for rendering 2d objects on a cartesian chart.

I am not sure that orient property makes it easier to use. I could do it like that, but then you always need 2 series to render actual boxes and it may not be that simple?

ColinEberhardt:
> In this case they will have to set xHigh and xLow as well. Other properties can be zeroed simply.

Having to set xHigh and xLow when you are only interested in error bars on the y dimensions is a bit of a pain!

> It is exposed as .width from the svg

True, however, I don't think this is rebound onto the series, so there is no way of setting it.

> you always need 2 series to render actual boxes and it may not be that simple?

If you Google Image search 'error bars' it looks like the majority of people only render error bars across a single dimension. 

I really think an orient approach will work better, as it has been 'proven' to a certain extent with the bar series. One other issue I have spotted (that would be eliminated via orient) is that your current bar width calculation is only on the x-values. This means it cannot easily be used to render the following:

bjedrzejewski:
Ok, I will refactor it to be in a single dimension then :)

site/src/components/series/errorBar.md
bjedrzejewski: Ok, changed

src/series/errorBase.js
bjedrzejewski: Cleaned up for aesthetics

src/svg/errorBar.js
bjedrzejewski: Fixed, will refactor to a single dimmension anyway

visual-tests/series/errorBar.js
bjedrzejewski: I used the var for ease of debugging, changed it now. Didn't know about the d3, great tip!

src/series/errorBase.js
bjedrzejewski: I am following the pattern seen in ohlc here, I will refactor to single dimension and then it may change anyway. Thanks for suggestion.

bjedrzejewski:
@ColinEberhardt @owennw have a look at the refactored version

ColinEberhardt:
Thanks, I'll take a look in a bit - on a train at the moment

ColinEberhardt:
Thanks for the updates - looks much better. Also as an added bonus you can now have error bars with different widths in the x and y directions :-)

I've made a few minor tweaks and will merge via a different PR.

PR #732, opened 2015-11-04
Subsampling modemedian

WPFerg:
Creates a new component of `d3fc.data`, `samplers`, which subsamples data in order to make rendering/processing large data sets easier.

The first subsampler is the *mode median* subsampler, which looks at the mode and the median of a range of data sets.

### DIFF
--- a/site/src/components/samplers/modeMedian.md
+++ b/site/src/components/samplers/modeMedian.md
@@ -5,14 +5,19 @@ component: data/samplers/modeMedian.js
 namespace: samplers
 
 example-code: |
-    var data = fc.data.random.financial()(10000);
+    var data = fc.data.random.financial()(20);
 
-    var subsampledData = fc.data.samplers.modeMedian()
-                                         .number(5)
-                                         .field('low')(data);
+    var sampler = fc.data.samplers.modeMedian()
+                                  .number(5)
+                                  .value(function(d) { return d.low; });
+
+    var sampledData = sampler(data);
 
     d3.select('#subsampled-data')
-      .text(JSON.stringify(subsampledData, null, 2));
+      .text(JSON.stringify(sampledData, null, 2));
+
+    d3.select('#full-data')
+      .text(JSON.stringify(data, null, 2));
 
 ---
 
@@ -24,9 +29,14 @@ The example below creates large array of data points, reducing it to 5.
 {{{example-code}}}
 ```
 
-Which gives the following:
+Which gives the following subsampled data:
 
 <pre id="subsampled-data">Loading...</pre>
+
+...when given the following data:
+
+<pre id="full-data">Loading...</pre>
+
 <script type="text/javascript">
     (function() {
         {{{example-code}}}
--- a/site/src/components/samplers/modeMedian.md
+++ b/site/src/components/samplers/modeMedian.md
@@ -1,0 +1,35 @@
+---
+layout: component
+title: Mode Median Sampling
+component: data/samplers/modeMedian.js
+namespace: samplers
+
+example-code: |
+    var data = fc.data.random.financial()(10000);
+
+    var subsampledData = fc.data.samplers.modeMedian()
+                                         .number(5)
+                                         .field('low')(data);
+
+    d3.select('#subsampled-data')
+      .text(JSON.stringify(subsampledData, null, 2));
+
+---
+
+The mode-median sampling component is a fast method of subsampling data to improve performance with large data sets.
+
+The example below creates large array of data points, reducing it to 5.
+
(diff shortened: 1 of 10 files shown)


src/data/samplers/modeMedian.js
```diff
@@ -0,0 +1,107 @@
+import d3 from 'd3';
+import {identity, noop} from '../../util/fn';
+
+export default function() {
+
+    var undefinedValue = d3.functor(undefined),
+        number = d3.functor(10),
```
ColinEberhardt: This isn't quite right, I'll pop over and show you how functors work in a bit ...

src/data/samplers/modeMedian.js
```diff
@@ -0,0 +1,107 @@
...
+        accumulator = noop,
+        field = identity;
```
ColinEberhardt: Can you call this 'value' so that it matches the property name found in our other calculator components?

src/data/samplers/modeMedian.js
```diff
@@ -0,0 +1,107 @@
...
+        number = d3.functor(x);
+        return modeMedian;
+    };
+
+    modeMedian.field = function(x) {
+        if (!arguments.length) {
+            return field;
+        }
+
+        if (typeof x === 'string') {
```
ColinEberhardt: I'd remove this logic, and only accept a function. The only place we employ this 'magic' is `fc.util.extent`, as a rule accessors should be functions, this is so that when ES6 comes around they will change from

```
value(function(d) { return d.value; })
```

to

```
value (d => d.value)
```

src/data/samplers/modeMedian.js
```diff
@@ -0,0 +1,107 @@
...
+            } else {
+                return bucket[Math.floor(bucket.length / 2)];
+            }
+        });
+
+        // First and last data points are their own buckets.
+        return [].concat(data[0], subsampledData, data[data.length - 1]);
+    };
+
+    var getGlobalMinMax = function(data) {
```
ColinEberhardt: Can you change these from anonymous functions into function declarations? see this file for the pattern:

https://github.com/ScottLogic/d3fc/blob/master/src/series/bar.js

And this page for more info:

http://stackoverflow.com/questions/336859/var-functionname-function-vs-function-functionname

src/data/samplers/modeMedian.js
```diff
@@ -0,0 +1,107 @@
...
+        accumulator = noop,
+        field = identity;
+
+    var modeMedian = function(data) {
+
+        if (number() > data.length) {
+            return data;
+        }
+
+        var minMax = getGlobalMinMax(data);
```
ColinEberhardt: d3 extent actually provides this for you: https://github.com/mbostock/d3/wiki/Arrays#d3_extent

site/src/components/samplers/modeMedian.md
```diff
@@ -0,0 +1,35 @@
+---
+layout: component
+title: Mode Median Sampling
+component: data/samplers/modeMedian.js
+namespace: samplers
+
+example-code: |
+    var data = fc.data.random.financial()(10000);
+
+    var subsampledData = fc.data.samplers.modeMedian()
```
ColinEberhardt: I'd re-write this as:

```
var sampler = fc.data.samplers.modeMedian()
                                         .number(5)
                                         .field('low');

var sampledData = sampler(data);
```

This makes the component construction and use of the component more distinct

site/src/components/samplers/modeMedian.md
```diff
@@ -0,0 +1,35 @@
...
+                                         .number(5)
+                                         .field('low')(data);
+
+    d3.select('#subsampled-data')
```
ColinEberhardt: Can you also show the original data on this page? I think it will provide a much better illustration of the functionality. You might need to reduce the points from 10,000 to something more manageable!

WPFerg:
Pushed up a change that alters `number` to `bucketSize`

ColinEberhardt:
> Pushed up a change that alters number to bucketSize

Great - this looks much better. Last thing, are you going to update the website page to give a more visual illustration of this components functionality

WPFerg:
The website was updated in a previous commit -- it may have gotten lost when I sent up the second: 
https://github.com/WPFerg/d3fc/commit/b2cd8119b87b59ca04f8aecf520d9545aacddd39#diff-5988621afa4ff3712be6233b2cf757ba

ColinEberhardt:
Great job, I made a few more tweaks here #746 - nothing major, just spotted the plural namespace. Easier for me to quickly fix it.

Issue #728, opened 2015-11-03
Allow extent to support symmetrical domains


owennw:
fc.util.extent could support symmetrical domains, centered around a point with the range either set or computed somehow from the data.

owennw:
First thoughts:

- Add a new function to extent, perhaps called `about(input)` that accepts an input, and uses that value as the centre point about which to create the range.
- The calculated range would be relative to this point, and then extended symmetrically in both directions using the max value calculated.

So if, for example, the data were shaped in such a way to have produced a range of `[6,  13]` then the product of modifying the extent would be:
`.about(9)` -> `[5, 13]`
`.about(5)` -> [-3, 13]

How does that sound?

ColinEberhardt:
Sounds spot on - the only think I am not sure about is the name. How about one of the following:

 + `symmetricalAround()`
 + `symmetric`
 + `symmetry`

owennw:
Given that this will be the interface the natural question is how will this play with the other properties, specifically `pad` and `include`?

In my opinion `symmetricalAround()` should occur before both the `pad` and the `include`.

ColinEberhardt:
I think so, yes - although hard to say without trying it out against a few use cases.

djmiley:
Does this work correctly with dates? When calling on `fc.util.extent().fields('date').symmetricalAbout(new Date(345345350000))(data);` , my output is `[-757805300000, "Thu Dec 11 1980 01:15:50 GMT+0000 (GMT Standard Time)1103150650000"]`.

Whereas when calling on `fc.util.extent().fields('date').symmetricalAbout(345345350000)(data);`, my output is `[-757805300000, 1448496000000]`

djmiley:
I've looked in https://github.com/ScottLogic/d3fc/blob/master/tests/util/extentsSpec.js and it only makes the promise that it should pad dates. 

Include works though, but probably need some tests on this regard.

PR #733, opened 2015-11-04
Allow extent to support symmetrical domains #728

owennw:


ColinEberhardt:
Looks great :-)

My only request, before merging, is that you roll this new feature out across the website examples. I think it would be useful for MACD, Elder and Force Index - perhaps others.

Thanks :-)

owennw:
Done, raised #734 to deal with the extent for Elder Ray. Took a look at the others and didn't see any candidates.

ColinEberhardt:
As I mentioned in #732, I think I'd prefer to see the component construction separated from the invocation. I think this is easier for people who are new to d3 / d3fc to understand. In this case that would mean the following pattern:

```
var yDomain = fc.util.extent
     .fields(function(d) { return Math.abs(d.macd.macd); })
     .symmetricAbout(0);

foo.domain(yDomain(data))
```

owennw:
I've pushed what I meant to now..

ColinEberhardt:
Nice job :+1:

PR #738, opened 2015-11-05
Random data

owennw:
Note: This is technically a breaking change as now setting the `filter` to `null` will throw an error. If that's a problem this can be made backwards compatible, although I think the design is cleaner as is.

### DIFF
--- a/site/src/components/data/financial.md
+++ b/site/src/components/data/financial.md
@@ -20,7 +20,7 @@ The random financial data generator component is a useful testing utility that g
 {{{example-code}}}
 ```
 
-N.B. There is a default skip weekends filter applied after the data is generated for the specified number of days. Therefore the output will contain less than 10 items.
+N.B. There is a default skip weekends filter applied after the data is generated for the specified number of days. However the output will still contain 10 items.
 
 <pre id="financial"></pre>
 <script type="text/javascript">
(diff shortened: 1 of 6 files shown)


ColinEberhardt:
Thanks for the change - I'm happy with the majority of the changes apart from the new `skipWeekends` property. I think the non standard property implementation is an indication that something isn't quite right here:

https://github.com/ScottLogic/d3fc/pull/738/files#diff-3fcc876522bdd0aa2319befe2d0b4298R146

I think a better alternative would be to have supply a pre-defined filter, something like `fc.data.random.filter.skipWeekends` - this is very similar to the way we make point-snapping pluggable for the crosshair.

owennw:
Is this the sort of thing you were thinking?

ColinEberhardt:
> Is this the sort of thing you were thinking?

Much better :-)

ColinEberhardt:
Sorry it took a little while to get back to this review. While the financial data generator looks to be functionality correct, there are a few aspects of its implementation that I am not so keen on.

The `calculateInterval` performs date manipulation using a `millisecondsPerYear` magic number - have you seen d3's [time interval functions](https://github.com/mbostock/d3/wiki/Time-Intervals), these provide a much clearer alternative to manipulating at the millisecond level.

The overall construction of the algorithm is a bit odd - first it generates an array of data, then recursively tops it up. Could the entire random data construction take place within a simple `while(length < requiredLength)` style loop? Where the filer is applied within the body of the loop?

I'm also not keen on the way that the component mutates the `startPrice` and `startVolume` properties as it is executed. However, I can see that this was not due to your change!

My general feeling is that this component is starting to look a bit messy, so needs a more thorough clean-up.

PR #739, opened 2015-11-05
Playground patches and tweaking live reload set up

jdunkerley:
The site now will include a copy of d3fc in root directory. There should also be no LiveReload in the output of the CI builds.

The playground:
- uses the copy of the d3fc included in site so should avoid version issues
- clears the timeouts/intervals on the iFram

Grunt adjusted to included the components in site:Serve. As this breaks the ability to have both visualTests and site served there is a new bothSites:serve task in GruntFile.


chrisprice:
* I like the local library idea :+1: 
* Is it possible to inject an environment variable into assemble in some way to conditionally switch the live-reload? 
* I'm not sure I understand why the livereload ports had to change? I think it's possible to have concurrent watch jobs running (or I think I've seen it before...)

My gut feel to seeing so many changes in the gruntfile is that there must be a better way. I'm happy to be proved wrong but I'd like to be sure there isn't a simpler way.

jdunkerley:
@chrisprice: port change was back to default as because both site and visual tests are dependent on components couldn't use the the concurrent watch set up. Happy to undo the port change if easier.

Will look into environment variables for Assemble (couldn't spot it immediately), but chose this as a solution to get livereload working properly (currently broken on site:serve) and not be published to website.

jdunkerley:
Closing as should never have done as one PR!

PR #743, opened 2015-11-06
Local d3fc files and Playground fix

jdunkerley:
Copies the d3fc files to the site
Uses these in the Playground
Fix for intervals / timeout in iFrame

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -75,7 +75,7 @@ module.exports = function(grunt) {
                     {
                         expand: true,
                         cwd: 'site/src',
-                        src: ['**/*.md', '*.md', '**/*.hbs', '*.hbs', '!_*/*', '!playground/examples/**'],
+                        src: ['**/*.md', '*.md', '**/*.hbs', '*.hbs', '!_*/*'],
                         dest: 'site/dist'
                     }
                 ]
@@ -175,6 +175,12 @@ module.exports = function(grunt) {
                         cwd: 'site/src/',
                         src: ['**/*', '!_*', '!**/*.hbs', '!**/*.md', '!**/*.yml'],
                         dest: 'site/dist/'
+                    },
+                    {
+                        expand: true,
+                        cwd: 'dist/',
+                        src: ['d3fc.*'],
+                        dest: 'site/dist'
                     }
                 ]
             }
--- a/site/src/_layouts/default.hbs
+++ b/site/src/_layouts/default.hbs
@@ -57,6 +57,6 @@ script-extra: ''
 
     {{#if script}}<script src="{{script}}"></script>{{/if}}
     {{#if script-extra}}{{{script-extra}}}{{/if}}
-    <script src="//localhost:36729/livereload.js"></script>
+    <script src="//localhost:35729/livereload.js"></script>
   </body>
 </html>
--- a/site/src/_layoutsPlayground/component.hbs
+++ b/site/src/_layoutsPlayground/component.hbs
@@ -3,14 +3,15 @@
 <head>
     <meta charset="UTF-8">
     <title>{{default title 'd3fc'}}</title>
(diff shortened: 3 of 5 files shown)


Gruntfile.js
```diff
@@ -75,7 +75,7 @@ module.exports = function(grunt) {
                     {
                         expand: true,
                         cwd: 'site/src',
-                        src: ['**/*.md', '*.md', '**/*.hbs', '*.hbs', '!_*/*', '!playground/examples/**'],
+                        src: ['**/*.md', '*.md', '**/*.hbs', '*.hbs', '!_*/*'],
```
chrisprice: What does this do?

site/src/_layouts/default.hbs
```diff
@@ -57,6 +57,6 @@ script-extra: ''
 
     {{#if script}}<script src="{{script}}"></script>{{/if}}
     {{#if script-extra}}{{{script-extra}}}{{/if}}
-    <script src="//localhost:36729/livereload.js"></script>
+    <script src="//localhost:35729/livereload.js"></script>
```
chrisprice: So this fixes the fact that the port is currently wrong?

site/src/_layoutsPlayground/component.hbs
```diff
@@ -3,14 +3,15 @@
 <head>
     <meta charset="UTF-8">
     <title>{{default title 'd3fc'}}</title>
-    <link href="https://cdnjs.cloudflare.com/ajax/libs/d3fc/{{ package.version }}/d3fc.min.css" rel="stylesheet" />
+    <link href="/d3fc.css" rel="stylesheet" />
+    <!-- https://cdnjs.cloudflare.com/ajax/libs/d3fc/{{ package.version }}/ -->
```
chrisprice: Is this missing the filename?

site/src/_layoutsPlayground/component.hbs
```diff
@@ -3,14 +3,15 @@
...
 </head>
 <body>
     <svg id="chart" width="100%" height="300px"></svg>
 
     <!-- <script src="https://cdnjs.cloudflare.com/ajax/libs/d3/{{ slice package.dependencies.d3 1 }}/d3.js"></script> -->
```
chrisprice: I know this is old but as this isn't the full list of dependencies, and you're referencing the bundle, is it worth removing these?

Gruntfile.js
jdunkerley: I had an example directory in first playground cut. left in gruntfile

site/src/_layouts/default.hbs
jdunkerley: yes

site/src/_layoutsPlayground/component.hbs
jdunkerley: These were for local mode, which is less useful now as we have a local d3fc bundle.

Will retire local mode and update

site/src/_layoutsPlayground/component.hbs
```diff
@@ -3,14 +3,15 @@
...
     <!-- <script src="https://cdnjs.cloudflare.com/ajax/libs/css-layout/{{ slice package.dependencies.css-layout 1 }}/css-layout.js"></script> -->
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/d3fc/{{ package.version }}/d3fc.bundle.min.js"></script>
+    <script src="/d3fc.bundle.js"></script>
```
chrisprice: I'm always nervous about rooted paths, I can't think of a reason why it would matter at the minute but please can you change this to relative?

jdunkerley:
@chrisprice have retired local mode and made paths relative.

Gruntfile.js
chrisprice: Ah :)

chrisprice:
:+1:

PR #744, opened 2015-11-06
Waterfall

owennw:
This is the first pass at a waterfall series -- hopefully the style of inheriting from bar a new bar base class is okay.

There was a suggestion of a generic floating bar series: that is not what I've done here. If that's still interesting then that can be extracted from the waterfall series and refactored.

### DIFF
--- a/site/src/components/series/waterfall.md
+++ b/site/src/components/series/waterfall.md
@@ -1,0 +1,60 @@
+---
+layout: component
+title: Waterfall Series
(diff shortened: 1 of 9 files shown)


ColinEberhardt:
As per our discussion in Slack, I think this needs a bit of re-working. The waterfall concept should be introduced via a non-visual component in the same way that d3 stack layout works.

ColinEberhardt:
relpaced by #772

PR #745, opened 2015-11-06
Playground has Colin's examples

jdunkerley:
Tiny tweaks needed to the examples.

Some minor changes to GruntFile (including support for watch:components at same time as site:serve).

New link on examples page (see http://jdunkerley.github.io/d3fc/examples/bubble/).
image

New section on playground with examples (see http://jdunkerley.github.io/d3fc/playground/).
image

Online version at http://jdunkerley.github.io/d3fc/


site/src/lib/playground.js
```diff
@@ -3,6 +3,7 @@ function createPlayground() {
     'use strict';
 
     var scriptTarget,
+        root,
```
chrisprice: These changes seem pretty brittle,  I wonder if there's a better way? Maybe we could reference the file from the YAML front-matter and have the template spit out a relative path?

Gruntfile.js
```diff
@@ -253,10 +259,12 @@ module.exports = function(grunt) {
                     '<%= meta.metaJsFiles %>',
                     'site/src/**/*'
                 ],
-                tasks: ['site']
+                tasks: ['site'],
+                options: {
+                    livereload: true
```
chrisprice: Like it! Does this mean we can now do `watch:components site:serve`?

Is it worth specifying the port number here (rather than using the default) as we're reliant on it elsewhere?

jdunkerley: yep can do watch:components and site:serve together now. Probably with a little tuning could work a way to get watch:components to kick site rebuild but didnt have in this pass.

Will specify port for clarity

site/src/lib/playground.js
jdunkerley: yes think could do something in the YAML but hadnt found a nice way yet and thought better to get examples in playground sooner rather than later

chrisprice: Ok makes sense, can we raise an issue to track it?

chrisprice:
:+1:

site/src/lib/playground.js
jdunkerley: Added the issue to improve this part.

chrisprice:
Thanks!

chrisprice:
Doh, I think this has broken the build https://travis-ci.org/ScottLogic/d3fc/builds/89974564

PR #748, opened 2015-11-06
Fix hidden axes when used with bootstrap

### LINKED ISSUE #737: Document (or fix) the conflict in catesian chart with Bootstrap's SVG overflow rule
As bootstrap is so prevalent and the issue basically makes it look like the charts are broken, I think we should be more proactive in fixing this.

chrisprice:
Fixes #737


ColinEberhardt:
:+1:

PR #749, opened 2015-11-07
Convert svg.axes-container to g.axes-container

chrisprice:
Alternative fix for #737 

I kept thinking about the previous fix and I couldn't work out why the `.axes-container` needed to be an `svg` element. I've changed it to be a `g` and can't see any difference. Therefore, I've removed the CSS changes as well as removing the `overflow` rule for the element (as `g` elements don't have an `overflow` property)


ColinEberhardt:
I agree, I really can't see any reason why they should be an svg, i can't imagine wanting to clip them

Issue #718, opened 2015-11-02
Add a recreation of "The Wealth & Health of Nations" (with transitions and interaction) to the website


jleft:
Add a recreation of [Mike Bostock's recreation](http://bost.ocks.org/mike/nations/) of Gapminder’s [Wealth & Health of Nations](http://www.gapminder.org/world/).

~~Current work in progress: http://jleft.github.io/wealth-and-health-of-nations/~~

---

(Thought I'd add an issue in case someone was working on something similar.)

ColinEberhardt:
I did add a very simple example using the same data to illustrate a bubble chart here:

http://d3fc.io/examples/bubble/

However, I'd really like to include your more interactive example.

jleft:
Great, sounds good. There's a few more things I'd like to do, then I think I'll be fairly happy it with and ready to put up a PR.

+ [ ] Add axis labels
+ [ ] Update to d3fc 3.x.x -> 4.x.x
+ [ ] Investigate performance

PR #751, opened 2015-11-07
Add website example: The Wealth and Health of Nations

jleft:
#718. I think this is ready for review now - any feedback appreciated.

wealth-and-health-of-nations


ColinEberhardt:
Amazing example, thanks @jleft

site/src/examples/wealth-and-health-of-nations/index.js
```diff
@@ -0,0 +1,232 @@
...
+                label.enter()
+                    .attr('text-anchor', 'end');
+
+                label.text(data.year);
+
+                var overlay = overlayDataJoin(selection, [data]);
+
+                overlay.enter()
+                    .each(function() {
+                        labelBox = label.node().getBBox();
```
ColinEberhardt: that's a neat idea, grabbing the BBox via `each` to keep the whole thing 'fluent'

site/src/examples/wealth-and-health-of-nations/index.js
```diff
@@ -0,0 +1,232 @@
...
+            .yDomain([10, 85])
+            .margin({left: 30, bottom: 20})
+            .yOrient('left')
+            .xTicks(12, d3.format(',d'))
+            .plotArea(point)
+            .decorate(function(selection) {
+                var plotAreaContainerEnter = selection.enter()
+                    .select('.plot-area-container');
+
+                plotAreaContainerEnter.append('g')
```
ColinEberhardt: This feel like quite a bit of effort to add the labels. Could you use the existing label, styling it with `text-anchor: start;`, then in decorate apply a translation?

site/src/examples/wealth-and-health-of-nations/index.css
```diff
@@ -0,0 +1,41 @@
+#wealth-and-health-of-nations {
+  margin: 20px auto;
+  width: 100%;
+  height: 400px;
+}
+#wealth-and-health-of-nations > .cartesian-chart,
+.plot-area-container > .plot-area {
+  overflow: visible;
```
chrisprice: This shouldn't be necessary anymore with the latest from master.

chrisprice:
Not sure how the examples page works anymore (hopefully it'll be replaced with the playground page soon!), the link appears on the right but not in the thumbnail -

image

site/src/examples/wealth-and-health-of-nations/index.js
chrisprice: Only if you're expecting a singleton...

chrisprice:
Looks really cool, I suppose it's just a pity that the library isn't really saving us much code :(

ColinEberhardt:
> Looks really cool, I suppose it's just a pity that the library isn't really saving us much code :(

True - I couldn't spot any obvious candidates for re-useable components. However, my suggestion to re-use the cartesian axis labels should save ~20 lines.

> Not sure how the examples page works anymore 

It is a bit of a mess - I'd like to split it into two groups of examples and add a bit of general commentary to the page so that people know why it is distinct from the component documentation. Anyhow, I'll wait until @jdunkerley has finished his playground changes before doing anything.

jdunkerley:
@ColinEberhardt the playground landing page has an assemble built examples section which I think could be cut and pasted into the examples page if we want. Makes it the same way as the frontpage but with thumbnail images and added a text label to be in keeping with other gallery sections.

site/src/examples/wealth-and-health-of-nations/index.js
jleft: I've updated to use the existing labels and apply a translation.

Using `text-anchor` didn't work quite as I was hoping. As it positions the text relative to the centre of the axis label container, I couldn't see a nice way to use this to align the labels precisely with the end of the axes.

This is where the x-axis label was positioned when using `text-anchor: start;`

label-alignment

I settled on this: 

```js
var xAxisWidth = selection.select('.x-axis')
    .layout('width');

selection.select('.x-axis .label')
    .attr('transform', 'translate(' + xAxisWidth / 2 + ', -26)');
```

axis-label-aligned


Not sure if there's a neater way to achieve the same thing?

site/src/examples/wealth-and-health-of-nations/index.css
jleft: Seems like these styles are still required, as `svg:not(:root) { overflow: hidden; }` needs to be overridden for `.cartesian-chart`, and I want to override the default overflow (`hidden`) of `.plot-area`.

jleft:
> Amazing example, thanks @jleft

Thanks @ColinEberhardt!

> Not sure how the examples page works anymore (hopefully it'll be replaced with the playground page soon!), the link appears on the right but not in the thumbnail -

Thanks for pointing this out, @chrisprice, I've added the thumbnail.

> Looks really cool, I suppose it's just a pity that the library isn't really saving us much code :(

Yeah, but I felt that Cartesian chart made it very easy to get an initial version of the chart up and running.

The year label was the only part which I thought was trickier to add; I'm not sure how d3fc could have helped with this... maybe a text rendering component and/or a rect rendering component, but they seem to general (not sure if it would be worth having a component which only adds text or a rectangle - it's easy enough to do with just d3); conversely, a year label component would be far too specialised!

---

Thanks for the feedback, I think I've addressed all of the comments - ready for another review.

site/src/examples/wealth-and-health-of-nations/index.css
chrisprice: I hadn't realised the data breached the x domain, that seems like a pretty specialist use-case so happy with that one. I think you can get rid of the first rule using the margin settings through e.g. -
```js
.margin({top: 1, right: 20, bottom: 20, left: 30})
```

site/src/examples/wealth-and-health-of-nations/index.js
```diff
@@ -0,0 +1,223 @@
...
+                var yAxisHeight = selection.select('.y-axis')
+                    .layout('height');
+
+                selection.select('.y-axis .label')
+                    .attr('transform', 'rotate(-90) translate(' + yAxisHeight / 2 + ', 36)');
+
+                var plotAreaContainerEnter = selection.enter()
+                    .select('.plot-area-container');
+
+                plotAreaContainerEnter.append('g')
```
ColinEberhardt: Sorry, I think my original comment was lost ... anyhow, this feels like quite a bit of effort to add the labels. Have you tried grabbing the Cartesian labels within decorate and adding a transform to offset them?

You can use CSS to move them from the centre of the axis to the right / top edge.

chrisprice: This seems to work -

```js
.decorate(function(selection, data) {

                selection.select('.x-axis>g')
                    .layout({
                      'position': 'absolute',
                      'right': 0,
                      'bottom': 26
                    });

                selection.select('.y-axis>g')
                    .layout({
                      'position': 'absolute',
                      'top': 0,
                      'left': 36
                    });

                var plotAreaContainerEnter = selection.enter()
                    .select('.plot-area-container');

                plotAreaContainerEnter.append('g')
                        .attr('class', 'year label')
                        .layout({
                            position: 'absolute',
                            right: 0,
                            bottom: 24
                        });

                selection.layout();

                // Update the year label when updating the chart
                selection.select('.year.label')
                    .datum(data.year)
                    .call(yearLabel);
            });
```

site/src/examples/wealth-and-health-of-nations/index.js
```diff
@@ -0,0 +1,223 @@
...
+                plotAreaContainerEnter.layout();
+
+                // Update the year label when updating the chart
+                selection.select('.year.label')
+                    .datum(data.year)
+                    .call(yearLabel);
+            });
+
+        function chart(selection) {
+            selection.each(function(data) {
```
chrisprice: I don't think there's any need for the `.each` anymore.

site/src/examples/wealth-and-health-of-nations/index.css
jleft: Nice idea - I didn't think of use margin settings :). Updated.

site/src/examples/wealth-and-health-of-nations/index.js
ColinEberhardt: > Sorry, I think my original comment was lost ... 

Ignore me - Github was doing that thing of collapsing my comment because it was on an old diff. Good to see that adding a label translation has simplified things.

jleft: I like the idea of using flexbox layout here, it feels nicer than having to do the manual calculations involving the width/height of the parent container - I've updated to reflect this.

jleft:
I've made all of the suggested changes. Please let me know if there are any further changes to make :)

Issue #753, opened 2015-11-08
OHLC/candlestick enter transition grows from left instead of center


chrisprice:
Therefore it can overlap and just generally looks wrong.

Issue #752, opened 2015-11-08
Gridlines don't transition as expected


chrisprice:
There's no way to set the key functions so you can only get an index based transition.

PR #756, opened 2015-11-08
Fix transitions

chrisprice:
Fixes for #753 and #752


ColinEberhardt:
LGTM - part of the problem with the current visual test for transitions is that it scrolls in the x direction whilst the enter transition is in progress. I slowed it down to check the ohlc / candlestick enter transitions, but didn't spot this issue as a result of the scrolling.

PR #758, opened 2015-11-08
Fix Point Example

jdunkerley:
Fix Travis problem


chrisprice:
:unamused: :palm_tree: 

:+1: 

(In other news, give GitHub a rest and checkout Kung Fury!)

PR #759, opened 2015-11-09
Create Largest Triangle 3-Bucket subsampler

WPFerg:
Implements a largest triangle sampler, looking at the selected value in the previous bucket, each data point in the current bucket, and the average x, y in the next bucket.

### DIFF
--- a/site/src/components/sampler/largestTriangle3.md
+++ b/site/src/components/sampler/largestTriangle3.md
@@ -1,0 +1,60 @@
+---
+layout: component
+title: Largest Triangle 3 Sampling
+component: data/sampler/largestTriangle3.js
+namespace: samplers
+
+example-code: |
+    // create some random data
+    var data = fc.data.random.financial()(1000);
+
+    // configure the sampler
+    var sampler = fc.data.sampler.largestTriangle3()
+      .bucketSize(20)
+      .xValue(function(d) { return d.date; })
+      .yValue(function(d) { return d.low; });
+
+    // sample the data
+    var sampledData = sampler(data);
+
+    xScale.domain(fc.util.extent().fields('date')(sampledData))
+    yScale.domain(fc.util.extent().fields('low')(sampledData));
+
+    // render the sampled data
+    var sampledLine = fc.series.line()
+        .xScale(xScale)
+        .yScale(yScale)
+        .yValue(function (d) { return d.low; });
+
+    container.append('g')
+         .datum(sampledData)
+         .call(sampledLine);
+
+    // render the original data
+    var originalLine = fc.series.line()
+         .xScale(xScale)
+         .yScale(yScale)
+         .yValue(function (d) { return d.low; })
+         .decorate(function(selection) {
+             selection.enter().style('stroke-opacity', '0.5');
+         });
+
+    container.append('g')
+          .datum(data)
+          .call(originalLine);
+
+---
+
+The Largest Triangle 3 sampling component is a method of subsampling data to improve performance with large data sets. The algorithm entails calculating the largest areas between a data point and its neighbouring buckets.
+
+The sampler requires both the `xValue` and `yValue` properties in order to generate calculate the area. You can configure the sampling frequency by setting the `bucketSize` property.
+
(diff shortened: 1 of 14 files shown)


site/src/components/sampler/largestTriangle3.md
```diff
@@ -0,0 +1,60 @@
...
+         .xScale(xScale)
+         .yScale(yScale)
+         .yValue(function (d) { return d.low; })
+         .decorate(function(selection) {
+             selection.enter().style('stroke-opacity', '0.5');
+         });
+
+    container.append('g')
+          .datum(data)
+          .call(originalLine);
```
bjedrzejewski: Consistent spacing would be good here.

WPFerg:
Just pushed up a change that should hopefully have better indentation.

src/data/sampler/largestTriangle3.js
```diff
@@ -0,0 +1,101 @@
...
+            lastSelectedY = yValue(highestItem);
+
+            return highestItem;
+        });
+
+        // First and last data points are their own buckets.
+        return [].concat(data[0], subsampledData, data[data.length - 1]);
+    };
+
+    function getBuckets(data) {
```
ColinEberhardt: This looks very similar to the mode-median getBuckets function. Can you split that out as a separate component? e.g.

```
var bucket = tfc.data.sampler.bucket().bucketSize(20);
var bucketedData = bucket(data);
```

You can then re-use this component in both of your samplers, re-binding it to expose `bucketSize`.

tests/data/sampler/largestTriangle3Spec.js
```diff
@@ -0,0 +1,62 @@
...
+
+    var data;
+    var dataGenerator;
+
+    describe('simple xy data', function() {
+
+        beforeEach(function() {
+
+            data = [
+                { x: 0, y: 6 }, // Bucket 1
```
ColinEberhardt: How did you create this test data? Would it be possible to use someone else's implementation as the basis for your test?

https://github.com/pingec/downsample-lttb

src/data/sampler/largestTriangle3.js
```diff
@@ -0,0 +1,101 @@
+import d3 from 'd3';
+import {identity, noop} from '../../util/fn';
+
+export default function() {
+
+    var bucketSize = 10,
+        xValue = identity,
+        yValue = identity;
+
+    var largestTriangle3 = function(data) {
```
ColinEberhardt: I'm not keen on the name - how about lttb?

ColinEberhardt:
This looks really good, a few minor comments, but other than that - I like it :+1:

tests/data/sampler/largestTriangle3Spec.js
WPFerg: I hand-worked these out, but I've also ran it on the codebase above and have similar results. They're not identical, however, due to slight implementation details.

For example, the first and last buckets in the above module are always first and last, meaning the first/last few data points are always discarded. To counteract this, I don't discard these data points, but evenly partition between elements `1..n-1`. These are the only differences in the expected output in the algorithm, the data points in the middle are identical otherwise :)

WPFerg:
Just pushed up a change for bringing `bucket` into it's own module. I'll do another PR to change mode-median to that after. I've left the name as-is now, as that's something that can be easily changed

ColinEberhardt:
@WPFerg thanks for the updates, rather than change the name at a later date and refactor mode-medan, my preference would be to do it as part of this PR. Tidy up work has a habit of being forgotten, or people get side-tracked ;-)

For now, how about calling it `largestTriangleThreeBucket` as discussed?

WPFerg:
No problem -- I've made the changes.

ColinEberhardt:
Thanks for the update - one very minor thing, you can use `d3.rebind` to expose the properties of wrapped components. I've made a minor update and will merge this in via a different PR.

As discussed, it would be great to see this in action with a more realistic example. Rather than financial data, how about using some other time-series data, such as temperature data? I'm pretty sure you can find that in large volumes on the web.

@jdunkerley any idea how we integrate this nicely with the playground?

PR #769, opened 2015-11-10
Line annotations now have an orientation class

### LINKED ISSUE #767: Could be easier to style crosshair
The crosshair doesn't add any classes to help select its child elements - the vertical and horizontal line annotations are particularly hard to differentiate.

It would be nice if classes were added to help select the relevant elements (possibly in a way similar to the indicator renderers).

Potentially a separate issue, but a potential solution, is that the annotation component doesn't add a class to indicate its orientation - maybe that's something it could do?

---

My use case is that I'd like to hide the horizontal line. Would be good if something like this would work:

```css
.crosshair > .horizontal {
  visibility: hidden;
}
```

ColinEberhardt:
Fixes #767

### DIFF
--- a/src/annotation/line.js
+++ b/src/annotation/line.js
@@ -15,8 +15,7 @@ export default function() {
 
     var dataJoin = _dataJoin()
         .selector('g.annotation')
-        .element('g')
-        .attr('class', 'annotation');
+        .element('g');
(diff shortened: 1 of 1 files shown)


jleft:
:+1:

PR #770, opened 2015-11-10
A few websites tweaks (and deleted linearTimeSeries!)

ColinEberhardt:
A number of website changes:

 - Change getting started guide to use cartesian rather than linear time series
 - Deleted linearTimeSeries, I think we've pretty much settled on cartesian as the 'base' chart
 - Deleted Yahoo Finance chart - this was aready slightly broken (issues with the line annotation call-out rendering). With linear time series being removed, it would need a big re-write. Also, I'd like to get rid of the HTML legend too!
 - Added a page that describes transition support


chrisprice:
Just the one comment.

ColinEberhardt:
Thanks @chrisprice - example simplified as per our discussion, looks much better.

PR #772, opened 2015-11-11
Add a waterfall series type closes #591

### LINKED ISSUE #591: Add a waterfall series type
Very similar to a bar chart series type but bars start at the position of the last bar.

Main complication is that you would need to access the value of the previous point in the data series.

owennw:
V2 of the waterfall series.

### DIFF
--- a/site/src/components/series/waterfall.md
+++ b/site/src/components/series/waterfall.md
@@ -1,0 +1,69 @@
+---
+layout: component
+title: Waterfall Series
+component: series/waterfall.js
+tags:
+  - playground
+namespace: series
+
+example-code: |
+  var data = [
+    { month: 'January', profit: 4000 },  
+    { month: 'February', profit: 2000 },
+    { month: 'March', profit: -1000 },
+    { month: 'April', profit: 1500 },
+    { month: 'May', profit: 100 },
+    { month: 'June', profit: 500 },
+    { month: 'July', profit: -100 },
+    { month: 'August', profit: 800 },
+    { month: 'September', profit: 1200 },
+    { month: 'October', profit: 1500 },
+    { month: 'November', profit: 1400 },
+    { month: 'December', profit: 2000 }
+  ];
+
+  var waterfallData = fc.series.algorithm.waterfall()
+      .xValueKey('month')
+      .yValue(function(d) { return d.profit; })
+      .startsWithTotal(true)
+      .total(function(d, i, data) {
+          if ((i + 1) % 3 === 0) {
+              return 'Q' + ((i + 1) / 3) + ' total';
+          }
+      })(data);
+
+  var xScale = d3.scale.ordinal()
+      .domain(waterfallData.map(function(d) { return d.x; }))    
+      .rangeRoundBands([0, width], 0.1);
+  yScale.domain(fc.util.extent().fields('y1').pad(0.2).include(0)(waterfallData));
+
+  var waterfall = fc.series.waterfall()
+      .xScale(xScale)
+      .yScale(yScale);
+
+  container.append('g')
+      .datum(waterfallData)
+      .call(waterfall);
+---
+
+The [Waterfall series](https://en.wikipedia.org/wiki/Waterfall_chart) renders the given data as a series of vertical bars, showing the cumulative effect of each bar. The series begins and ends with the total values.
+
+All bars are classed with 'waterfall'. Ascending and descending bars are classed with 'up' and 'down' respectively.
+
+The data must be shaped using `fc.series.algorithm.waterfall()`. The new data has a `x` property specifying the x-location, `y0` and `y1` properties specifying the y0 and y1-locations respectively, and a `direction`. The algorithm assumes each object has an `x` property which defines its x-location, and a `y` property which defines its y-location. However, these can be changed via the `xValueKey` and `yValue` properties.
+
+The first column can be declared as a total by using the `startsWithTotal` property. Other total columns can be inserted into the data by using the algorithm's `totals` property. For a given datapoint this property should return the x-location for the total bar to be inserted. If the `totals` are not specified a final total bar will be inserted by default.
+
+```js
+{{{example-code}}}
+```
+
+Which gives the following:
+
+{{>example-fixture}}
+
+The data that identifies the x-value is identified via the `xValueKey` property.
+
+This series has the same `yValue` and `decorate` properties as the [point series](./point). You can also specify the width of the bars via the `barWidth` property.
+
+The waterfall series supports horizontal display, using the `orient` property, similar to [bar](./bar).
--- a/src/series/algorithm/waterfall.js
+++ b/src/series/algorithm/waterfall.js
@@ -1,0 +1,103 @@
+import d3 from 'd3';
+import {noop} from '../../util/fn';
+
+export default function() {
+
+    var xValueKey = '',
+        yValue = function(d) { return d.y; },
+        startsWithTotal = false,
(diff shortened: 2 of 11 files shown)


src/series/algorithm/waterfall.js
```diff
@@ -0,0 +1,100 @@
+import d3 from 'd3';
+import {noop} from '../../util/fn';
+
+export default function() {
+
+    var xValue = function(d) { return d.date; },
```
ColinEberhardt: The [long term vision](https://github.com/ScottLogic/d3fc/wiki/d3fc-Vision) is to split the d3fc core and financial concepts, for that reason I wouldn't use default values of `date` and `close` in the waterfall, I doubt many people render waterfall charts of financial data!

src/series/algorithm/waterfall.js
```diff
@@ -0,0 +1,100 @@
...
+        var length = data.length,
+            i,
+            previousEnd = yValue(data[0]),
+            start,
+            end,
+            total,
+            result = [];
+
+        // First value is a total
+        result.push({
```
ColinEberhardt: is its always? see: https://en.wikipedia.org/wiki/Waterfall_chart#/media/File:Waterfallchart_ex2.jpg

src/series/waterfall.js
```diff
@@ -0,0 +1,31 @@
...
+                g.enter()
+                    .attr('class', 'waterfall ' + bar.orient())
+                    .classed('up', function(d) { return d.direction === 1; })
+                    .classed('down', function(d) { return d.direction === -1; });
+            });
+
+        bar(selection);
+    };
+
+    d3.rebind(waterfall, bar, 'xScale', 'xValue', 'x1Value', 'x0Value', 'yScale', 'yValue', 'y1Value', 'y0Value', 'decorate', 'barWidth', 'orient', 'key');
```
ColinEberhardt: How about using `fc.util.rebindAll`?

ColinEberhardt:
This is a big improvement, I like the waterfall series is a simple extension of the bar series.

The algorithm looks OK, but has a couple of issues:

  1. It assumes that the start will always be a total
  2. I'm not keen on the way that the user provides 'totals' as an array.  

As an alternative to (2) why not use a visitor pattern?

Let's say we have some data which contains the monthly profits:

```
{
   { date: 2011-01-01, profit: 4000 },
   { date: 2011-02-01, profit: -2000 },
   ...
   { date: 2011-12-01, profit: 8000 },
}
```

I'd like to render this as a waterfall with quarterly totals, as follows:

```
var data = { ... see above ... };

// I'd expect to render this on an ordinal axis, so convert each date into a string ...
data.forEach(function(d) {
  d.month = d3.format(d.date, 'blah'); 
});

var algorithm = fc.series.algorithm.waterfall()
  .total(function(datum, index, data) {
    if (datum.date.month() % 4 === 0) {
       return "Q" + Math.floor(datum.date.month() / 4) + " total"; 
    }
  });

var waterfall = algorithm(data);
```

Here the `total` function takes a 'callback' with arguments of (1) the current datapoint, (2) its index, (3) the entire array of data. If it returns a value, this is used as the 'x' component and a total is constructed. If not, no total is added.

You can create a default total implementation that always creates a total at the end of the series.

It feels like a bit of an anti-pattern to rely on a function without an explicit return actually returning `undefined` - might be worth asking @chrisprice for his opinion on this.

As the component has to both read and write the x value, I don't think a value accessor is appropriate. If you look at the `spread` component, it uses an x value key for the same reason.

I'd also add a property:

```
fc.series.algorithm.waterfall().startsWithTotal(true);
```

This would be used to indicate that the first datum in the input array of data is actually a total.

src/series/algorithm/waterfall.js
owennw: What I mean by 'total' is that the column has its full height, i.e. `y0` is 0. In a sense it also is a total, by virtue of being the first point.

site/src/components/series/waterfall.md
```diff
@@ -0,0 +1,69 @@
...
+    { month: 'June', profit: 500 },
+    { month: 'July', profit: -100 },
+    { month: 'August', profit: 800 },
+    { month: 'September', profit: 1200 },
+    { month: 'October', profit: 1500 },
+    { month: 'November', profit: 1400 },
+    { month: 'December', profit: 2000 }
+  ];
+
+  var waterfallData = fc.series.algorithm.waterfall()
```
ColinEberhardt: This looks a lot better - a really flexible API.

src/series/algorithm/waterfall.js
```diff
@@ -0,0 +1,107 @@
...
+            end,
+            total,
+            result = [],
+            finalIndex = length - 1,
+            customTotals = false;
+
+        if (startsWithTotal) {
+            // First value is a total
+            result.push({
+                domain: data[0][xValueKey],
```
ColinEberhardt: I'm not keen on the property names that the algorithm outputs, how about

 - domain => x
 - start => y0
 - end => y1

This is similar to the [d3 stack layout](https://github.com/mbostock/d3/wiki/Stack-Layout) datum properties.

src/series/algorithm/waterfall.js
```diff
@@ -0,0 +1,107 @@
...
+                start: 0,
+                end: previousEnd,
+                direction: 0
```
ColinEberhardt: Rather than -1, 0, 1 which are not easy to understand, I'd prefer 'up', 'down', 'unchanged' - again there is a precedent in d3 for using strings as enumerations

src/series/algorithm/waterfall.js
```diff
@@ -0,0 +1,107 @@
...
+                    start: 0,
+                    end: end,
+                    direction: 0
+                });
+            }
+
+            previousEnd = end;
+        }
+
+        if (!customTotals) {
```
ColinEberhardt: I'm pretty sure you could remove this logic by having `totals` default to a function that generates a total for the last data point.

ColinEberhardt:
Thanks @owennw a big improvement, a few comments on the algorithm implementation. Also looking at the visual tests, I'm not sure the rendered series looks correct for the input data?

One final thing, the algorithm could really do with some unit tests!

src/series/algorithm/waterfall.js
owennw: The 0 isn't so much 'unchanged' (it can be), but demarks a totals column which has its css default to a standard bar colour. Perhaps that logic can be described by 'unchanged' though.

ColinEberhardt: yeah - I wasn't sure of what to call '0', happy for you to propose an alternative.

ColinEberhardt:
Thanks for the updates - a big improvement, pretty much ready to merge. However on further testing I do think the algorithm is wrong, it doesn't maintain a running total.

As an example:

```
        var data = [
            { quarter: 'Q1', profit: 4000 },
            { quarter: 'Q2', profit: 2000 },
            { quarter: 'Q3', profit: -1000 },
            { quarter: 'Q4', profit: 1500 }
        ];
```

This should result in 3 rising bars, for three quarters with profit, and one falling. This is what you actually get:

screen shot 2015-11-26 at 08 13 52

By changing the algorithm to maintain a total, this gives the correct result:

screen shot 2015-11-26 at 08 14 24

Once this is fixed, I'm happy to merge

ColinEberhardt:
Just awaiting a squash on this one ...

ColinEberhardt:
Thanks @owennw  :-)

PR #773, opened 2015-11-11
WIP: Sampler Example

WPFerg:
Uses a large data set with the low barrel example, with automatic zooms and choice between sampling method.


site/src/examples/sampler/index.css
```diff
@@ -0,0 +1,67 @@
...
+}
+
+.candlestick.down {
+  fill: #c60;
+}
+
+.bar {
+  stroke: #06c;
+  fill: #9cf;
+}
```
chrisprice: Please can you add a newline at the end of the file?

site/src/examples/sampler/index.js
```diff
@@ -0,0 +1,399 @@
+/* global d3:false, fc:false */
+    'use strict';
```
chrisprice: The indentation seems to be a bit awry, please could you reformat to use 4 spaces?

chrisprice:
Thanks @WPFerg, I think the example is great and a really compelling visualisation of what sampling is about and the benefits of using it. However, as there's so much commonality between this and the low barrel example (through no fault of your own!) I would like to hold off merging this. 

There's some on-going work to extract the common components of the low-barrel example into the library itself. Once this has been done the amount of code in your example should drop dramatically and it'll be easier to comment on the specifics of your code.

As such I've marked this as a WIP and I've added a couple of high-level comments, but I'll save the rest for later.

WPFerg:
No problem, I've made the indentation changes. I've got another PR coming for one more sampler that I'd like to integrate in to this, so I'll revisit it later.

ColinEberhardt:
@WPFerg I'm going to close this for now - good to see the code being made good use of in your blog post. Once we have a navigator component I'd like to revisit this example and potentially bring it into the project.

PR #775, opened 2015-11-11
SVG Tooltip Component

ColinEberhardt:
This component is intended as a replacement to the HTML legend.

Our examples integrate with d3-svg-legend, which is rendered based on a scale rather than a data-join. For low-barrel and tooltips we need a component that is driven from the data-join. This PR includes a tooltip which does just that.

As well as a tooltip I've added a `container` component which can render a background behind tooltips, legends or other components. I'm considering adding an `anchor` property which accepts `top-left`, `top-right` etc... as a way to specify the origin of the container.

Via decoration it is very easy to make the tooltip either follow the crosshair origin, or render it at one of the corners of the plot area.

I'm pleased with the reduction of code in the low-barrel example!

One problem I am still struggling with is how to make the tooltip render at different locations based on the location within the plot area in order to avoid cropping.


visual-tests/chart/tooltip.js
```diff
@@ -0,0 +1,72 @@
...
+        .height(70)
+        .margin(5)
+        .child(tooltip);
+
+    var crosshair = fc.tool.crosshair()
+        .snap(fc.util.seriesPointSnapXOnly(candlestick, data.series))
+        .xLabel('')
+        .yLabel('')
+        .decorate(function(selection) {
+            selection.select('.point')
```
ColinEberhardt: By selecting `.point` this causes the tooltip to follow the crosshair. Nice :-)

src/chart/container.js
```diff
@@ -0,0 +1,86 @@
+import d3 from 'd3';
+import dataJoinUtil from '../util/dataJoin';
+import {noop} from '../util/fn';
+import expandMargin from '../util/expandMargin';
+
+export default function() {
+
+    var width = 50,
+        height = 50,
+        child = noop,
```
chrisprice: `component`?

src/chart/container.js
```diff
@@ -0,0 +1,86 @@
...
+        padding = 5;
+
+    var dataJoin = dataJoinUtil()
+        .selector('g.container')
+        .element('g')
+        .attr('class', 'container');
+
+    var container = function(selection) {
+        selection.each(function(data, index) {
+            var expandedMargin = expandMargin(margin);
```
chrisprice: Seems like a good candidate for `layout()`?

src/chart/container.js
```diff
@@ -0,0 +1,86 @@
...
+            var expandedPadding = expandMargin(padding);
+
+            var g = dataJoin(this, [data]);
+
+            var gEnter = g.enter();
+            gEnter.attr('transform', 'translate(' + expandedMargin.left + ', ' + expandedMargin.top + ')');
```
chrisprice: Implementation allows but ignores `right` and `bottom`.

src/chart/tooltip.js
```diff
@@ -0,0 +1,91 @@
...
+                return {
+                    datum: data,
+                    label: d3.functor(item[0]),
+                    value: d3.functor(item[1])
+                };
+            });
+
+            var g = dataJoin(this, legendData);
+
+            var gEnter = g.enter()
```
chrisprice: Just keep access `g.enter()`?

src/chart/tooltip.js
```diff
@@ -0,0 +1,91 @@
...
+                .attr('transform', function(d, i) {
```
chrisprice: Another candidate for `layout()`?

src/chart/tooltip.js
```diff
@@ -0,0 +1,91 @@
...
+
+    tooltip.decorate = function(x) {
+        if (!arguments.length) {
+            return decorate;
+        }
+        decorate = x;
+        return tooltip;
+    };
+
+    tooltip.rowHeight = function(x) {
```
chrisprice: Could be removed with `layout()`

src/chart/tooltip.js
```diff
@@ -0,0 +1,91 @@
...
+        if (!arguments.length) {
+            return rowHeight;
+        }
+        rowHeight = x;
+        return tooltip;
+    };
+
+    tooltip.labelWidth = function(x) {
```
chrisprice: Could be made a percentage with `layout()`

ColinEberhardt:
@chrisprice I think this is ready for review - sorry it's a bit of a big one!

Summary of changes:

 + new tooltip component, uses the same pattern as the HTML legend and your tooltip in low-barrel, basically an 'items' property that defines an array of labels / values. Internally it uses flexbox for layout (which works quite nicely), as an added bonus you can change the container to `flexDirection:row` to change the tooltip orientation. Nice!
+ small update to `d3.selection.layout()` hard to see as a result of the rename, it no longer sets the `transform` for the root node. With this set-up I expect the rectangles component to set the location of each child 'rectangle', together with layout-width / layout-height. I then expect to be able to use layout on the children - prior this change it was destroying the translation applied by the rectangles layout
+ new rectangles layout component - does what it says on the tin.
+ `withinContainer` strategy (crappy name), ensures that rectangles are within the parent container bounds.
+ container component now uses flexbox
+ crosshair and snap is now always in screen coords
+ visual test for 'rectangles' demonstrates the pluggable nature of the strategy by adding a non-colliding strategy. 

TODO (post merge):
+ Try to render 'swatches' with the tooltip
+ Make it easier to set width / height for layout, currently you have to use `attr('layout-width', foo)`, the layout API allows you to set width / height via layout(width, height), but this also invokes the layout algorithm. I don't think we have ever used this API method.
+ There is a big in low-barrel (which has been there for a while), where the crosshair 'point' is rendered in both series. Previously it just happened to work because the two charts have different y-domains.

src/layout/flexbox.js
```diff
@@ -0,0 +1,178 @@
...
+    node.element.setAttribute('layout-x', node.layout.left);
+    node.element.setAttribute('layout-y', node.layout.top);
+
+    //for svg / rect set the position / dimensions via x/y/width/height properties
+    if (node.element.nodeName.match(/(?:svg|rect)/i)) {
+        node.element.setAttribute('width', node.layout.width);
+        node.element.setAttribute('height', node.layout.height);
+        node.element.setAttribute('x', node.layout.left);
+        node.element.setAttribute('y', node.layout.top);
+    } else if (subtree) {
```
chrisprice: Could the subtree logic also apply to `svg`/`rect` nodes i.e. should it cover both this and the previous block?

src/chart/tooltip.js
```diff
@@ -0,0 +1,87 @@
...
+
+    tooltip.decorate = function(x) {
+        if (!arguments.length) {
+            return decorate;
+        }
+        decorate = x;
+        return tooltip;
+    };
+
+    tooltip.labelWidth = function(x) {
```
chrisprice: Could this be called `split` or something less `width` orientated? I really like that by flipping flex-direction you can change the orientation but then that seems at odds with calling it `width`.

src/chart/tooltip.js
```diff
@@ -0,0 +1,87 @@
...
+                });
+
+            g.select('.value')
+                .text(function(d, i) {
+                    return d.value.call(this, d.datum, i);
+                });
+
+            decorate(g, data, index);
+
+            container.layout();
```
chrisprice: I think technically this could be done outside the each but I'm not sure it makes any difference.

src/chart/tooltip.js
```diff
@@ -0,0 +1,87 @@
...
+import dataJoinUtil from '../util/dataJoin';
+import {noop} from '../util/fn';
+
+export default function() {
+
+    var labelWidth = 50,
+        decorate = noop;
+
+    var items = [
+        ['datum:', function(d) { return d.datum; }]
```
chrisprice: Unsure about the structure, I don't like unnamed arguments so I would prefer to see an `{label:  () =>, value: () => }[]` but I also like the brevity. 

Maybe it could be done similar to how d3 does it on ordinal scales e.g. 

```js
  tooltip()
    .labels([a, b, c])
    .values([a, b, c])
```

src/layout/rectangles.js
```diff
@@ -0,0 +1,121 @@
...
+            yRange = range(yScale);
+        strategy.containerWidth(Math.max(xRange[0], xRange[1]))
+            .containerHeight(Math.max(yRange[0], yRange[1]));
+
+        selection.each(function(data, index) {
+            var g = dataJoin(this, data);
+
+            // obtain the rectangular bounding boxes for each child
+            var childRects = data.map(function(d, i) {
+                return {
```
chrisprice: I'm wondering about the accessors and whether they're appropriate for all situations. E.g. if I want to measure bounding boxes then I'll need to call a method which will return both width and height but I'll only be able to supply them one at a time. Maybe the accessors could be `position` and `size` instead?

src/layout/rectangles.js
```diff
@@ -0,0 +1,121 @@
...
+            var layout = strategy(childRects);
+
+            // offset each rectangle accordingly
+            g.attr('transform', function(d, i) {
+                var offset = layout[i];
+                return 'translate(' + offset.x + ', ' + offset.y + ')';
+            });
+
+            // set the layout width / height so that children can use SVG layout if required
+            g.attr('layout-width', function(d, i) { return childRects[i].width; });
```
chrisprice: Can this be done with one of the `layout` shortcuts?

src/layout/rectangles.js
```diff
@@ -0,0 +1,121 @@
...
+                    width: width(d, i),
+                    height: height(d, i)
+                };
+            });
+
+            // apply the strategy to derive the layout
+            var layout = strategy(childRects);
+
+            // offset each rectangle accordingly
+            g.attr('transform', function(d, i) {
```
chrisprice: Any reason for not chaining these calls?

src/layout/rectangles.js
```diff
@@ -0,0 +1,121 @@
...
+
+            // set the layout width / height so that children can use SVG layout if required
+            g.attr('layout-width', function(d, i) { return childRects[i].width; });
+            g.attr('layout-height', function(d, i) { return childRects[i].height; });
+
+            g.call(component);
+        });
+    };
+
+    rebindAll(rectangles, strategy);
```
chrisprice: Does it make sense to consider a prefix here?

src/layout/rectangles.js
```diff
@@ -0,0 +1,121 @@
...
+
+    var rectangles = function(selection) {
+
+        var xRange = range(xScale),
+            yRange = range(yScale);
+        strategy.containerWidth(Math.max(xRange[0], xRange[1]))
+            .containerHeight(Math.max(yRange[0], yRange[1]));
+
+        selection.each(function(data, index) {
+            var g = dataJoin(this, data);
```
chrisprice: Hm this is another implementation of dataJoin on one dataset but use properties from another. It's not exactly the same as multi but I do wonder whether it's worth considering breaking that out from multi in the future.

src/tool/container.js
```diff
@@ -0,0 +1,57 @@
...
+        component = noop;
+
+    var dataJoin = dataJoinUtil()
+        .selector('g.container')
+        .element('g')
+        .attr({'class': 'container', 'layout-style': 'flex: 1'});
+
+    var container = function(selection) {
+        selection.each(function(data, index) {
+            var expandedPadding = expandMargin(padding);
```
chrisprice: The util method should be renamed.

src/tool/container.js
```diff
@@ -0,0 +1,57 @@
...
+                    position: 'absolute',
+                    top: expandedPadding.top,
+                    left: expandedPadding.left,
+                    bottom: expandedPadding.bottom,
+                    right: expandedPadding.right
+                });
+
+            d3.select(this).layout();
+
+            g.select('.inner').call(component);
```
chrisprice: I don't think you need the class name here

src/tool/container.js
```diff
@@ -0,0 +1,57 @@
...
+
+    var padding = 0,
+        component = noop;
+
+    var dataJoin = dataJoinUtil()
+        .selector('g.container')
+        .element('g')
+        .attr({'class': 'container', 'layout-style': 'flex: 1'});
+
+    var container = function(selection) {
```
chrisprice: Any reason not to permit decoration?

src/tool/container.js
```diff
@@ -0,0 +1,57 @@
...
+
+export default function() {
+
+    var padding = 0,
+        component = noop;
+
+    var dataJoin = dataJoinUtil()
+        .selector('g.container')
+        .element('g')
+        .attr({'class': 'container', 'layout-style': 'flex: 1'});
```
chrisprice: Split across lines?

src/tool/container.js
```diff
@@ -0,0 +1,57 @@
...
+
+    var container = function(selection) {
+        selection.each(function(data, index) {
+            var expandedPadding = expandMargin(padding);
+
+            var g = dataJoin(this, [data]);
+
+            g.enter().append('rect')
```
chrisprice: Split across lines?

src/tool/container.js
```diff
@@ -0,0 +1,57 @@
...
+                .layout('flex', 1);
+            g.enter().append('g')
```
chrisprice: Split across lines?

src/tool/tool.js
```diff
@@ -1,9 +1,11 @@
 import crosshair from './crosshair';
 import fibonacciFan from './fibonacciFan';
 import measure from './measure';
+import container from './container';
 
 export default {
     crosshair: crosshair,
     fibonacciFan: fibonacciFan,
+    container: container,
```
chrisprice: I wouldn't go hunting here for a container, not sure where I would start mind you...

src/layout/strategy/noop.js
```diff
@@ -0,0 +1,29 @@
+import d3 from 'd3';
+
+export default function() {
+
+    var containerWidth = 1,
```
chrisprice: Why does a noop need to care about containerWidth/Height?

chrisprice: For strategies that need it are the scales a more interesting thing to pass in?

src/layout/strategy/strategy.js
```diff
@@ -0,0 +1,5 @@
+import withinContainer from './withinContainer';
+
+export default {
+    withinContainer: withinContainer
```
chrisprice: No noop strategy.

src/layout/strategy/withinContainer.js
```diff
@@ -0,0 +1,41 @@
+import d3 from 'd3';
```
chrisprice: Rename to `within`?

src/layout/strategy/withinContainer.js
```diff
@@ -0,0 +1,41 @@
...
+import {rebindAll} from '../../util/rebind';
+
+export default function() {
+
+    var containerWidth = 1,
```
chrisprice: Same comment as `noop`. Could just be called `width`/`height`?

src/tool/crosshair.js
```diff
@@ -18,8 +18,8 @@ export default function() {
         },
         decorate = noop;
 
-    var x = function(d) { return d.xInDomainUnits ? xScale(d.x) : d.x; },
```
chrisprice: Do these changes need to be made to the other tools? e.g. https://github.com/ScottLogic/d3fc/blob/master/src/tool/measure.js#L19

chrisprice:
A suitably large number of comments for a suitably large PR but I don't think many of them will be controversial. Also worth a squash or maybe splitting up the commits if you want to leave some comments as issues.

src/layout/flexbox.js
ColinEberhardt: I'm not sure it should - if you invoke layout on an SVG element, wouldn't you expect it's width and height to be set as a result? In the instance where the SVG does not have layout-width or layout-height set and its dimensions are measured, the width and height applied would have no effect. However if it did have layout-width or layout-height attributes, these values would then be turned into width / height which makes sense to me.

src/chart/tooltip.js
ColinEberhardt: I checked, and yes it can, but I don't think it makes the code any clearer.

src/chart/tooltip.js
ColinEberhardt: split sounds good to me - done.

src/chart/tooltip.js
ColinEberhardt: The problem with that interface is I think it becomes quite unreadable in practice....

```
 tooltip()
    .labels(['high', 'low', 'open', 'close'])
    .values([
       function(d) { return d.high; },
       function(d) { return d.low; },
       function(d) { return d.high; },
       function(d) { return d.cose; },
    ])
```

It's hard to see the correlation between labels and values.

I also like the brevity of the current interface, it's similar to the way D3 expresses ranges

src/layout/rectangles.js
ColinEberhardt: Makes sense, required a bit of functor juggling but I like the slightly more compact interface.

src/layout/rectangles.js
ColinEberhardt: No - I'll raise an issue regarding this point. Currently `layout(width, height)` sets the width and height, but also invokes layout on the selection too. In this instance I just want to set the width / height attributes.

src/layout/rectangles.js
ColinEberhardt: Done

src/layout/rectangles.js
ColinEberhardt: I'm not really sure we need one - the `rectangles` component only has 4 properties, so the chance of collision is pretty small. I'm also not sure what a suitable prefix would be.

src/tool/container.js
ColinEberhardt: done

src/tool/container.js
ColinEberhardt: Yeah, you're right, it doesn't add anything. Removed

src/tool/container.js
ColinEberhardt: nope - just forgot. Added.

src/layout/flexbox.js
chrisprice: I was just thinking that the logic to apply to an `svg`/`rect` should be the same as the others. Maybe that just means we should exclude `x` and `y`?

src/layout/strategy/strategy.js
ColinEberhardt: Added - and removed containerWidth / height from noop.

src/layout/strategy/noop.js
ColinEberhardt: I'm not so sure about supplying scales to strategies. I think of them as more akin to our indicator calculators. They operate on the screen coordinate system.

src/layout/strategy/strategy.js
chrisprice: Does that mean noop is now equivalent to fc.util.fn.identity?

src/layout/strategy/withinContainer.js
ColinEberhardt: Not keen, within what?

chrisprice: `boundingBox`?

src/layout/strategy/withinContainer.js
ColinEberhardt: With a couple of couple of different types of rectangle in action (container and children) I  think I'd prefer to be explicit.

src/tool/crosshair.js
ColinEberhardt: Well spotted, done.

src/layout/flexbox.js
ColinEberhardt: Yep, excluding x / y makes sense to me.

src/layout/strategy/withinContainer.js
ColinEberhardt: That's a better one.

ColinEberhardt:
phew - all done. I've re-based and squashed all the pre-review changes into a single commit. post-review updates are one-per-commit. I'll squash once we're done with the updates.

src/layout/rectangles.js
chrisprice: Agreed.

chrisprice:
:+1:

ColinEberhardt:
thanks for the review comments ...

--- insert celebratory gif here ---

PR #778, opened 2015-11-13
Largest Triangle 1 Sampler

WPFerg:
Creates a third sampler, largest triangle 1, which compares each data point to its two neighbours, picking the largest effective area per bucket.

### DIFF
--- a/site/src/components/sampler/largestTriangleOne.md
+++ b/site/src/components/sampler/largestTriangleOne.md
@@ -11,8 +11,8 @@ example-code: |
     // configure the sampler
     var sampler = fc.data.sampler.largestTriangleOneBucket()
         .bucketSize(20)
-        .xValue(function(d) { return d.date; })
-        .yValue(function(d) { return d.low; });
+        .x(function(d) { return d.date; })
+        .y(function(d) { return d.low; });
 
     // sample the data
     var sampledData = sampler(data);
@@ -46,9 +46,16 @@ example-code: |
 
 The Largest Triangle One Bucket component subsamples data by calculating the largest areas between a data point and its immediate neighbours.
 
-The sampler requires both the `xValue` and `yValue` properties in order to calculate the area. You can configure the sampling frequency by setting the `bucketSize` property.
+The sampler requires both the `x` and `y` properties in order to calculate the area. You can configure the sampling frequency by setting the `bucketSize` property. If you're dealing with irregularly spaced data, you can call the respective scales on the accessor functions to ensure the area calculated is more representative of the data being displayed, for example:
 
-The example below creates an array of 1,000 datapoints, sampling it using a bucket-size of 20:
+```js
+    var sampler = fc.data.sampler.largestTriangleOneBucket()
+        .bucketSize(20)
+        .x(function(d) { return xScale(d.date); })
+        .y(function(d) { return yScale(d.low); });
+```
+
+The example below creates an array of 1,000 datapoints, sampling it using a bucket-size of 20. For simplicity of the example, the scale functions aren't called when accessing `x` or `y`.
 
 ```js
 {{{example-code}}}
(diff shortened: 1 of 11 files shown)


site/src/components/sampler/largestTriangleOne.md
```diff
@@ -0,0 +1,59 @@
...
+        });
+
+    container.append('g')
+        .datum(data)
+        .call(originalLine);
+---
+
+The Largest Triangle One Bucket component subsamples data by calculating the largest areas between a data point and its immediate neighbours.
+
+The sampler requires both the `xValue` and `yValue` properties in order to generate calculate the area. You can configure the sampling frequency by setting the `bucketSize` property.
```
chrisprice: "generate calculate"

chrisprice: If you're calculating area do you not need to use the scaled values? There's no guarantee that the values themselves have any linear relationship with the scaled values.

WPFerg: Whoops, I'll make that typo change soon. For the use of scale, I hadn't considered that. The three bucket doesn't have it in so it'd probably be good to make any prospective change to that too. I don't mind having it in or not; it's a question of whether or not that could overcomplicate the example.

chrisprice: Thanks. I hadn't considered that it would complicate the example. Maybe it should be added to the documentation and an explanation of why you didn't need it?

Also can we change the properties to be called e.g. `x` rather than `xValue` to indicate they should be scaled where appropriate. I think we already do this for some components like the svg path generators.

WPFerg:
I've pushed the discussed changes. If these are okay I'll make the same changes to the three-bucket variant, too.

chrisprice:
:+1:

WPFerg:
I've pushed the updated three-bucket. Should the respective change be made to mode-median as well for similarity? If so, should the rename be from `value` to `y`?

ColinEberhardt:
> I've pushed the updated three-bucket.

Thanks for the updates - looks good.

>  Should the respective change be made to mode-median as well for similarity? If so, should the rename be from value to y?

Good question! I am wondering whether a non-linear scale would actually effect the output of the sampler? For example, if you were using a log scale, would this effect the outcome of the mode-median sampler?

site/src/components/sampler/largestTriangleThreeBucket.md
```diff
@@ -47,7 +47,14 @@ example-code: |
 
 The Largest Triangle Three Bucket sampling component is a method of subsampling data to improve performance with large data sets. The algorithm entails calculating the largest areas between a data point and its neighbouring buckets.
 
-The sampler requires both the `xValue` and `yValue` properties in order to calculate the area. You can configure the sampling frequency by setting the `bucketSize` property.
+The sampler requires both the `x` and `y` properties in order to calculate the area. You can configure the sampling frequency by setting the `bucketSize` property. If you're dealing with irregularly spaced data, you can call the respective scales on the accessor functions to ensure the area calculated is more representative of the data being displayed, for example:
```
ColinEberhardt: > If you're dealing with irregularly spaced data, you can call the respective scales on the accessor functions to ensure the area calculated is more representative of the data being displayed, for example:

This doesn't sound right to me. I thought the issue was that the algorithm assumed that the x and y scales are continuous and linear? I'd suggest the following:

"As the  The Largest Triangle Three Bucket sampler computes areas, if you have a non-linear or discontinuous scale, the values supplied to the sampler must be scaled accordingly. You can apply the respective scales in the accessor functions as follows:"

WPFerg:
For mode median, I don't think so, since its only influence is the y-axis read off. Log/linear axes on the y axis shouldn't have an impact.

site/src/components/sampler/largestTriangleThreeBucket.md
WPFerg: Thanks -- I thought my wording sounded a little off, I'll switch it out

WPFerg:
I've updated the wording for the website

ColinEberhardt:
Thanks for the updates @WPFerg - I did one final tweak, see #780 - the re-computation of the `[xValue, yValue]` array was bugging me ;-)

PR #781, opened 2015-11-17
Updated text for samplers

WPFerg:
Updated the component documentation to add some of the wording in the [blog post](http://blog.scottlogic.com/2015/11/16/sampling-large-data-in-d3fc.html)


ColinEberhardt:
Thanks for the update - can you please do one last thing? Can you link to that guys thesis which you obtained the algorithms from? I think it would be a useful reference for anyone wishing to find out more.

WPFerg:
Added them to the one-bucket and mode median. I left it out on the three-bucket as I already had the link in for image attributiono

ColinEberhardt:
Perfect - really appreciate the update.

PR #782, opened 2015-11-17
Changed z-order of chart axis labels

### LINKED ISSUE #766: Cartesian chart axis labels are ordered before plot area
The [axis labels in cartesian chart](https://github.com/ScottLogic/d3fc/blob/master/src/chart/cartesian.js#L74-L83) are ordered before the plot area. If the labels were positioned to be in the same space as the plot area, they would be obscured by the background rect element (and any other element in the plot area).

I can't see a case where the labels should be under the plot area. I'd suggest re-ordering the element, so the labels appear above the plot area.

ColinEberhardt:
Fixes #766 

Note, I discovered that the selectors which applied the required svg layout attributes to the label containers only found the correct elements because they were the first ones with the class of `x-axis` or `y-axis`. I've added a second class to disambiguate.

@jleft does this fix the issue for you?

### DIFF
--- a/src/chart/cartesian.js
+++ b/src/chart/cartesian.js
@@ -66,22 +66,7 @@ export default function(xScale, yScale) {
 
             var svg = containerDataJoin(container, [data]);
             svg.enter().html(
-                '<g class="title"> \
-                    <g layout-style="height: 0; width: 0"> \
-                        <text class="label"/> \
-                    </g> \
-                </g> \
-                <g class="y-axis"> \
-                    <g layout-style="height: 0; width: 0"> \
-                        <text class="label"/> \
-                    </g> \
-                </g> \
-                <g class="x-axis"> \
-                    <g layout-style="height: 0; width: 0"> \
-                        <text class="label"/> \
-                    </g> \
-                </g> \
-                <g class="plot-area-container"> \
+                '<g class="plot-area-container"> \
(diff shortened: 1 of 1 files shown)


jleft:
> @jleft does this fix the issue for you?

Yes, it does, thank you :smiley:

src/chart/cartesian.js
```diff
@@ -66,31 +66,31 @@ export default function(xScale, yScale) {
...
+                    <g class="axes-container" \
+                        layout-style="position: absolute; top: 0; bottom: 0; left: 0; right: 0"> \
+                        <g class="x-axis" layout-style="height: 0; width: 0"/> \
+                        <g class="y-axis" layout-style="height: 0; width: 0"/> \
                     </g> \
+                    <svg class="plot-area" \
+                        layout-style="position: absolute; top: 0; bottom: 0; left: 0; right: 0"/> \
                 </g> \
-                <g class="y-axis"> \
+                <g class="x-axis label"> \
```
jleft: I have previously used `.label` to style the text labels, it's not a big deal if these styles are also applied to the `g`, and I could always update to `text.label`, but maybe `.label-container` would make sense here?

ColinEberhardt: Good call - `label-container` sounds like a better idea

PR #783, opened 2015-11-17
Revert "Fix scale from center transition"

### LINKED ISSUE #774: candlestick / OHLC transition from the right!
Candlesticks enter by 'stretching' in from their right edge. They should expand about their center.

http://jsbin.com/zevagupaho/edit?html,js,output

see #753

ColinEberhardt:
This reverts commit 684d0b0eab884d0dcbe134bf966fa54f1e1f94d8.

Fixes #774

I'm not sure the original fix was correct. When d3 transitions a transform it transitions the canonical translate, rotate, scale and skew. Hence the translation applied via the matrix was being combined with the translation that positions the candlestick, resulting in an odd transition.

If you look at the jsbin:

http://jsbin.com/zevagupaho/edit?html,js,output

then roll back to d3fc 4.3.0 you can see that the transitions were correct.

Were you testing this in a chart where the x-values were also scrolling at the same time?

### DIFF
--- a/src/series/candlestick.js
+++ b/src/series/candlestick.js
@@ -19,12 +19,6 @@ export default function() {
         return 'translate(' + values.x + ', ' + values.yHigh + ')';
     }
 
-    function xScaleFromCenter(width, scale) {
-        var center = width / 2;
-        var offset = center - scale * center;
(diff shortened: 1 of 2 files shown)


chrisprice:
Yep, you're right. Further proof - http://jsbin.com/pumizagifu/1/edit?html,js,output

Issue #543, opened 2015-09-11
Support merging of properties in d3.selection.prototype.layout


chrisprice:
`d3.selection.prototype.layout({})` should apply the key/values in addition to what is already there, replacing key values as appropriate.

Currently the behavior is to replace everything with the specified key/values, including removing non-conflicting values.

This would remove the need to mix constant and variable values being assigned such as in - https://github.com/ScottLogic/d3fc/pull/518/files#diff-54543ef480c3f51e0d0698f36a9ddfd7R93

ColinEberhardt:
With our d3v4 release we will no longer use flexbox layout - see https://github.com/d3fc/d3fc-element

PR #784, opened 2015-11-18
Layout merges properties (#543)

WPFerg:
Allows the merging of properties as outlined in #543


ColinEberhardt:
Nice job, thanks @WPFerg :-)

PR #787, opened 2015-11-20
Ensure label-container class isn't removed when setting orient on Cartesian chart's axis label containers

jleft:
Also fix the axis labels in the Wealth & Health of Nations example.


ColinEberhardt:
Well spotted - I wonder whether this code should use `selection.class` rather than `attr`? I know that there are some performance implications, however at least the use of `class` would be less error prone.

PR #790, opened 2015-11-21
Documentation updates

ColinEberhardt:
That's the problem with adding a bunch of new components - lots more documentation required.

@ScottLogic/d3fc - as this is a documentation PR if you haven't seen the components I recently added, why not read these docs to see if they make sense?


site/src/components/chart/tooltip.md
```diff
@@ -0,0 +1,50 @@
...
+
+  var svg = d3.select('#tooltip')
+    .attr({width: 100, height: 60});
+
+  svg.datum(data)
+    .call(tooltip);
+---
+
+The tooltip component renders the bound data is a table, where the mapping from the data to the tabular form is expressed via the `items` property as demonstrated below:
+
```
owennw: I think there's a typo here.

site/src/components/chart/tooltip.md
```diff
@@ -0,0 +1,50 @@
...
+{{{example-code}}}
+```
+
+Which results in the following:
+
+<svg id="tooltip"> </svg>
+<script type="text/javascript">
+(function() {
+  {{{example-code}}}
+}());
```
owennw: The tabbing here is inconsistent with the legend.md above.

site/src/components/tool/container.md
```diff
@@ -0,0 +1,18 @@
...
+layout: component
+title: Container
+component: tool/container.js
+namespace: tool
+
+example-code: |
+  var tooltipContainer = fc.tool.container()
+        .padding(5)
+        .component(tooltip);
+
```
owennw: The lines here look over-tabbed.

site/src/components/chart/tooltip.md
```diff
@@ -0,0 +1,50 @@
+---
+layout: component
+title: Tooltip
+component: chart/tooltip.js
+namespace: chart
+
+example-code: |
```
owennw: Do we want frontpage or playground tags?

site/src/components/layout/rectangles.md
```diff
@@ -0,0 +1,75 @@
+---
+layout: component
+title: Rectangles
+component: layout/rectangles.js
+namespace: layout
+
+example-code: |
```
owennw: Do we want frontpage or playground tags?

site/src/components/tool/crosshair.md
```diff
@@ -0,0 +1,122 @@
...
+      .position([10, 10])
+      .size([50, 30])
+      .component(tooltip);
+
+  // create an array which will hold the crosshair datapoint
+  var crosshairData = [];
+
+  // use a multi-series to render both the line and crosshair
+  var multi = fc.series.multi()
+    .series([line, crosshair, tooltipLayout])
```
owennw: The comment should include the tooltipLayout too.

site/src/components/tool/crosshair.md
```diff
@@ -0,0 +1,122 @@
...
+```
+
+This renders a tooltip in the top-left corner:
+
+<div id="example-code-2" class="chart"> </div>
+<script type="text/javascript">
+(function() {
+    var desiredWidth = $('#example-code-2').width(),
+        desiredHeight = desiredWidth / 2.4; //keeps the width-height ratio at 600-250 (defaults for createFixture)
+    var f = createFixture('#example-code-2', desiredWidth, desiredHeight, null, function() { return true; });
```
owennw: Do we prefer `d3.functor(true)` for brevity?

site/src/components/tool/crosshair.md
```diff
@@ -0,0 +1,122 @@
...
+    var container = f.container, data = f.data,
+      xScale = f.xScale, yScale = f.yScale;
+    {{{example-code-2}}}
```
owennw: The tabbing is a bit off.

site/src/components/layout/rectangles.md
```diff
@@ -0,0 +1,75 @@
...
+      .size([80, 30])
+      .position([function(d) { return d.x; }, function(d) { return d.y; }])
+      .component(label);
+
+  // bind the data and render
+  svg.datum(data)
+      .call(layout);
+---
+The rectangles layout component provides a mechanism for child components based on rectangular bounding boxes. It is typically used to render tooltips or labels on a chart.
+
```
owennw: I'm not entirely sure what this first sentence means. I was expecting more of a verb after the words 'mechanism for' (i.e. "provides a mechanism for child components, based on rectangular bounding boxes, ... to do something").

site/src/components/tool/crosshair.md
```diff
@@ -0,0 +1,122 @@
...
+```js
+{{{example-code}}}
+```
+
+Which gives the following:
+
+{{>example-fixture}}
+
+You can control the snapping behaviour of the crosshair by supplying a snapping function to the `snap` property. A number of snap functions are [supplied as part of the library]({{package.repository.url}}/blob/master/src/util/snap.js).
+
```
owennw: Is it worth including an example with this snapping behaviour?

owennw:
I've had a read through it all. Minor comments, mostly regarding formatting and the odd typo. I think overall I understand the documentation and the new components, but it might be worth sourcing other opinions too.

site/src/components/chart/tooltip.md
ColinEberhardt: fixed - thanks.

site/src/components/chart/tooltip.md
ColinEberhardt: nah - it is a pretty boring example. Also, the way that playgrounds work at the moment (relying on the test fixture), it's not possible to add. Hopefully this will be resolved soon ...

site/src/components/layout/rectangles.md
ColinEberhardt: ditto

site/src/components/tool/crosshair.md
ColinEberhardt: done

site/src/components/tool/crosshair.md
ColinEberhardt: no, `d3.functor` isn't really about brevity, it is there to allow properties to support functions or constants.

site/src/components/layout/rectangles.md
ColinEberhardt: well spotted ... fixed.

site/src/components/tool/crosshair.md
ColinEberhardt: yes, although I think I'll do this via an example instead ...

ColinEberhardt:
Thanks for the feedback - much appreciated

PR #793, opened 2015-11-25
Greedy strategy, collision utils

WPFerg:
Creates a greedy layout strategy, choosing the best option at the time of label placement, and introduces some collision utilities

A couple of thoughts/things to note:
* I haven't exposed the utils `collisions` and `container` because I can't think of many use cases for them.
* The anchor location (the point intended for the label) is bound to each object in the selection as the field `anchor`, which seemed better than placing an attribute on each `g` element created

### DIFF
--- a/src/layout/rectangles.js
+++ b/src/layout/rectangles.js
@@ -12,6 +12,7 @@ export default function(layoutStrategy) {
 
     var xScale = d3.scale.identity(),
         yScale = d3.scale.identity(),
+        anchor = noop,
         strategy = layoutStrategy || identity,
         component = noop;
 
@@ -56,6 +57,11 @@ export default function(layoutStrategy) {
                 return 'translate(' + offset.x + ', ' + offset.y + ')';
             });
 
+            data.forEach(function(d, i) {
+                var pos = position(d, i);
+                anchor(i, pos[0] - layout[i].x, pos[1] - layout[i].y);
+            });
+
             // set the layout width / height so that children can use SVG layout if required
             g.attr({
                 'layout-width': function(d, i) { return childRects[i].width; },
@@ -81,6 +87,14 @@ export default function(layoutStrategy) {
             return position;
         }
         position = arrayFunctor(x);
+        return rectangles;
+    };
+
+    rectangles.anchor = function(x) {
+        if (!arguments.length) {
+            return anchor;
+        }
+        anchor = x;
         return rectangles;
     };
 
--- a/src/layout/strategy/collision.js
+++ b/src/layout/strategy/collision.js
@@ -1,0 +1,39 @@
+import d3 from 'd3';
+
+function isIntersecting(a, b) {
+    return !(a.x >= (b.x + b.width) ||
+        (a.x + a.width) <= b.x ||
+        a.y >= (b.y + b.height) ||
+        (a.y + a.height) <= b.y);
+}
+
+function areaOfIntersection(a, b) {
+    var left = Math.max(a.x, b.x);
+    var right = Math.min(a.x + a.width, b.x + b.width);
+    var top = Math.max(a.y, b.y);
+    var bottom = Math.min(a.y + a.height, b.y + b.height);
+    return (right - left) * (bottom - top);
+}
+
+function collideAll(data) {
+    var collisions = [];
+    for (var i = 0; i < data.length; i++) {
+        collisions = collisions.concat(collidePoint(data, i));
+    }
+    return collisions;
+}
+
+function collidePoint(data, pointIndex) {
+    var pointA = data[pointIndex];
+    return data.filter(isIntersecting.bind(null, pointA)).map(function(pointB) {
+        return [pointA, pointB];
+    });
+}
+
+export function areaOfCollisions(data) {
+    var collisionData = collideAll(data);
+    var eachCollisionArea = collisionData.map(function(collision) {
+        return areaOfIntersection(collision[0], collision[1]);
+    });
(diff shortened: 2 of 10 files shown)


ColinEberhardt:
Thanks - this is looking really good. I'll start having a look at the implementation in detail shortly. From a bit of testing I am not quite getting the results I'd expect.

Given the following data:

```
    var data = [
        { x: 150, y: 100, data: 'hi'},
        { x: 160, y: 100, data: 'hello'}
    ];
```

And using the greedy algorithm, the rectangles don't move:

screen shot 2015-11-26 at 09 08 32

In this case I'd expect that the algorithm could arrange them so that they are not overlapping. Is this the result that you expect?

WPFerg:
Thanks, I'd forgotten some equals out of the intersection logic. It used to only score the greedy algorithm based off of the number of intersections, but using the area is a better measure so I've changed it appropriately.

src/layout/rectangles.js
```diff
@@ -56,6 +56,14 @@ export default function(layoutStrategy) {
                 return 'translate(' + offset.x + ', ' + offset.y + ')';
             });
 
+            g.each(function(d, i) {
```
ColinEberhardt: as discussed - I'd like to see whether the anchor can be written back to the bound data, rather than being associated with the DOM.

src/layout/strategy/collision.js
```diff
@@ -0,0 +1,88 @@
+export default function() {
+
+    var data = [];
+    var collision = function(inputData) {
```
ColinEberhardt: I don't think this really works as a component. The typical pattern is a [fluent builder pattern](http://blog.crisp.se/2013/10/09/perlundholm/another-builder-pattern-for-java) which constructs a function which is then either invoked with a selection or some data.

In this case you are returning something more complex that has various methods on it. 

I think this would be better as a simple set of utility functions - there is very little shared between utility method.

src/layout/strategy/collision.js
```diff
@@ -0,0 +1,88 @@
...
+        data = inputData;
+
+        return collision;
+    };
+
+    collision.getAllCollisions = function(index) {
```
ColinEberhardt: is this used anywhere?

src/layout/strategy/collision.js
```diff
@@ -0,0 +1,88 @@
...
+            (a.y + a.height) < b.y);
+    }
+
+    function collideAll(callback) {
+        for (var i = 0; i < data.length; i++) {
+            collidePoint(i, callback);
+        }
+    }
+
+    function collidePoint(pointIndex, callback) {
```
ColinEberhardt: Rather than having a 'visitor' callback, how about a more functional approach ... where as an input you have an array of data, then use Array.filter to output the points that collide?

src/layout/strategy/container.js
```diff
@@ -0,0 +1,29 @@
+export default function() {
+    var containerWidth = 0,
+        containerHeight = 0;
+
+    var container = function(point) {
+        return !(point.x < 0 || point.y < 0 ||
```
ColinEberhardt: it looks like this strategy returns booleans - I thought the others returned bounding boxes?

src/layout/strategy/greedy.js
```diff
@@ -0,0 +1,80 @@
...
+
+export default function() {
+
+    var container = containerUtils();
+    var collider = collision();
+
+    var strategy = function(data) {
+        var builtPoints = [];
+
+        // Use forEach over map so the array can be built one at a time.
```
ColinEberhardt: not sure what this comment means?

src/layout/strategy/greedy.js
```diff
@@ -0,0 +1,80 @@
...
+        data.forEach(function(point) {
+            var allPointPlacements = getAllPlacements(point);
+            var scoredPlacements = allPointPlacements.map(function(unscoredPlacement) {
+                return checkPlacement(builtPoints, unscoredPlacement);
+            });
+
+            var bestScore = Infinity;
```
ColinEberhardt: We have similar logic implemented using map-reduce elsewhere:

https://github.com/ScottLogic/d3fc/blob/master/src/util/snap.js#L19

(which is just showing off ;-) )

I think this is a good candidate for a utility function, that takes an array of items and a scoring function.

visual-tests/layout/rectangles.js
```diff
@@ -149,13 +166,42 @@
     var yScale = d3.scale.linear()
         .range([height, 0]);
 
-    var smart = fc.layout.rectangles(strategy())
-        .xScale(xScale)
-        .yScale(yScale)
-        .size([itemWidth, itemHeight])
-        .component(label);
+    function useStrategy(strategyToUse) {
```
ColinEberhardt: I like the update to the visual tests. Can you please delete all the un-used junk at the top of this file too?

WPFerg:
I've done the refactor, the big changes:

* Collision.js is now a helper function collection
* The way collision.js does collisions is more functional (using filter, map, etc) and outputs a collision list over a collision map.
* Retrieving anchor points is done using `.anchor(function callback (index, x, y) {})`, defaulting with noop,
* Added `getBestPlacement(placements, scorer)`, which returns the best placement using the given scorer

WPFerg:
Pushed up a further refactor, adding `fc.util.minimum` and squashing the commits :)

ColinEberhardt:
Awesome - great job @WPFerg - and thanks for putting up with all the changes I requested!

PR #796, opened 2015-11-26
Implements symmetricalAbout for dates

djmiley:
Nothing very clever, just a reimplementation of what is done for the padding.

### DIFF
--- a/src/util/extent.js
+++ b/src/util/extent.js
@@ -55,12 +55,25 @@ export default function() {
         var max = dataMax;
 
         if (symmetricalAbout != null) {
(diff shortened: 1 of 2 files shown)


ColinEberhardt:
A useful addition - thanks.

chrisprice:
FYI I think this could be simplified a little bearing in mind the behaviour of [value of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/valueOf)

ColinEberhardt:
@chrisprice good call, actually pretty much all of the date-specific logic can be removed, see https://github.com/ScottLogic/d3fc/pull/797

PR #797, opened 2015-11-26
Simplified extent

ColinEberhardt:
Simplified extent by converting from Date to number at the start of the 'extent' logic, then converting back at the end.

### DIFF
--- a/src/util/extent.js
+++ b/src/util/extent.js
@@ -51,48 +51,26 @@ export default function() {
             });
         });
 
-        var min = dataMin;
-        var max = dataMax;
+        var dateExtent = Object.prototype.toString.call(dataMin) === '[object Date]';
 
+        var min = dateExtent ? dataMin.getTime() : dataMin;
+        var max = dateExtent ? dataMax.getTime() : dataMax;
+
+        // apply symmetry rules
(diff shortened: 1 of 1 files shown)


chrisprice:
My expectation was that there would be no need to treat dates any
differently apart from wrapping them at the end i.e. no getTime calls. Did
that not work?

src/util/extent.js
```diff
@@ -51,49 +51,27 @@ export default function() {
...
+            var symmertical = dateExtent ? symmetricalAbout.getTime() : symmetricalAbout;
+            var distanceFromMax = Math.abs(max - symmertical),
+                distanceFromMin = Math.abs(min - symmertical),
+                halfRange = Math.max(distanceFromMax, distanceFromMin);
 
-            min = new Date(oldMin - delta);
-            max = new Date(oldMax + delta);
+            min = symmertical - halfRange;
+            max = symmertical + halfRange;
         }
```
owennw: There's a typo in the variable name `symmertical`.

ColinEberhardt: well spotted - thanks! fixed'n'squashed

ColinEberhardt:
> My expectation was that there would

No, because ... WAT?!

PR #798, opened 2015-11-27
Local-search label placement algorithm

WPFerg:
Uses a local search approach to place labels -- directly trying to resolve any possible collisions. It automatically terminates if it reaches an area of local minimum.

I've also edited the `minimum` util to allow parameters to be passed in to the scoring function (saves having to bind variables each time), but it can be switched back with no real differences

### DIFF
--- a/src/layout/strategy/collision.js
+++ b/src/layout/strategy/collision.js
@@ -1,43 +1,18 @@
 import d3 from 'd3';
 
-function isIntersecting(a, b) {
-    return areaOfIntersection(a, b) > 0;
+export function isIntersecting(a, b) {
+    return !(a.x >= (b.x + b.width) ||
+       (a.x + a.width) <= b.x ||
+       a.y >= (b.y + b.height) ||
+       (a.y + a.height) <= b.y);
 }
 
-function areaOfIntersection(a, b) {
+export function areaOfIntersection(a, b) {
     var left = Math.max(a.x, b.x);
     var right = Math.min(a.x + a.width, b.x + b.width);
     var top = Math.max(a.y, b.y);
     var bottom = Math.min(a.y + a.height, b.y + b.height);
     return (right - left) * (bottom - top);
-}
-
-function collideAll(data) {
-    var collisions = [];
-    for (var i = 0; i < data.length; i++) {
-        collisions = collisions.concat(collidePoint(data, i));
-    }
-    return collisions;
-}
-
-function collidePoint(data, pointIndex) {
-    var pointA = data[pointIndex];
-    return data.filter(isIntersecting.bind(null, pointA)).map(function(pointB) {
-        return [pointA, pointB];
-    });
-}
-
-export function allCollisionIndices(data) {
-    var collisions = collideAll(data);
-
-    // Convert [[a, b], [a, c]] => [a, b, c]
-    var uniqueCollisions = collisions.map(function(d) {
-        return data.indexOf(d[0]);
-    }).filter(function(d, i, arr) {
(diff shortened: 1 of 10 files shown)


WPFerg:
Pushed up a slight refactor, bringing shared functions into their own utility file, making the scorer function a factory (code readability), and making the algorithm more stateless.

src/layout/strategy/collision.js
```diff
@@ -30,8 +30,36 @@ function collidePoint(data, pointIndex) {
...
+        if (uniqueCollisions.indexOf(pointA) < 0) {
+            uniqueCollisions.push(pointA);
+        }
+
+        if (uniqueCollisions.indexOf(pointB) < 0) {
+            uniqueCollisions.push(pointB);
+        }
+    });
+
+    return uniqueCollisions;
```
ColinEberhardt: Isn't there a simpler way to compute this? Isn't it just a filter over the data where you check for any intesections?

src/layout/strategy/local.js
```diff
@@ -0,0 +1,83 @@
...
+        var originalData = data;
+        var iteratedData = data;
+
+        var lastIterationScore = Infinity;
+        for (var i = 0; i < iterations; i++) {
+            iteratedData = iterate(originalData, iteratedData);
+
+            // Reached a local minimum?
+            var thisIterationScore = areaOfCollisions(iteratedData);
+            if (lastIterationScore === thisIterationScore) {
```
ColinEberhardt: Where if `lastIterationScore` updated in this loop?

src/layout/strategy/local.js
```diff
@@ -0,0 +1,83 @@
...
+                return iteratedData;
```
ColinEberhardt: Rather than an early-return, I think this would be more clearly expressed as a while loop, i.e. repeat while i < iterations AND score < previousScore

src/layout/strategy/local.js
```diff
@@ -0,0 +1,83 @@
...
+    };
+
+    function iterate(originalData, iteratedData) {
+        var collidingPoints = allWithCollisions(iteratedData);
+        var totalNoOfCollisions = areaOfCollisions(iteratedData);
+
+        // Try to resolve collisions from each node which has a collision
+        for (var i = 0; i < collidingPoints.length; i++) {
+
+            var pointIndex = collidingPoints[i];
```
ColinEberhardt: Perhaps use Array.forEach to avoid the need to index into the array?

ColinEberhardt:
Just a few comments, I'm still a little unsure about the construction of the algorithm, I'll pop over and have a chat with you later.

ColinEberhardt:
@WPFerg - I've had a go at modifying the implementation. It's not 100% complete, but hopefully you get the idea:

https://github.com/ColinEberhardt/d3-financial-components/blob/WPFerg-labels-local/src/layout/strategy/local.js

This removes the need for the various utility methods within `collision.js` which I don't think were adding value, their outputs were a little confusing. I think you could also make more use of higher order functions, filter, map, reduce, forEach etc ...

I'd be happy to talk this over with you this morning.

WPFerg:
Thanks -- I've done the refactors you suggested yesterday and changed it to a more functional approach.

I've left the area calculations in `collision.js`, and refactored greedy accordingly (I've changed the tests due to the area of collisions calculation no longer doing self-collisions, which caused a change in output here due to the rectangles not all being the same size).

ColinEberhardt:
Many thanks for the updates - I've rebased and squashed, will merge shortly when Travis is happy

PR #741, opened 2015-11-05
Add transition support to ohlc/candlestick

chrisprice:
Closes #740 and contributes to with #212

Issue #480, opened 2015-07-07
Add path data generator component


jleft:
D3 has some [great SVG path data generators](https://github.com/mbostock/d3/wiki/SVG-Shapes#path-data-generators). However, sometimes it makes sense to use the path data instructions directly, but constructing this can often involve some fairly ugly string concatenation; [OHLC, for example](https://github.com/ScottLogic/d3fc/blob/master/src/svg/ohlc.js#L26-30):

```js
var moveToLow = 'M' + xValue + ',' + yLow,
    verticalToHigh = 'V' + yHigh,
    openTick = 'M' + xValue + ',' + yOpen + 'h' + (-halfWidth),
    closeTick = 'M' + xValue + ',' + yClose + 'h' + halfWidth;
return moveToLow + verticalToHigh + openTick + closeTick;
```

Creating a component which handles this in a friendlier manner could be of benefit. I was thinking of something like this:

```js
var path = fc.svg.path()
    .moveTo(xValue, yLow)
    .verticalLineTo(yHigh)
    .moveTo(xValue, yOpen)
    .horizontalLineToRelative(-halfWidth)
    .moveTo(xValue, yClose)
    .horizontalLineToRelative(halfWidth);
return path();
```

I've written some tests for this and will finish the implementation, assuming we think this is a worthwhile addition? Any ideas or suggestions welcome!

Credit to @chrisprice for giving me the idea a while ago :smiley:

ColinEberhardt:
Nice idea, reminds me of this:

https://github.com/VerbalExpressions/CSharpVerbalExpressions/

bjedrzejewski:
@jleft are you still working on it? Sounds like a great idea and we have some time in London right now- let me know and maybe I can help.

jleft:
@bjedrzejewski - I haven't worked on this for a while, so please feel free to pick it up :smiley: I'll see if I've got anything worth pushing up (might be useful as a starting point), but won't be able to do so until this evening.

bjedrzejewski:
@jleft ok, no worries. I will see if you find something that you have worked on and pick it up from there tomorrow/once you pushed.

chrisprice:
d3-path beat us to it.

PR #800, opened 2015-11-30
Boxplot Closes #701

### LINKED ISSUE #701: Adding Box Plot series
This is a series popular among statisticians and different scientists:

https://en.wikipedia.org/wiki/Box_plot

I believe that we should support it (as we support candlesticks and other specialized series).

bjedrzejewski:
This is the boxplot pull request. Please review. There is no visual test written yet.


bjedrzejewski:
image

site/src/components/series/boxPlot.md
```diff
@@ -0,0 +1,46 @@
...
+      .datum(data)
+      .call(boxPlot);
+
+---
+
+A [box plot series](https://en.wikipedia.org/wiki/Box_plot) is a convenient way of graphically depicting groups of 
+numerical data through their quartiles. Boxes can be renderer vertically or horizontally based on the value of the `orient` property.
+
+The upper and lower end of each box are defined by the `boxUpper` and `boxLower` properties.
+The upper and lower whisker of each box are defined by the `whiskerUpper` and `whiskerLower` properties.
```
ColinEberhardt: It's probably worth noting that when the bar is vertical the box / whisker properties are in the yScale coordinate system, whereas when it is horizontal, they are in the xScale coordinate system.

src/series/boxBase.js
```diff
@@ -0,0 +1,129 @@
...
+                y: yScale(yValue(d, i)),
+                boxHigh: xScale(boxHigh(d, i)),
+                boxLow: xScale(boxLow(d, i)),
+                whiskerHigh: xScale(whiskerHigh(d, i)),
+                whiskerLow: xScale(whiskerLow(d, i))
+            };
+        }
+    };
+    base.defined = function(d, i) {
+        return whiskerLow(d, i) != null && whiskerHigh(d, i) != null
```
ColinEberhardt: How about adding a utility function that makes this a bit easier to manage?

```
allPropertiesDefined[whiskerHigh, whiskerLow, boxHigh, boxLow, xValue, yValue)
```

src/series/boxPlot.js
```diff
@@ -0,0 +1,86 @@
...
+import boxBase from './boxBase';
+
+export default function() {
+
+    var decorate = noop,
+        barWidth = 5,
+        orient = 'vertical',
+        base = boxBase();
+
+    var dataJoin = dataJoinUtil()
```
ColinEberhardt: @jdunkerley - just wondering if you made any more progress on your datajoin utilities? I'd love to see this repeated pattern eliminated!

src/series/boxPlot.js
```diff
@@ -0,0 +1,86 @@
...
+                    .y(values.y)
+                    .x(values.x)
+                    .boxHigh(values.boxHigh)
+                    .boxLow(values.boxLow)
+                    .whiskerHigh(values.whiskerHigh)
+                    .whiskerLow(values.whiskerLow);
+
+                gboxPlot.select('path')
+                    .attr('d', pathGenerator([d]))
+                    .attr('stroke', 'black');
```
ColinEberhardt: This should be a class

src/svg/boxPlot.js
```diff
@@ -0,0 +1,151 @@
...
+                whiskerToBotBox = boxLow(d, i) - whiskerLow(d, i),
+                boxToMid = y(d, i) - boxLow(d, i),
+                midToBox = boxHigh(d, i) - y(d, i),
+                boxToWhiskerHigh = whiskerHigh(d, i) - boxHigh(d, i),
+                yBottom = y(d, i) - whiskerLow(d, i),
+                yTop = whiskerHigh(d, i) - y(d, i),
+                xBottom = x(d, i) - whiskerLow(d, i),
+                xTop = whiskerHigh(d, i) - x(d, i);
+
+            var boxSerieVertical = '';
```
ColinEberhardt: [nitpick] I think you're missing an 's' in the variable name.

src/svg/boxPlot.js
```diff
@@ -0,0 +1,151 @@
...
+
+                var verticalBar = 'v' + (-halfWidth) + 'v' + (2 * halfWidth) + 'v' + (-halfWidth),
+                    horizontalToBotBox = 'h' + (-whiskerToBotBox),
+                    horizontalToMidBox = 'h' + (-boxToMid),
+                    horizontalToMidBoxDown = 'h' + (boxToMid),
+                    horizontalToTopBox = 'h' + (-midToBox),
+                    horizontalToTopBoxDown = 'h' + (midToBox),
+                    horizontalToWhisker = 'h' + (-boxToWhiskerHigh);
+
+                boxSerieHorizontal = 'M' + xBottom + ',0'
```
ColinEberhardt: A path mini-language would help a lot here https://github.com/ScottLogic/d3fc/issues/480

ColinEberhardt:
Just a few minor comments, but other than that, looks really good to me. 

One thing that probably requires a little thought is transitions, when a box-plot enters it should probably transition in the same way as OHLC / candlestick, i.e. grow from its centre-point. See #741

bjedrzejewski:
Ok, will fix the code at some point this week! Thank you for review and comments.

bjedrzejewski:
Hi @ColinEberhardt, sorry that it took so long, but I think I applied all your suggestions and this should be ready for merge/another review.

src/series/boxBase.js
```diff
@@ -0,0 +1,128 @@
...
+                y: yScale(yValue(d, i)),
+                boxHigh: xScale(boxHigh(d, i)),
+                boxLow: xScale(boxLow(d, i)),
+                whiskerHigh: xScale(whiskerHigh(d, i)),
+                whiskerLow: xScale(whiskerLow(d, i))
+            };
+        }
+    };
+    base.defined = function(d, i) {
+        return defined(d, i, whiskerLow, whiskerHigh, boxLow, boxHigh, xValue, yValue);
```
ColinEberhardt: The purpose of the `defined` function is to determine whether the value accessors return suitable values for the given `d` and `i` arguments. I wouldn't expect `defined` to include the `d` and `i` arguments in its list.

See the similar implementation here:


https://github.com/ScottLogic/d3fc/blob/master/src/series/errorBase.js#L45

src/svg/boxPlot.js
```diff
@@ -0,0 +1,151 @@
...
+                whiskerToBotBox = boxLow(d, i) - whiskerLow(d, i),
+                boxToMid = y(d, i) - boxLow(d, i),
+                midToBox = boxHigh(d, i) - y(d, i),
+                boxToWhiskerHigh = whiskerHigh(d, i) - boxHigh(d, i),
+                yBottom = y(d, i) - whiskerLow(d, i),
+                yTop = whiskerHigh(d, i) - y(d, i),
+                xBottom = x(d, i) - whiskerLow(d, i),
+                xTop = whiskerHigh(d, i) - x(d, i);
+
+            var boxSeriesVertical = '';
```
ColinEberhardt: considering the if / else block below - is there any reason why you have separate `boxSeriesVertical` and `boxSeriesHorizontal` variables? wouldn't a single `boxSeriesPath` variable be sufficient?

src/util/defined.js
```diff
@@ -0,0 +1,11 @@
+// Checks that passes properties are 'defined', meaning that calling them with d, i returns non null values
+export default function(d, i) {
+    for (var c = 0, j = arguments.length; c < j; c++) {
+        if (arguments[c] !== d && arguments[c] !== i) {
```
ColinEberhardt: Ah ... I can see why your code in `boxBase` passed `d` and `i`!

What this function should do is take a list of functions, returning a new function which you invoke like so:

```
function upper(d,i) { ... };
function lower(d,i) { ... };
function center(d,i) { ... };

defined(upper, lower, center)(d, i);
```

Does that make sense?

Considering the usage above, I'd move this to the `fc.util.fn` namespace.

ColinEberhardt:
Thanks for the updates @bjedrzejewski - almost there! Just a few minor tweaks required.

One last thing, can you also roll-out usage of the new `defined` utility function in order to keep the series implementations consistent? Thanks.

src/util/defined.js
bjedrzejewski: It makes perfect sense, I will refactor, move it to fc.util.fn and roll-out the usage.

bjedrzejewski: I am trying to do this:

`// Returns function that checks if passes properties are 'defined', meaning that calling them with (d, i) returns non null values
export function defined() {

    var self = this;

    return function(d, i) {
        for (var c = 0, j = self.arguments.length; c < j; c++) {
            if (self.arguments[c](d, i) === null) {
                return false;
            }
        }
    };
}`

But eslint stops me from `var self = this;` any suggestions?

bjedrzejewski: Nevermind:

var givenArguments = this.arguments;

sorry :)

bjedrzejewski:
@ColinEberhardt there is a major issue with using defined in that way:

`base.defined = defined(errorLow, errorHigh, xValue, yValue);`

Every time any of the `errorLow, errorHigh, xValue, yValue` are being redefined, the `base.defined = defined(errorLow, errorHigh, xValue, yValue);` should be called again. For example from the:

<code>
base.errorLow = function(x) {
        if (!arguments.length) {
            return errorLow;
        }
        errorLow = d3.functor(x);
        base.defined = defined(errorLow, errorHigh, xValue, yValue);
        return base;
    };
</code>

Can you see a better solution?

ColinEberhardt:
Very good point ...

> Maybe we should make it another issue / PR?

Yes, totally agree, pull those changes out of this PR so that it can get merged. Thanks :-)

bjedrzejewski:
It works if we use it like this:

`base.defined = function(d, i) {
        return defined(d, i, xValue, yOpenValue, yLowValue, yHighValue, yCloseValue);
    };`


I'm happy to commit that.

ColinEberhardt:
> It works if we use it like this:

:+1: from me

bjedrzejewski:
@ColinEberhardt ready to merge then!

ColinEberhardt:
> ready to merge then!

Sorry - not quite ... see the comment above.

bjedrzejewski:
@ColinEberhardt I updated to the style that we agreed. Sorry I have got confused in the discussion a bit.

ColinEberhardt:
no worries - still one final tweak needed ;-)

bjedrzejewski:
I removed that check :)

ColinEberhardt:
I've squashed the commits and added a visual test ... see the branch here:

https://github.com/ColinEberhardt/d3-financial-components/commits/bjedrzejewski-boxplot

I do think I have found an issue. If you use my visual test, sometimes the chart renders boxes where the whiskers fall outside of the y axis range:

image

If you look at the screenshot above, the data for the highlighted box doesn't seem to match with the rendered view. Is there an issues with your SVG path logic?

bjedrzejewski:
It definitely looks suspicious. I will investigate tomorrow.

chrisprice:
I've landed this directly onto master (e743e0f70ddf6730dc8edc29ed730710ca2e33d8). I've additionally fixed the bug and simplified the implementation of the errorBar and boxPlot series (b22a125c27aea560e314e1f17a8f5430e080895f & f44c859d5d095e34d14037b0754182e43dfc46b6).

Thanks @bjedrzejewski!

PR #801, opened 2015-11-30
Implements asymmetrical padding in util.extent

djmiley:


djmiley:
Implemented the percentages wrongly. Will also update the site documentation.

ColinEberhardt:
When discussed with @chrisprice  I was concerned that this would make the API more complex, however, now that I have seen that it is an overload of the current `pad` method, I like it :-)

chrisprice:
Thanks Drew, a very thorough implementation. I suppose the type being flexible (number or tuple of numbers) is consistent with how we handle margins elsewhere (expanded versus unexpanded) so that shouldn't be a problem.

PR #802, opened 2015-11-30
Support jsnext:main

tlsim:
See https://github.com/rollup/rollup/wiki/jsnext:main
I have used this with some success in another project that has d3fc as a dependency and bundles using rollup. As it stands I think `fc.version` will be incorrect for the consumer, but it seems unlikely that anyone would want to rely on this property at runtime.


chrisprice:
Agreed, thanks Tom.

PR #803, opened 2015-12-01
Implements absolute padding

### LINKED ISSUE #794: fc.util.extent() supports relative padding, but not absolute padding
Currently `fc.util.extent()` supports padding of data by a percentage, but in order to pad by a fixed amount, the user has to transform this padding into either a percentage, or points to be included in the extent- for which they probably need to run fc.util.extent() beforehand to find the original extent.

The use-case that I am currently trying to implement is the padding of a time domain by half a period in the situation, where calculations to ascertain this are uncertain due to possible missing data points.

Another use-case could be a simple graph displaying variations over time, where an absolute padding could act as a visual representation of the variance of the data. So if the padding was 5, and the range of the data was 10, it would take up 50% of the graph height. If the range of the data was 30, it would take up 75% of the graph height. I imagine that this is functionality that people may desire?

djmiley:
Resolves #794

### DIFF
--- a/site/src/components/util/extent.md
+++ b/site/src/components/util/extent.md
@@ -22,19 +22,34 @@ A range can be symmetrical about a given value, using the `symmetricalAbout` pro
       .symmetricalAbout(20); // [0, 40]
 ```
 
- The range can be modified by using the `pad` property. Pad will scale the calculated range by the given amount. If the calculated range is `[10, 20]` then the result of padding by '1' will be `[5, 25]`. `pad(0)` is an identity.
+ The range can be modified by using the `pad` property, and specifying a 'padUnit' to vary the type of padding required. The padUnit can be either 'domain' for padding the domain absolutely or 'percent' for padding by a percentage. It is set by default to 'percent'. Domain padding will increase the calculated range by the amount given. If the calculated range is `[10, 20]`, with padUnit set to 'domain', then the result of setting pad to '2' will be `[8, 22]`. `pad(0)` is an identity in both cases.
 
 ```js
   fc.util.extent()
       .fields(fields)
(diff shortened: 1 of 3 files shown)


ColinEberhardt:
Hi @djmiley - sorry it's taken a while to review this. I've had a look at these changes and discussed with @chrisprice also. I'm happy with the concept of adding absolute as well as percentage padding to the API. However, I'm not so keen on the proposed API.

A few issues:

 1. The updated API is pretty verbose, whereas D3 favours quite a terse syntax
 2. The precedence order of absolute / percent padding is not terribly obvious
 3. This is a breaking change, in order to release a [few other updates would be required](https://github.com/ScottLogic/d3fc/search?utf8=%E2%9C%93&q=.pad).

How about as an alternative ... keep the existing `pad` method, but supplement the behavior with a `padUnit` property:

```
fc.util.extent()
      .fields(fields)
      .padUnit('percent')
      .pad(0.5)(data);
```

Where `padUnit` takes values of `percent` or `domain`, with `percent` being the default.

I think this fixes the issues mentioned above.

djmiley:
Thanks for looking at this @ColinEberhardt . That sounds good, I originally thought about implementing this way but wasn't entirely sure which was the best option.

ColinEberhardt:
Thanks for the updates @djmiley - looks good to me :-)

ColinEberhardt:
@djmiley I'll perform a release later today so that you can pick up these changes. I'm just awaiting a few updates on #800 which can hopefully be merged too.

ColinEberhardt:
@djmiley this has been released https://github.com/ScottLogic/d3fc/releases/tag/v5.2.0

PR #805, opened 2015-12-02
Simulated annealing label placement algorithm

WPFerg:
The last of the label placement algorithms, using simulated annealing. ([Description of algorithm](http://www.theprojectspot.com/tutorial-post/simulated-annealing-algorithm-for-beginners/6))

A few things to note:
* Each iteration, the algorithm compares the current state (placements) with the a potential next state (one rectangle moved to one other placement)
* The scoring function is the area of intersections, discouraging labels from leaving the container -- but does not prevent them.
* Because of the random nature of the algorithm, outputs will be different each time, and the tests have mocked `Math.random` to be consistent across tests

### DIFF
--- a/src/layout/strategy/annealing.js
+++ b/src/layout/strategy/annealing.js
@@ -1,0 +1,88 @@
+import d3 from 'd3';
+import {rebindAll} from '../../util/rebind';
+import minimum from '../../util/minimum';
+import {allWithCollisions, totalCollisionArea} from './collision';
+import containerUtils from './container';
+import {getAllPlacements} from './placement';
+
+export default function() {
+
+    var container = containerUtils();
+    var temperature = 1000;
+    var cooling = 1;
+
+    var strategy = function(data) {
+
+        var originalData = data;
+        var iteratedData = data;
+
+        var lastScore = Infinity;
+        var currentTemperature = temperature;
+        while (currentTemperature > 0) {
+
+            var potentialReplacement = getPotentialState(originalData, iteratedData);
+
+            var potentialScore = scorer(potentialReplacement);
+
+            // Accept the state if it's a better state
+            // or at random based off of the difference between scores.
+            // This random % helps the algorithm break out of local minima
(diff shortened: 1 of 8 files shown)


src/layout/strategy/annealing.js
```diff
@@ -0,0 +1,88 @@
...
+        // For one point choose a random other placement.
+        var dataClone = iteratedData.slice();
+
+        var victimLabelIndex = Math.floor(dataClone.length * Math.random());
+        var label = originalData[victimLabelIndex];
+
+        var replacements = getAllPlacements(label);
+        var randomPlacementIndex = Math.floor(replacements.length * Math.random());
+
+        dataClone[victimLabelIndex] = replacements[randomPlacementIndex];
```
ColinEberhardt: The need to clone an array and replace a single item at a given index comes up over and over again in these algorithms. It might be a good idea to move this into a common utility function, e.g.

```
cloneAndReplace(sourceArray, index, replacement)
```

Also, how about adding a function for randomly picking an element from an array? e.g.

```
pickRandom(sourceArray) - randomly returns an array element
randomIndex(sourceArray) - returns a random index
```

These could be added to fc.utils, perhaps in a single file 'array.js'?

With these two changes, `getPotentialState` should be just one or two lines ...

ColinEberhardt:
This looks really good - great job :+1: 

As mentioned in the comment, it would be good to see some of the common array manipulation moved into a utility function.

Also, a few of the strategies have logic that tries to ensure the child items are within the bounding container. This looks like a non-standard modification to the algorithms. I'm wondering whether this should be an options, e.g. `enforceBounds(true)` - perhaps have a think about that and add as a future enhancement?

WPFerg:
Perhaps only enforce the bounds if `containerWidth` and `containerHeight` are called? To me that would probably be a sensible base case, as that's all the width/height are used for.

WPFerg:
I've pushed up the changes needed. I'll submit another PR for the `enforceBounds` condition when `containerWidth`/`Height` are called

ColinEberhardt:
Thanks for the updates - much appreciated :-)

PR #809, opened 2015-12-10
Callout

WPFerg:
Implements a callout in `fc.layout.callout`, using flexbox layout.

Set-up should be done by providing a strategy to use, and an `origin` accessor function to indicate where the data is on the chart that the callout should point to (optional). Data provided in the main function call should be in the same format as in `fc.layout.rectangles` in that it's an array of objects of `x, y, width, height` for where the label of the callout should be located. Further decoration, such as text placement, is done using `decorate`.

### DIFF
--- a/src/layout/callout.js
+++ b/src/layout/callout.js
@@ -1,0 +1,77 @@
+import d3 from 'd3';
+import {noop, identity} from '../util/fn';
+import {rebindAll} from '../util/rebind';
+import layoutRectangles from './rectangles';
+import arrayFunctor from '../util/arrayFunctor';
+
+export default function(layoutStrategy) {
+
+    var origin = d3.functor([0, 0]),
+        component = noop,
+        decorate = noop,
+        strategy = layoutStrategy || identity;
+
+    var rectangles = layoutRectangles(strategy);
+
+    var line = d3.svg.line()
+        .x(function(d) { return d[0]; })
+        .y(function(d) { return d[1]; });
+
+    var callout = function(data) {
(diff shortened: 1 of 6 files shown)


WPFerg:
Pushed a change that means the origin point is in absolute space, rather than relative to the anchor point -- this meant adding `rectangles.translate` that keeps track of the translations similar to `rectangles.anchor`.

chrisprice:
Just been looking at the visual test and I'm not sure it's behaving how I would expect -

image

I've removed the fill on callout 8 and you can see the path cuts across the rect. 

Thinking more about it, I think the API is back to front from what I would expect. I wouldn't expect a callout component to require me to specify the location of the label. Instead I would expect that to be derived by the algorithm based on the `origin`. I'm interested in showing labels for data-points, not locating labels and then referencing data-points (even if that's what the algorithm ultimately does).

I might just have the wrong end of the stick though!

WPFerg:
Yeah, I can see where you're coming from. WRT the path, it's drawing a line straight from the anchor point to the origin. From what I remember, It's just a wrapper around the rectangles placement. 

Having a further think now, a force-based floating placement would probably work to derive an anchor point, but would need additional context (ie. line/area/etc collision checking). Even a simple placement based off of where the origin is in the container could work. This sort of algorithm would likely prove useful in being its own component and reused in this one. I'll see what I can come up with

chrisprice:
Thanks. In the mean time I'm going to close this PR as I don't think the future implementation will have much relation to this one.

PR #810, opened 2015-12-10
Non-enforced bounds on layout rectangle container utility

WPFerg:
Small change that doesn't enforce container bounds if they aren't changed.

### DIFF
--- a/src/layout/strategy/container.js
+++ b/src/layout/strategy/container.js
@@ -1,12 +1,14 @@
 export default function() {
-    var containerWidth = 0,
-        containerHeight = 0;
+    var containerWidth = null,
+        containerHeight = null;
 
     var container = function(point) {
-        return !(point.x < 0 || point.y < 0 ||
+        // If the bounds haven't been defined, then don't enforce
+        return (containerWidth == null && containerHeight == null) ||
+            (!(point.x < 0 || point.y < 0 ||
             point.x > containerWidth || point.y > containerHeight ||
             (point.x + point.width) > containerWidth ||
-            (point.y + point.height) > containerHeight);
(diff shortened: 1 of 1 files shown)


src/layout/strategy/container.js
```diff
@@ -3,10 +3,12 @@ export default function() {
         containerHeight = 0;
 
     var container = function(point) {
-        return !(point.x < 0 || point.y < 0 ||
+        // If the bounds haven't been defined, then don't enforce
```
chrisprice: I'm not keen on the conflation of `0` and `falsey`. If `0` is a valid value then we should check against `null` otherwise we should explicitly check against `0`.

WPFerg: I think, in this instance, evaluating against `null` would be more sensible, since `0` would probably be a valid value (if it was for a case of explicitly always failing). I've made the change.

src/layout/strategy/container.js
```diff
@@ -1,12 +1,14 @@
 export default function() {
-    var containerWidth = 0,
-        containerHeight = 0;
+    var containerWidth = null,
+        containerHeight = null;
 
     var container = function(point) {
-        return !(point.x < 0 || point.y < 0 ||
+        // If the bounds haven't been defined, then don't enforce
+        return (containerWidth === null && containerHeight === null) ||
```
chrisprice: Sorry I know this runs slightly counter to my previous comment but I would always go for ```== null```. I've yet to see a compelling reason to distinguish between the two.

WPFerg: Thanks, fixed + squashed.

chrisprice:
:+1:

PR #811, opened 2015-12-11
Rectangle strategy web docs

WPFerg:
Just a few web docs for the rectangle strategies:

* I gave them their own section. It doesn't really make much sense for them to have their own section, but at the same time it doesn't make much sense for them to be in the Layout section either.
* The example code is just a trimmed down version of the rectangles example, with the core of that being hidden.

### DIFF
--- a/site/src/_config.yml
+++ b/site/src/_config.yml
@@ -1,4 +1,4 @@
-component-ids: [introduction, chart, series, indicator, scale, svg, annotation, tool, layout, util, data, samplers]
+component-ids: [introduction, chart, series, indicator, scale, svg, annotation, tool, layout, rectangle-strategy, util, data, samplers]
 
 components:
   introduction:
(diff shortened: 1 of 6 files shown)


ColinEberhardt:
Thanks - great work on these docs. I think they might work better on a single page. See this update:

https://github.com/ColinEberhardt/d3-financial-components/commit/ef122e8a4fbce922ea612a67f12b4e0effe7e42e

I also changed the `anchor` function so that it supplies datum and index, which is more in-line with the other accessor functions. It also removes the need to maintain a separate `anchors` array - I think the most natural way to use this would be to write the anchor data back onto the bound data.

Anyhow, take a look at the changes and see what you think!

WPFerg:
That's all fine to me -- good idea about the `anchor` function too :)

ColinEberhardt:
Thanks :-)

PR #816, opened 2015-12-24
Update Wealth & Health of Nations example for better experience on mobile/tablet

jleft:
- Add loading message when data is loading, with progress event (useful when loading over a slow connection)
- Update year label size to vary based upon chart width (label was far too large at small screen sizes)
- Update x-axis tick count to vary based upon width of chart (tick labels would overlap)
- Update to recalculate layout when window is resized
- Update indentation for consistency
- Handle touch start and end events to update year label active state
- Update point radius to respond to chart width


ColinEberhardt:
Great stuff - thanks @jleft

Issue #763, opened 2015-11-09
Tidy fc.data.random.financial


owennw:
This component is a bit messy, and required a refactor.

1. The `calculateInterval` performs date manipulation using a `millisecondsPerYear` magic number.

2. The component mutates the `startPrice`,  `startDate` and `startVolume` properties as it is executed.

3. There are many properties that are not documented, for example the `mu` and `sigma` properties.

This list is not exhaustive.

ColinEberhardt:
I'd combine this with #754 - basically reimplement the component, rather than just try to fix up the current implementation.

ColinEberhardt:
At the same time I think it's worth looking at the API that this component exposes. Currently you ask it to give you 'n' points from a specified start date. For most charts what you actually want is to generate data for a pre-determined data range at a specific period, e.g. 3 months @ 1day, 1day @ 5 mins.

However, we also need to be able to stream new data into a chart, which would require a different interface. Should a streaming interface be stateful?

Probably worth asking @chrisprice for his opinions on the interface also.

chrisprice:
For stateful functionality I'd like to see something like -

```js
var generator = fc.data.random.financial()
  .start(new Date())
  .interval(d3.time.minute);

generator(new Date() + 1d); // [{ date: now, ... }, { date: now+1m, ... }]

var stream = generator.stream();

stream(); // { date: now, ... } 
stream(); // { date: now+1m, ... }
```

owennw:
Does that mean removing the ability to specify the number of points completely, in favour of setting a period and duration, and letting however many points that amounts to fall out?

ColinEberhardt:
> Does that mean removing the ability to specify the number of points completely, in favour of setting a period and duration, and letting however many points that amounts to fall out?

I'd say so, yes. I think it is an API that s a better reflection of how the component would be used.

Issue #754, opened 2015-11-08
Allow control of the time increment in fc.data.random.financial


chrisprice:
Currently it always ticks by 1 day, it would be nice to be able to e.g. generate a 15min delayed 5 second stream.

chrisprice:
Now part of d3fc/d3fc-random-data which includes this functionality.

PR #832, opened 2016-01-08
New interface for random.financial

tlsim:
* Implements the interface discussed in #763, allowing for controlling the time increment #754.
Time increment is provided as a number of seconds, not a d3 time interval - it's not possible to create a d3 time interval of 5 mins, for example.
* Replaces volume model with something simpler, and makes it configurable as a functor property.
* Component function now does not mutate state - stateful usage is provided by the closure returned from `generator.stream()`, as proposed.

Todo:
- [x] Convert usages
- [x] Improve documentation

### DIFF
--- a/site/src/components/data/financial.md
+++ b/site/src/components/data/financial.md
@@ -10,7 +10,7 @@ externals:
   financial-stream-example-html: financial-stream-example.html
 ---
 
-The random financial data generator component generates open-high-low-close-volume financial data. 
+The random financial data generator component generates open-high-low-close-volume financial data.
 Prices are calculated using the [random walk](./walk.html) component.
 A `filter` can be supplied to exclude points from the output. 
 Pre-defined filters can be used, for example to skip weekends use `fc.data.random.filter.skipWeekends`.
@@ -31,10 +31,12 @@ Generates the following:
 {{{ dynamic-include 'codepen' html="financial-example-html" js="financial-example-js" }}}
 
 ## Stream
-The component exposes a stateful streaming interface.
+If you need to make successive calls to generate data while keeping track of the latest date and price, 
+use the streaming interface.
 You can see this in action in the [streaming chart example](../../examples/streaming/index.html).
-First create a stream from a random.financial instance with `financial.stream()`.
-Data points can then be generated using the stream's `next`, `take`, and `until` methods — the stream will use the latest date and price for subsequent calls.
+First create a stream from an instance of `fc.data.random.financial` by calling the `stream` method.
+Data points can then be generated using the stream's `next`, `take`, and `until` methods 
+— the stream will use the latest date and price for subsequent calls.
 
 An example using `next`, `take` and `until`:
 
--- a/site/src/components/data/financial.md
+++ b/site/src/components/data/financial.md
@@ -6,9 +6,14 @@ namespace: Data
 externals:
   financial-example-js: financial-example.js
   financial-example-html: financial-example.html
+  financial-stream-example-js: financial-stream-example.js
+  financial-stream-example-html: financial-stream-example.html
 ---
 
-The random financial data generator component is a useful testing utility that generates data via a random walk algorithm.
+The random financial data generator component generates open-high-low-close-volume financial data. 
+Prices are calculated using the [random walk](./walk.html) component.
+A `filter` can be supplied to exclude points from the output. 
+Pre-defined filters can be used, for example to skip weekends use `fc.data.random.filter.skipWeekends`.
 
 The following example:
 
@@ -18,11 +23,28 @@ The following example:
 
 Generates the following:
 
-{{{ dynamic-include 'codepen' html="financial-example-html" js="financial-example-js" }}}
-
 {{{financial-example-html}}}
 <script type="text/javascript">
 {{{financial-example-js}}}
-</script>
+</script> 
 
-A `filter` can be supplied to skips other days, or change the start price, volume and date. Pre-defined filters can be used, for example to skip weekends use `fc.data.random.filter.skipWeekends`.
+{{{ dynamic-include 'codepen' html="financial-example-html" js="financial-example-js" }}}
+
+## Stream
+The component exposes a stateful streaming interface.
+You can see this in action in the [streaming chart example](../../examples/streaming/index.html).
+First create a stream from a random.financial instance with `financial.stream()`.
+Data points can then be generated using the stream's `next`, `take`, and `until` methods — the stream will use the latest date and price for subsequent calls.
+
+An example using `next`, `take` and `until`:
+
+```js
+{{{ codeblock financial-stream-example-js}}}
+```
+
+{{{financial-stream-example-html}}}
+<script type="text/javascript">
+{{{financial-stream-example-js}}}
+</script> 
+
+{{{ dynamic-include 'codepen' html="financial-stream-example-html" js="financial-stream-example-js" }}}
--- a/site/src/examples/streaming/streaming.js
+++ b/site/src/examples/streaming/streaming.js
@@ -1,11 +1,10 @@
 // create some test data
-var generator = fc.data.random.financial();
-var data = generator(110);
+var stream = fc.data.random.financial().stream();
+var data = stream.take(110);
 
 function renderChart() {
     // add a new datapoint and remove an old one
-    var datum = generator(1)[0];
-    data.push(datum);
+    data.push(stream.next());
     data.shift();
 
     // compute the bollinger bands
--- a/src/data/random/financial.js
+++ b/src/data/random/financial.js
@@ -1,4 +1,4 @@
-import walk from './walk';
+import _walk from './walk';
 import d3 from 'd3';
 
 export default function() {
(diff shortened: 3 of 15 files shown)


tlsim:
We have the use case of using this component to create an initial array, then add new data using the stream, and we want the open price of the first streamed point to be related to the last point of initial data.

What's preferred? 
```javascript
var generator = fc.data.random.financial()
  .start(new Date(2016, 0, 1));

// Currently
var data = generator(new Date(2016, 1, 1));
generator.startDate(new Date(2016, 1, 1));
generator.startPrice(/* something sensible */);
var stream = generator.stream();
data.push(stream());

// Or this?
var stream = generator.stream(); 
var data = stream(/* end date and/or number of points? */); // Returns array
data.push(stream()[0]);

// Or revert to having generator mutate date and price?
var data = generator(new Date(2016, 1, 1));
var stream = generator.stream();
data.push(stream());
```

ColinEberhardt:
Thanks @tlsim - this is going to be a big improvement. Here's some notes on our discussion:

The generator shouldn't have an API that distinguishes between generating an array of items versus streaming single datapoints. One good way to solve this is to make everything a streaming operation.

```
// a factory that creates streams with a given initial state.
var generator = fc.data.random.financial()
  .start(new Date(2016, 0, 1));

// construct a stream from the generator - you can of course do this multiple times. 
var stream = generator();

// populate an array by asking for a number of points
var data = stream.next(20);

// stream new items into the array
data.push(stream.next());
data.push(stream.next());
data.push(stream.next());


// another way to populate the initial data
var stream2 = generator();
var data2 = stream2.until(function(datum) { datum.date > new Date(2015,12,12); });

```

chrisprice:
One final (?!) tweak to the API -

```js
var dataGenerator = fc.data.random.financial();

var stream = dataGenerator.stream();

var snapshot = stream.take(2); // starts from start (first call to stream) [{}, {}]

var datum = stream.next(); // continues from above {}

stream.until(function(datum) { datum.date > new Date(2015,12,12); }) // continues from above [{},...]

var data = dataGenerator(2); // always starts from start date [{}, {}]
```

The above code tries to ensure consistent return types and allow the following 3 use cases -

* Generate n data sets of a fixed range
* Subscribe to a stream of data points
* Request a snapshot to back-fill a range, then stream in new data

tlsim:
Updated.
One more thing to check:
`stream.until` takes function(datum) but `filter` takes function(date). This might feel inconsistent. Is this worth changing?

ColinEberhardt:
> One more thing to check: stream.until takes function(datum) but filter takes function(date). This might feel inconsistent. Is this worth changing?

:+1: for changing the `filter` function for consistency.

tlsim:
@ColinEberhardt @chrisprice thanks for the guidance! I think this is ready to review.

site/src/components/data/financial.md
```diff
@@ -18,11 +23,28 @@ The following example:
...
 {{{financial-example-js}}}
-</script>
+</script> 
+
+{{{ dynamic-include 'codepen' html="financial-example-html" js="financial-example-js" }}}
+
+## Stream
+The component exposes a stateful streaming interface.
+You can see this in action in the [streaming chart example](../../examples/streaming/index.html).
+First create a stream from a random.financial instance with `financial.stream()`.
```
chrisprice: I think this is a bit confusing, maybe reword it to -

First create a stream from an instance of ```fc.data.random.financial``` by calling the ```stream()``` method.

chrisprice: Actually you've used the convention of methods without parenthesis below so stick to that (i.e. ```stream```).

site/src/components/data/financial.md
```diff
@@ -18,11 +23,28 @@ The following example:
...
 {{{financial-example-html}}}
 <script type="text/javascript">
 {{{financial-example-js}}}
-</script>
+</script> 
+
+{{{ dynamic-include 'codepen' html="financial-example-html" js="financial-example-js" }}}
+
+## Stream
+The component exposes a stateful streaming interface.
```
chrisprice: I would try and make it clearer what the difference between the stream and the base component is.

site/src/examples/streaming/streaming.js
```diff
@@ -1,11 +1,10 @@
 // create some test data
```
chrisprice: These are great examples of how much better things are with these changes!

src/data/random/financial.js
```diff
@@ -1,155 +1,127 @@
...
-                volume: volumes[currentStep]
-            });
-            currentIntraStep += stepsPerDay;
-            currentStep += 1;
-            startDate.setUTCDate(startDate.getUTCDate() + 1);
-        }
-        return ohlcv;
+    var startDate = new Date();
+    var startPrice = 100;
+    var granularity = 60 * 60 * 24;
```
chrisprice: This could be spun out as another issue but would it make sense to use [```d3.time.intervals```](https://github.com/mbostock/d3/wiki/Time-Intervals#interval) for this? Not only does it require mental gymnastics to parse the value, it's also in seconds rather than milli-seconds as everywhere else in JS.

src/data/random/financial.js
```diff
@@ -1,155 +1,127 @@
...
-        return calculateOHLC(days, prices, volumes).filter(function(d) {
-            return filter(d.date);
-        });
+    function calculateOHLC(start, price) {
+        var prices = walk
+            .period(granularityInYears())(price);
+        var datum = {
+            date: start,
+            open: prices[0],
+            high: Math.max.apply({}, prices),
```
chrisprice: I think the context argument should be ```Math```? E.g. this is what Babel produces for spread operators -
```js
Math.max(...[1,2,3])
/// --->
Math.max.apply(Math, [1, 2, 3]);
```

src/data/random/financial.js
```diff
@@ -1,155 +1,127 @@
...
-    };
-    gen.startPrice = function(x) {
-        if (!arguments.length) {
-            return startPrice;
+    financial.stream = makeStream;
+
+    function makeStream() {
+        var latest;
+        function getDatum() {
+            if (!latest) {
```
chrisprice: Another advantage of ```d3.time.interval``` would allow the offset to be stored, simplifying this a bit.

src/data/random/financial.js
```diff
@@ -1,155 +1,127 @@
...
         }
-        startPrice = x;
-        return gen;
-    };
-    gen.startVolume = function(x) {
-        if (!arguments.length) {
-            return startVolume;
+        function getNextDatum() {
+            var ohlc = getDatum();
+            while (filter && !filter(ohlc)) {
```
chrisprice: Would ```do/while``` simplify this?

src/data/random/financial.js
```diff
@@ -1,155 +1,127 @@
...
+        };
+        stream.take = function(numPoints) {
+            var point;
+            var data = [];
+            for (point = 0; point < numPoints; point += 1) {
+                data.push(getNextDatum());
+            }
+            return data;
+        };
+        stream.until = function(comparison) {
```
chrisprice: Would it make sense to pass ```d, i``` to the comparison function to match the d3 pattern? If so, could take just be a specialisation of this function?

src/data/random/financial.js
```diff
@@ -1,155 +1,127 @@
...
+            var data = [];
+            if (!comparison || latest && comparison(latest)) {
+                return data;
+            }
+            var ohlc = getNextDatum();
+            while (comparison) {
```
chrisprice: I can't quite decipher this from looking but this code looks like it's going to cause the value which fails the comparison test to be skipped on the next request for more data?

chrisprice: Scratch that, I had misunderstood the API.

chrisprice: Now I understand the API behaviour (stream.until should generate data up to and including the datum that satisfies the specified condition), I wonder if it should more closely match the [RxJS takeWhile behaviour](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/takewhile.md). Maybe another one for the issue pile?

src/data/random/financial.js
tlsim: Yeah I took another look and it seems possible. We would need to pass `step` to get multiples of each interval, and use `interval.offset(date, step)` internally.
E.g. 5 minute intervals:
```javascript
fc.data.random.financial()
  .startDate(new Date(2016, 0, 1))
  .startPrice(100)
  .interval([d3.time.minute, 5]);
```

src/data/random/financial.js
tlsim: NIce idea. I'll see what I can do.

src/data/random/financial.js
tlsim: Yeah I don't much like how this turned out - I think the takeUntil behavior would work, would just have to move setting the latest point out of `getDatum`, otherwise the next call to a stream method after an `until` call returns the date unexpectedly incremented.

tlsim:
Updated to address comments. Matching RxJS takeUntil more closely has simplified things. Also now allows for setting the units of mu and sigma with an interval offset, so no more 'ms per year' constant.

ColinEberhardt:
Thanks for the update @chrisprice are you ok to review?

chrisprice:
LGTM :+1: 

Merging is going to break the site as per #873.

Issue #757, opened 2015-11-08
Improve Playground's method of mapping supplemental files


jdunkerley:
At the moment the playground does this but pattern matching the 'data. but can be done with Assemble variables.

Perhaps something like a relativePath variable in YAML.

owennw:
I'm not sure what the first sentence means..

chrisprice:
Fixed (in a manner of speaking) by 2bf4206de72876ee4b4044f2ae88f691c54400eb

Issue #721, opened 2015-11-02
Adjust Assemble to do one pass using JS files for playground


jdunkerley:
If we move to JS files for components can probably get back to one pass and allow more complicated examples in the playground.

Based on discussion with @chrisprice

chrisprice:
Fixed (in a manner of speaking) by 2bf4206de72876ee4b4044f2ae88f691c54400eb

Issue #719, opened 2015-11-02
Add Examples to Playground


jdunkerley:
Playground only set up on components series and indicators

owennw:
Is the playground gallery staying? / are we going to do this?

ColinEberhardt:
That's very much open to debate  / discussion in my opinion. First off, I think its worth working out technically how this would work.

Issue #703, opened 2015-10-29
Make {{>example-fixture}} work with more than 1 example


bjedrzejewski:
We often use {{>example-fixture}} to simplify the website code. However, when more than one example is there, we need to use something like:

```
<div id="series_point_2" class="chart"> </div>
<script type="text/javascript">
(function() {
    var desiredWidth = $('#series_point_2').width(),
        desiredHeight = desiredWidth / 2.4; //keeps the width-height ratio at 600-250 (defaults for createFixture)
    var f = createFixture('#series_point_2', desiredWidth, desiredHeight, null, function() { return true; });
    var container = f.container, data = f.data,
      xScale = f.xScale, yScale = f.yScale;
    {{{example-code-2}}}
}());
</script>
```

Which is a bit disappointing. Solution for this problem would be great.

jdunkerley:
If I ever get the playground to work will look at these

chrisprice:
It would probably start with burning assemble but I would also love to see
this

ColinEberhardt:
> However, when more than one example is there, we need to use something like:

I have to confess ... I am responsible for this nasty cut'n'paste approach. I blame Assemble. Burn it.

chrisprice:
Fixed by 2bf4206de72876ee4b4044f2ae88f691c54400eb

PR #835, opened 2016-01-13
Throwing away Assemble and starting again!

### LINKED ISSUE #730: A lot of the code within the website is not linted

ColinEberhardt:
fixes #730 #757 #719 #721 #703 #469 #686

I've been looking at getting rid of Assemble in favour of just  doing it ourselves using the underlying tools (markdown parser, globbing, handlebars, front matter parser, etc ...)

This should allow us to stop fighting Assemble the whole time.

I've found it to be quite straightforwards, the current code is here:

https://github.com/ColinEberhardt/d3-financial-components/tree/master/site-builder

It's just over 100 LOC, and hopefully will not grow much beyond this. It is a bit messy at the moment too!

Big wins:
 + It's x10 faster! Assemble was 2.5 secs, this is 250ms.
 + Example code is now linted, I've introduced an 'externals' concept (see [band](https://github.com/ColinEberhardt/d3-financial-components/blob/master/site/src/components/annotation/band.md)) where front matter variables can be loaded from external files.
 + **controversial** - I've removed the playground in favour of using codepen. We can create a codepen via a simple HTTP post giving users a way to save / share their code snippets. It also significantly reduces the amount of website-related code we have to manage. Here's the (messy) [codepen wire-up](https://github.com/ColinEberhardt/d3-financial-components/blob/master/site/src/_includes/codepen.hbs).
 + I've thrown away all of the messy test-fixture code (of which we had two versions, one for the site, one for playgrounds), in favour of just having clean example code.
 + With the codepend approach, plus externals, you can have as many 'playgrounds' as you like on a page.

I've mostly been focusing on the components pages, with a few tweaks to the menu they are almost done. After that I'll look at the index (the interesting part is the mini-component examples) and the examples pages.

Feedback welcome!

(Sorry about the playgrounds @jdunkerley!)


chrisprice:
I really like the sound of these changes! I was a big fan of the playground but if we get the same functionality without having to maintain it ourselves then that sounds like a win.

It sounds like these changes also address a piece of feedback I had whereby it wasn't obvious how to go from the website documentation to a working piece of code. In some cases they hadn't spotted the playground link and in others there wasn't one. (This has just made me think we should make the codepen links more prominent, e.g. buttons).

I'm not sure what feedback you're looking for at the minute, assuming you don't want me to look too much at the implementation yet? I'd expect any comments to be simplifying implementation or renaming of fields, neither of which are worth doing until you're finished.

chrisprice:
If you're not after feedback at the moment, can you drop the `awaiting-review` label? :neckbeard:

ColinEberhardt:
Neck beard? That's an odd emoji! Yes, too early for a detailed review. I just wanted to make people aware that this was in progress.

I'll hopefully have time to complete the migration and move the custom build code to a separate repo early next week. I'm resisting the urge to make too many changes to the site itself!

chrisprice:
:squirrel:

ColinEberhardt:
@ScottLogic/d3fc - ok, this is ready for feedback now. There's a few bits I still want to tidy up (and document!), but i think it is in a position where it could be merged as a replacement for the current assemble site.

chrisprice:
From reviewing the output -

* Top charts on homepage don't render `Uncaught TypeError: Cannot read property 'color' of undefined`. This is also causing problems on the example pages for bubble, scatter and stacked bar.
* Get rid of components from homepage?
* Markdown not rendering

image
* Not enough margin on chart 

image
* The [transitions example](http://localhost:8000/components/introduction/3-transitions.html) causes a growing chart on codepen
* [Error bar documentation](http://localhost:8000/components/series/errorBar.html) has a broken chart
* The currently selected example doesn't highlight in the navigation like it does on the documentation pages.

site/src/index.html
```diff
@@ -0,0 +1,153 @@
+---
+layout: default
+externals:
+  index-example-js: index-example.js
```
chrisprice: I don't think this is used. If it isn't this line **and the file** can be deleted

src/fc.css
```diff
@@ -66,7 +66,7 @@ text {
 .x-axis.top .label,
 .y-axis.left .label,
 .title .label {
-  dominant-baseline: hanging;
```
chrisprice: What do these changes do?

chrisprice:
Delete `site/src/lib`

site/src/_includes/codepen.hbs
```diff
@@ -0,0 +1,13 @@
+<div class="codepen-link">
+  <form action="http://codepen.io/pen/define" method="POST" target="_blank">
+    <input type="text" name="data" hidden="true"
+      value='{
+        "html":"{{ escape html}}",
+        "js": "{{{ escape js }}}",
+        "css": "{{#if css}}{{{ escape css }}}{{/if}}",
+        "js_external": "https://cdnjs.cloudflare.com/ajax/libs/d3fc/5.2.0/d3fc.bundle.js",
```
chrisprice: These hardcoded values are going to trip us up, worth changing to [this pattern](https://github.com/ScottLogic/d3fc/pull/835/files#diff-9b2d0e08433fde5a173fd219bf9ffad3L55)?

site/src/_layouts/default.hbs
```diff
@@ -55,8 +48,6 @@ script-extra: ''
       </div>
     </footer>
 
-    {{#if script}}<script src="{{script}}"></script>{{/if}}
-    {{#if script-extra}}{{{script-extra}}}{{/if}}
     <script src="//localhost:35729/livereload.js"></script>
```
chrisprice: Worth making this line conditional (if it's easy)?

package.json
```diff
@@ -46,15 +46,19 @@
...
-    "handlebars-helpers": "ColinEberhardt/handlebars-helpers",
+    "handlebars": "^4.0.5",
+    "handlebars-group-by": "^1.0.1",
+    "handlebars-helpers": "^0.5.8",
     "jasmine": "^2.3.2",
     "jit-grunt": "^0.9.1",
     "jquery": "^1.11.2",
     "jsdom": "^6.3.0",
     "matchdep": "^0.3.0",
+    "node.extend": "^1.1.5",
```
chrisprice: Can you not use `Object.assign`?

package.json
```diff
@@ -23,10 +23,10 @@
   ],
   "license": "MIT",
   "devDependencies": {
-    "assemble": "^0.4.42",
     "bootstrap": "^3.3.5",
     "browserstacktunnel-wrapper": "^1.4.2",
     "d3-bundler": "^0.2.7",
+    "gray-matter": "^2.0.2",
```
chrisprice: Do we actually have a dependency on this or is it left over from before `tiny-ssg`?

Gruntfile.js
```diff
@@ -60,47 +63,6 @@ module.exports = function(grunt) {
             }
         },
 
-        assemble: {
```
chrisprice: Woooohoooo!!!!

Gruntfile.js
```diff
@@ -333,6 +289,36 @@ module.exports = function(grunt) {
         }
     });
 
+    grunt.registerTask('tiny-ssg', 'builds the site', function() {
+        var done = this.async();
+        var globalData = {
+            package: grunt.file.readJSON('package.json')
+        };
+        process.chdir('site/src');
```
chrisprice: This seems harder than prefixing the source paths below?

src/fc.css
ColinEberhardt: well spotted - I forgot that i had made that tweak. I might move that into a separate PR as it will be buried within these changes.

chrisprice:
Looks much better to me. I think I've picked up most of the issues but there's a lot of changes to go through!

jdunkerley:
Sorry to see playgrounds go, but getting rid of Assemble gets my :+1:

chrisprice:
It is a shame but the open in codepen links work really well. With the
bonus that you can share your changes! Also I'd pretty much chew my arm off
before debugging assemble again!!

jdunkerley:
CodePen looks really nice, look forward to seeing new site live

ColinEberhardt:
@chrisprice I think that's everything fixed:

https://github.com/ScottLogic/d3fc/commit/70fee44b2df34c58e81593119f7293fe286dcfa4

Thanks @jdunkerley :-)

chrisprice:
I don't think any of the gruntfile or package.json comments have been addressed -

* [] Do we actually have a dependency on this or is it left over from before `tiny-ssg`?
* [] Can you not use `Object.assign`?
* [] This seems harder than prefixing the source paths below?

I'm going to make a few fixes myself to the homepage and then merge anyway but maybe you can take a look?

chrisprice:
I've pushed these changes directly to master but I'm leaving this open until the above comments are addressed.

ColinEberhardt:
> I don't think any of the gruntfile or package.json comments have been addressed

Good point, I did miss a few of them. I'll push an update later ...

ColinEberhardt:
> Do we actually have a dependency on this or is it left over from before tiny-sag?

My bad, I always forget to back out package installation when switching to a different solution. Fixed in #843 

> Can you not use Object.assign?

I use `node.extend` to do a deep copy of the objects - aiming for a more functional approach.

> This seems harder than prefixing the source paths below?

I think this comment has disappeared in the squash, although I guess you are referring to the way you have to change the working directory? I want to allow tiny-ssg to be configurable with a source folder, however, when I introduced this it made things far more complicated. I think I need to try a different angle.

chrisprice:
Closed by #843

PR #840, opened 2016-01-15
Maintain crosshair pointer while zooming with mouse wheel [fixes #836]

### LINKED ISSUE #836: Crosshair doesn't snap on zoom (on wheel event)
For an example see [Low Barrel](http://d3fc.io/examples/low-barrel/), and use the mouse wheel to zoom in - notice that the crosshair will remain in the same position and not snap.

I expect it's not uncommon to have a chart with a crosshair and zoom functionality - so a few people may run into this issue.

crosshair-zoom

mapmeld:
While you are zooming into a chart with a crosshair pointer, such as [the Low Barrel example](http://d3fc.io/examples/low-barrel/), the crosshair pointer stays relative to the div, until a mousemove event occurs.  Ideally the crosshair pointer should follow what it was pointing at, using the "onwheel" event.

Fixes #836

### DIFF
--- a/src/tool/crosshair.js
+++ b/src/tool/crosshair.js
@@ -56,6 +56,7 @@ export default function() {
                 .style('pointer-events', 'all')
                 .on('mouseenter.crosshair', mouseenter)
                 .on('mousemove.crosshair', mousemove)
+                .on('wheel.crosshair', mousewheel)
                 .on('mouseleave.crosshair', mouseleave);
 
             var overlay = container.selectAll('rect')
@@ -101,6 +102,16 @@ export default function() {
     }
 
     function mousemove() {
+        var mouse = d3.mouse(this);
+        var container = d3.select(this);
+        var snapped = snap.apply(this, mouse);
+        var data = container.datum();
+        data[data.length - 1] = snapped;
+        container.call(crosshair);
+        event.trackingmove.apply(this, arguments);
+    }
+
+    function mousewheel() {
         var mouse = d3.mouse(this);
         var container = d3.select(this);
         var snapped = snap.apply(this, mouse);
--- a/src/tool/crosshair.js
+++ b/src/tool/crosshair.js
@@ -56,7 +56,7 @@ export default function() {
                 .style('pointer-events', 'all')
                 .on('mouseenter.crosshair', mouseenter)
                 .on('mousemove.crosshair', mousemove)
-                .on('wheel.crosshair', mousewheel)
+                .on('wheel.crosshair', mousemove)
                 .on('mouseleave.crosshair', mouseleave);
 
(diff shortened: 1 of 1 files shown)


chrisprice:
Can I check if you've tested this on mobile with the pinch zoom gesture?

mapmeld:
@chrisprice I haven't tested what happens on a pinch zoom on either current or this modified version. It should only affect mouse wheel / scroll events.

ColinEberhardt:
Thanks for the contribution @mapmeld - I can probably test this on mobile at some point over the weekend if that would help ...

chrisprice:
Sorry to corrupt the PR with this discussion but I think we might be approaching this from the wrong angle, i.e. I think there's a general problem and this only solves one specific instance.

The general problem is that the crosshair is snapped to a domain co-ordinate only on mouse events (previously just `enter` and `move`). However, there are numerous things that could require the snapping to be re-evaluated e.g. -

* changing the existing data (any form of dynamic chart)
* adding a series to the existing data (in the case of a multi-point crosshair)
* changing the scale ranges/domains (the general case of this problem)
* changing the snap function
* other?

Therefore I wonder if we should store the cursor position in data and expand that into the crosshair points on each render i.e. run the snapping on render? 

This raises the question of how efficient the snap function is? That could be considered a separate issue or could fit nicely with breaking up crosshair in a similar way to the indicators package...

Again sorry for derailing the PR. I do appreciate the effort @mapmeld, we've gone round in circles with the crosshair a couple of times already!

ColinEberhardt:
That's a big can of worms you're opening up there! I'd be in favour of merging this PR if it makes things a little bit better, saving the more fundamental changes for a future update.

chrisprice:
Agreed raised as a separate issue (#841). 

If we want to solve the problem on mobile I think we also need to consider the `touchstart` & `touchmove` events in the same way [d3.zoom](https://github.com/mbostock/d3/blob/master/src/behavior/zoom.js) does. However, it might not be worth it given the complex interactions and the issue above.

Just the one comment on the actual implementation of this PR.

src/tool/crosshair.js
```diff
@@ -56,6 +56,7 @@ export default function() {
                 .style('pointer-events', 'all')
                 .on('mouseenter.crosshair', mouseenter)
                 .on('mousemove.crosshair', mousemove)
+                .on('wheel.crosshair', mousewheel)
```
chrisprice: Any reason not to reference `mousemove` rather than copy the implementation?

mapmeld: since it's working well with the same code, looks like they can be the same function. Just pushed a commit which eliminates the separate mousewheel function

chrisprice:
Thanks!

Issue #685, opened 2015-10-27
Make the d3fc.io website navigation mobile friendly


bjedrzejewski:
Make the navigation work well on the mobile for the website (be able to see the components list somehow).

The other task would be to make the examples work well on the mobile, but this is potentially much more work.

chrisprice:
2bf4206de72876ee4b4044f2ae88f691c54400eb has made this even worse than before

PR #849, opened 2016-01-27
Website updates

ColinEberhardt:
- Improves #685 by re-instating the mobile navigation
 - Components 'namespaces' are sorted
 - The 'Introduction' namespace is no longer special-cased

... and I'm going to stop messing around with the website build and do some real work after this ;-)

This PR basically moves to how I want to the static site generator to work, where you solve problems of structuring your metadata in code, rather than battling with templating engines.


ColinEberhardt:
> and I'm going to stop messing around with the website build and do some real work after this 

I lied, one last one ;-) https://github.com/ScottLogic/d3fc/commit/9412d091fb202ed8e03932312a4cad984e1b6045

Gruntfile.js
```diff
@@ -300,40 +298,18 @@ module.exports = function(grunt) {
...
         var config = {
             destinationFolder: '../dist',
             filePattern: ['components/**/*.md', 'index.html', 'examples/**/*.md'],
-            globalData: globalData
+            globalData: globalData,
+            sourceFolder: 'site/src'
         };
-        tinySSG.build(config)
+        siteBuilder(config)
             .then(function() {
```
chrisprice: Don't think the wrapper function is needed anymore?

Gruntfile.js
```diff
@@ -300,40 +298,18 @@ module.exports = function(grunt) {
...
-                process.chdir('../../');
                 done();
             })
             .catch(function(e) {
```
chrisprice: Might not be here either?

site/build/build.js
```diff
@@ -0,0 +1,79 @@
+var tinySSG = require('tiny-ssg');
+var handlebars = require('handlebars');
+var marked = require('marked');
+var matter = require('gray-matter');
+var process = require('process');
+
+var resolveExternals = require('./resolveExternals');
+var createComponentNavigation = require('./createComponentNavigation');
+
+// let's just pretend we have destructuring or ES6 module imports ...
```
chrisprice: Another way would be to export each step from a separate file...
```js
 var loadHandlebarsPartials = require('tiny-ssg/es5/steps/loadHandlebarsPartials');
 var loadGlobalData = require('tiny-ssg/es5/steps/loadGlobalData');
 var collectPagesFrontMatter = require('tiny-ssg/es5/steps/collectPagesFrontMatter');
 var addGlobalData = require('tiny-ssg/es5/steps/addGlobalData');
```

site/build/build.js
```diff
@@ -0,0 +1,79 @@
...
+require('./handlebars-helpers/escape').register(handlebars);
+require('./handlebars-helpers/codeblock').register(handlebars);
+require('./handlebars-helpers/json').register(handlebars);
+
+function build(config) {
+    // for dev builds don't syntax highlight
+    if (!config.globalData.dev) {
+        marked.setOptions({
+            highlight: function(code) {
+                return require('highlight.js').highlightAuto(code).value;
```
chrisprice: It's a bit weird to have the require here unless it's the require itself which is taking the time?

site/build/build.js
```diff
@@ -0,0 +1,79 @@
...
+                createComponentNavigation,
+                resolveExternals,
+                renderTemplate,
+                renderMarkdown,
+                renderLayout,
+                writePost(config.destinationFolder)
+            ]);
+        });
+    })
+    .then(function() { process.chdir(workingDirectory); });
```
chrisprice: There's no catch, does that mean that this could terminate with an error and be in the wrong directory?

site/build/build.js
```diff
@@ -0,0 +1,79 @@
...
+            }
+        });
+    }
+
+    var workingDirectory = process.cwd();
+    if (config.sourceFolder) {
+        process.chdir(config.sourceFolder);
+    }
+
+    return chainPromises({}, [
```
chrisprice: Might be worth looking at [`bluebird`](http://bluebirdjs.com/docs/getting-started.html)?

site/build/createComponentNavigation.js
```diff
@@ -0,0 +1,53 @@
+function groupBy(array, key) {
```
chrisprice: Might be worth using `lodash.groupby`?

chrisprice:
A few comments but nothing that would block a merge

ColinEberhardt:
> A few comments but nothing that would block a merge

Thanks @chrisprice - I'll take a look at the 'catch' issue before morning. That could make the build a little unpredictable. I've not heard of bluebird before, I'll take a look.

ColinEberhardt:
I've added catch (and tested it works), and addressed a few of your other comments - thanks for the review.

Issue #653, opened 2015-10-22
Grid lines and xaxis dont line up on frontpage example


jdunkerley:
![image](https://cloud.githubusercontent.com/assets/4699705/10660682/967d09a6-78a0-11e5-9b32-9f55ca1a5e12.png)

 Would be nice an easy change to make gridlines a multiple of axis ticks

ColinEberhardt:
I'll look at this when I tidy up the front page example.

PR #851, opened 2016-01-28
Gridlines now support ordinal scales

### LINKED ISSUE #850: Gridlines do not work with ordinal scales
The current gridline implementation expects scales to have a `ticks` function. It should follow the axis implementation where `domain` is used in this case.

ColinEberhardt:
Fixes #850 

Long-term it would probably make sense to add a base component that contains the tick logic, so that it can be shared between the axis and gridline component. This might also help resolve https://github.com/ScottLogic/d3fc/issues/653

However, for now this fix stops gridlines barfing when used with ordinal scales :-)

### DIFF
--- a/src/annotation/gridline.js
+++ b/src/annotation/gridline.js
@@ -7,8 +7,8 @@ export default function() {
 
     var xScale = d3.time.scale(),
         yScale = d3.scale.linear(),
-        xTicks = 10,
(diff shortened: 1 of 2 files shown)


chrisprice:
LGTM, worth raising an issue for the refactor?

PR #853, opened 2016-01-28
Random walk fix

ColinEberhardt:
It was outputting steps + 1 values

### DIFF
--- a/tests/data/random/walkSpec.js
+++ b/tests/data/random/walkSpec.js
@@ -19,11 +19,15 @@ describe('walk', function() {
         // b0ee0a397a6ecc4cf60663d5b51af15de3534ef1
         // and is assumed correct for the purposes of further refactor.
         expect(walkData)
-            .toEqual([10, 10.563051249613967, 10.862390962256633, 11.473999242766038, 11.799153741691457]);
+            .toEqual([ 10,
+              10.563051249613967,
+              10.862390962256633,
+              11.473999242766038,
+              11.799153741691457,
+              12.463506567556125 ]);
     });
 
     it('should accomodate boundary values', function() {
-        expect(fc.data.random.walk().steps(1)(10)).toEqual([10]);
-        expect(fc.data.random.walk().steps(0)(10)).toEqual([]);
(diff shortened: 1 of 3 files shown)


tests/data/random/walkSpec.js
```diff
@@ -0,0 +1,33 @@
+describe('walk', function() {
+
+    var element;
```
chrisprice: Copy/paste left over?

tests/data/random/walkSpec.js
```diff
@@ -0,0 +1,33 @@
...
+
+    beforeEach(function() {
+        element = document.createElement('svg');
```
chrisprice: Ditto?

chrisprice:
Referencing #832 as this change should be reflected there as well.

ColinEberhardt:
@chrisprice  Thanks for spotting, I've removed the leftovers

tests/data/random/walkSpec.js
```diff
@@ -0,0 +1,29 @@
...
+        var walk = fc.data.random.walk()
+            .steps(5);
+
+        var walkData = walk(10);
+
+        // this test data was generted from the random walk at the following point:
+        // b0ee0a397a6ecc4cf60663d5b51af15de3534ef1
+        // and is assumed correct for the purposes of further refactor.
+        expect(walkData)
+            .toEqual([10, 10.563051249613967, 10.862390962256633, 11.473999242766038, 11.799153741691457]);
```
tlsim: Should the output include the initial price? If so, I'd expect the output to have `steps` + 1 length if `steps` is the number of price increments. E.g. for steps = 1, period = T, Initial price S<sub>0</sub>, the output should be either [S<sub>0</sub>, S<sub>T</sub>] or [S<sub>T</sub>], but not [S<sub>0</sub>]. The current documentation for `steps` is a bit misleading here.

ColinEberhardt:
From discussion with @tlsim - the API for random walk is intended to represent the concept of dividing a period of time into `n` subdivisions. As a result, the output will be `n + 1` points.

PR #854, opened 2016-01-28
gridlines use the range function that accommodates ordinal scales

### LINKED ISSUE #653: Grid lines and xaxis dont line up on frontpage example

### LINKED ISSUE #852: Gridlines and axes should share common logic for tick placements
Continuation of #851 - might resolve #653 

Both gridlines and axes render ticks / lines at intervals specified by the associated scale. For linear they should use the scale's `tick` function, also axes support explicit `tickValues`. Both gridlines and axes should share a common API potentially implemented by common component.

ColinEberhardt:
Shouldn't have just relied on unit tests for #851 ;-)

 + Adds a ticks component (closes #852)
 + Streaming chart ticks and gridlines are aligned (fixes #653)

### DIFF
--- a/site/src/examples/streaming/streaming.js
+++ b/site/src/examples/streaming/streaming.js
@@ -32,14 +32,22 @@ function renderChart() {
       )
      .xDomain(xExtent(data))
      .yDomain(yExtent(data))
-     .xTicks(5)
      .yNice()
-     .yTicks(5)
      .chartLabel('Streaming Candlestick')
      .margin({left: 30, right: 30, bottom: 20, top: 30});
 
+    // obtain ticks from the underlying scales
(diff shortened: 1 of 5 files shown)


ColinEberhardt:
This gives us much better control over tick / gridline placement:

screen shot 2016-01-29 at 08 04 38

chrisprice:
Do you think the cartesian chart changes warrant some documentation around the difference between the different tick properties? It took me a little while to realise why the axis ticks would return something different to the scale ticks.

ColinEberhardt:
Thanks @chrisprice - updated based on review

Issue #681, opened 2015-10-26
Documentation examples do not support resizing


owennw:
As shown below, when the browser is narrow the sine graph example overlaps with the side bar.

![d3fc_cartesian](https://cloud.githubusercontent.com/assets/14889759/10730108/b6249e8e-7be5-11e5-8551-5fe45d4c3099.jpg)

bjedrzejewski:
I think this only happens when you start F12 Chrome dev tools or equivalent.

chrisprice:
None of the examples resize, I think that's the cause

PR #855, opened 2016-01-28
Support resizing for a number of examples

ColinEberhardt:
Improves #681 

The vast majority of the component examples now have a fixed width of 500px, so do not overlap the menu or have layout issues until the screen becomes quite narrow.

This changes ensures that all the 'full width' examples support resizing.


chrisprice:
:+1:

PR #856, opened 2016-01-29
Use text-before-edge baseline for labels

ColinEberhardt:
I picked the wrong baseline, it was originally 'hanging', should be 'text before edge'

Also, tweaked the build so that the site build is kicked off by changes in d3fc distribution


chrisprice:
:+1:

PR #858, opened 2016-01-29
Fix SVG axis tick and label data propagation

### LINKED ISSUE #857: SVG axis component doesn't guarantee correct propagation of data to its text and path elements
The SVG axis component groups each pair of ticks and their labels within a `g` element which has [tick values bound to them](https://github.com/ScottLogic/d3fc/blob/b0ee0a397a6ecc4cf60663d5b51af15de3534ef1/src/svg/axis.js#L99).

During _update_, the `g`'s child `path` and `text` [elements are selected using `selectAll`](https://github.com/ScottLogic/d3fc/blob/b0ee0a397a6ecc4cf60663d5b51af15de3534ef1/src/svg/axis.js#L114-L123), which has some special behaviour: 

> Only selectAll has special behavior regarding grouping; select preserves the existing grouping. The select method differs because there is exactly one element in the new selection for each element in the old selection. **Thus, select also propagates data from parent to child, whereas selectAll does not (hence the need for a data-join)!**

[- How Selections Work](http://bost.ocks.org/mike/selection/#non-grouping)

The use of `selectAll` over `select` causes the tick values not to propagate to the child `text` and `path` elements.

As there is only one `path` element and one `text` element within each `g`, I propose the `selectAll` is changed to `select`, so the grouping is maintained and the tick values will propagate.  

Check out [this CodePen](https://codepen.io/anon/pen/pgKvvg?editors=1010) for an example of where this can cause be a problem (also compares with the D3 axis):

d3fc-axis-selection

jleft:
Fixes #857

### DIFF
--- a/src/svg/axis.js
+++ b/src/svg/axis.js
@@ -111,14 +111,14 @@ export default function() {
 
             g.attr('transform', containerTranslate(scale, translate));
 
-            g.selectAll('path')
+            g.select('path')
                 .attr('d', function(d) {
                     return svgDomainLine(pathTranspose([
                         [0, 0], [0, sign * innerTickSize]
                     ]));
                 });
 
-            g.selectAll('text')
+            g.select('text')
                .attr('transform', translate(0, labelOffset))
                .text(tickFormatter);
 


chrisprice:
Nice find :+1:

PR #860, opened 2016-02-01
Add missing d3-legend to the dependency list on site Getting Started page

jleft:
I'm assuming the should also be listed here, as it's included in the bundle.


ColinEberhardt:
Thanks @jleft - well spotted

PR #874, opened 2016-02-08
Applied ema to force index and rewritten tests.

### LINKED ISSUE #871: Force index calculation is incomplete
The current implementation only calculates the force index for a single period. An exponential moving average is usually applied over multiple values to eliminate noise.

jdhodges:
Fixes #871.

### DIFF
--- a/src/indicator/algorithm/calculator/forceIndex.js
+++ b/src/indicator/algorithm/calculator/forceIndex.js
@@ -1,10 +1,16 @@
 import d3 from 'd3';
 import _slidingWindow from './slidingWindow';
+import {identity} from '../../../util/fn';
+import exponentialMovingAverage from './exponentialMovingAverage';
 
 export default function() {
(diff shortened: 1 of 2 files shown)


src/indicator/algorithm/calculator/forceIndex.js
```diff
@@ -13,7 +19,15 @@ export default function() {
         });
 
     var force = function(data) {
-        return slidingWindow(data);
+        emaComputer.value(value);
+        var forceIndex = slidingWindow(data).filter(function(val) {
```
ColinEberhardt: Can you just use `slidingWindow(data).filter(identity)` ?

ColinEberhardt:
One minor comment - other than that, looks good to me - thanks :-)

ColinEberhardt:
Nice job, thanks @jdhodges

PR #879, opened 2016-02-11
set svg vertical alignment to middle

### LINKED ISSUE #779: Growing chart issue
Arnaud threw together the following example of a chart which keeps growing vertically -

http://jsbin.com/deqema/edit?html,js,output

I'm not sure if there's anything we can/should do to fix this, as it relates to `svg` element's default `inline` display but it might be possible to do something in `innerDimensions`?

@ColinEberhardt

ColinEberhardt:
fixes #779

### DIFF
--- a/src/fc.css
+++ b/src/fc.css
@@ -75,6 +75,8 @@ text {
 
 .cartesian-chart, .multiples-chart {
   overflow: visible;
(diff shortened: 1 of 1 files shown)


chrisprice:
:+1:

PR #880, opened 2016-02-11
Financial website

### LINKED ISSUE #572: Split between core and financial functionality
d3fc started out as a set of financial components but has become more generic over time. Currently it sits half way between generic charting and the more specialized financial charting. 

It would be great if we could highlight the core charting strengths without detracting from the more complex financial components the library provides, and vice-versa.

The proposal is to make acknowledge the dual purpose of the library by making it more explicit, splitting d3fc into d3fc-core and d3fc-finance. What this would mean in practice is:

 + Maintain a single project and website (for the purposes of keeping the logistics simple)
 + Redesign the website to make the split more evident (e.g. separate component lists, make financial an 'add on')
 + Determine which components live in each package
 + Update the build to spit out separate core and financial modules.

While we are at it, we should review where financial concepts have influenced our designs:

 + Having default x and y values of `date` and `close` does not make sense outside of finance.
 + Drop linear time series? It is an inflexible sub-set of the cartesian chart functionality

@ScottLogic/d3fc - any other thoughts?

ColinEberhardt:
The 'components' section of the website has now become 'getting started', 'components', and 'financial'

I think this does enough to close #572 - I don't think we need to generate separate assets quite yet


chrisprice:
I'm getting the following warning trying to build this branch -
```
Running "tiny-ssg:dev" (tiny-ssg) task
Warning: ENOENT: no such file or directory, open '_layouts/section.hbs' Use --force to continue.
```

chrisprice:
When doing a ```site:serve```.

ColinEberhardt:
Ooops - I've added the missing file.

chrisprice:
The footer could do with some top padding (or otherwise). It's looking a bit cramped at the bottom -

image

chrisprice:
The getting started side menu looks a bit lost -

image

chrisprice:
For the getting started section, maybe split the first two off into How To and the others into Further Reading?

chrisprice:
Also, maybe change the section name to documentation rather than getting started?

Issue #881, opened 2016-02-12
It is difficult to get the width of the bars in various series


rjmcneill:
[ohlcBase](https://github.com/ScottLogic/d3fc/blob/f752ac1cd19cae25566db4297ae7d5fec8150f53/src/series/ohlcBase.js#L21-L23) has a method for getting the width of the bars.

Could be useful to be able to get the bar width in:
- bar series
- errorBar
- boxPlot
- macd
- elder ray

It also looks like you can't change the bar width on:
- grouped bar series
- macd
- elder ray

ColinEberhardt:
That's interesting, I don't think we'd ever considered that as being part of the public API. Out of interest, what are you using it for?

rjmcneill:
We're looking to use it to decorate the crosshair to be the full width of the bars.
We're using it [here][1] on the primary chart at the minute, but are looking into it for the secondaries.

[1]: http://scottlogic.github.io/BitFlux/master/

ColinEberhardt:
That makes sense. How about you split this into two PRs?

  1. Non-controversial, exposing bar width for the indicators you listed above - by using the rebind technique in your PR #882 
  2. Expose the `width` function for the other series types.

For (2) one thing to bear in mind is that the series first filter the data before computing the width. Therefore in order to get the correct width, users will also have to apply the filter, which would also have to be exposed (if it isn't already).

rjmcneill:
Width seems to work ok for OHLC because the scales are bound and directly set.
For series that use multi however, it won't work as the scales are set after the multiseries is called.
The limitation is that for multiseries, you have to render once before calling the width function to get the correct value returned.

Couple of options:
- Have the simple implementation that needs to be rendered once. This works fine for bar, but for series like MACD, the scales get set after rendering, so this implementation won't work until after a render.
```JavaScript
    bar.width = function(data) {
        return calculateWidth(data);
    };

    function calculateWidth(data) {
        var filteredData = data.filter(base.defined);
        return barWidth(filteredData.map(crossAxisValueFunction()));
    }
```
- Ensure the scales and excessors are set before calling the width function.  
``` JavaScript
    macd.barWidth = function(data) {
        // Could also set the scales here too
        divergenceBar.xValue(xValue)
            .yValue(function(d, i) { return root(d).divergence; });
        return divergenceBar.width(data);
    };
```
- If the width isn't supposed to be part of the public API, leave it as is at the minute. I'll look into getting the width outside of d3fc by building up the functions externally.

[...]

PR #882, opened 2016-02-12
Added bar width getters to macd renderer and bar series

rjmcneill:
#881 

This addresses bar series and macd.
I'd be happy to add this to other stuff if you think this is the right approach.

### DIFF
--- a/src/series/bar.js
+++ b/src/series/bar.js
@@ -106,6 +106,17 @@ export default function() {
         });
     };
 
+    bar.width = function(data, orientation) {
+        if (orientation === 'vertical') {
+            return barWidth(data.map(function(d, i) {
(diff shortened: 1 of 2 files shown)


src/series/bar.js
```diff
@@ -106,6 +106,17 @@ export default function() {
         });
     };
 
+    bar.width = function(data, orientation) {
```
ColinEberhardt: With this change it means that the bar exposes a `width` function that has different logic for calculating the widths than it uses itself. The two really should be unified.

ColinEberhardt:
I'm happy with the general idea behind this PR, unification of the series interfaces makes sense, but the way that it is being proposed for bar looks a bit risky

PR #883, opened 2016-02-12
Exposed barWidth functions on elderRay, macd, and groupedBar

rjmcneill:
Relates to #881 

Added barWidth setters on elderRay and macd and added a groupWidth setter on groupedBar.
Completes first point on https://github.com/ScottLogic/d3fc/issues/881#issuecomment-183324776


ColinEberhardt:
nice one, thanks @rjmcneill

PR #878, opened 2016-02-10
Removed financial-oriented series default accessors

ColinEberhardt:
d3fc is not just for financial charting! For that reason I am proposing changing the accessor defaults for non financial-specific series types.

PR #884, opened 2016-02-13
Fix sparkline visual test - specify correct x-value accessor

jleft:


ColinEberhardt:
Thanks - i thought there was a good chance if miss something!

ColinEberhardt:
Just spotted, it's broken on the website too ...

ColinEberhardt:
See #878

PR #891, opened 2016-02-15
Invoke decorate in multi with selection, data and index

### LINKED ISSUE #890: Multi series decorate is only invoked with selection
[Multi's decorate is only invoked with a selection](https://github.com/ScottLogic/d3fc/blob/6fe6438c5090e0d6d349298a7ed294aefb8be4e3/src/series/multi.js#L62), and not with `data` and `index`.

This is inconsistent with the other series which are all invoked as `decorate(g, data, index)`. I'd suggest that multi should be changed to be consistent - unless I'm missing something?

jleft:
Resolves #890

### DIFF
--- a/src/series/multi.js
+++ b/src/series/multi.js
@@ -37,11 +37,11 @@ export default function() {
 
     var multi = function(selection) {
 
-        selection.each(function(data) {
+        selection.each(function(data, i) {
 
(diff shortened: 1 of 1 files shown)


ColinEberhardt:
Thanks @jleft - you're on a roll at the moment!

PR #892, opened 2016-02-16
Remove unnecessary render of OHLC series in Bollinger Bands visual test

jleft:
OHLC series is passed to multi


ColinEberhardt:
:+1:

PR #898, opened 2016-02-17
Added error check around crosshair rendering

### LINKED ISSUE #897: Snapping crosshair on series with an initial bucket can cause render failure
To recreate, apply the attached patch and load the visual test for macd.
If you move the crosshair to the far left of the graph, the console will show errors from crosshair render.

[Patch File.](https://github.com/ScottLogic/d3fc/files/134388/crosshair-bug.txt)

rjmcneill:
Closes #897

### DIFF
--- a/src/tool/crosshair.js
+++ b/src/tool/crosshair.js
@@ -85,7 +85,9 @@ export default function() {
                     return [this];
                 });
 
-            crosshairElement.call(multi);
+            if (data[0] && !isNaN(data[0].x) && !isNaN(data[0].y)) {
+                crosshairElement.call(multi);
+            }
 
             decorate(crosshairElement, data, index);
         });


src/tool/crosshair.js
```diff
@@ -85,7 +85,9 @@ export default function() {
                     return [this];
                 });
 
-            crosshairElement.call(multi);
+            if (data[0] && !isNaN(data[0].x) && !isNaN(data[0].y)) {
```
ColinEberhardt: Is the `data[0]` check necessary? The `crosshairElement` is a selection that has been generated by the data join here:

https://github.com/rjmcneill/d3fc/blob/bug/crosshair-snap/src/tool/crosshair.js#L75

With selections you shouldn't have to consider 'no data' as a special case.

ColinEberhardt:
Thanks for the bug report and proposed fix - I'm going to give you patch a try, I think this might need some further consideration.

rjmcneill:
I'm happy to try the patch @ColinEberhardt yea!
I'll not be about tomorrow, but I'll have some time on Monday to run a few tests for you.

ColinEberhardt:
Fixed by #903

PR #896, opened 2016-02-16
Added a hyperlink helper

ColinEberhardt:
This helper is used to ensure that internal links are correct, making it easier to re-structure the site within issues.

The basic concept is to replace manual links:

    [cartesian chart](/components/chart/cartesian.html)

With the following:

    {{{ hyperlink 'container.html' title='container component' }}}

This constructs the correct hyperlink by finding the page that has a path ending with `container.html`.

I found a few broken links when implementing this ;-)

I'll stop messing around with the site build soon ...

PR #902, opened 2016-02-17
Fix site broken links

jleft:
Spotted a few broken links.

(Okay, maybe a couple more small PRs before the more interesting ones!)


ColinEberhardt:
These would actually be fixed by #896 - I'd appreciate a review of that one if you've got the time?

jleft:
> These would actually be fixed by #896 - I'd appreciate a review of that one if you've got the time?

Ah, sorry, I didn't spot that. Yeah, sure :)

PR #903, opened 2016-02-19
Crosshairs now filter the data before applying the snapping function

### LINKED ISSUE #897: Snapping crosshair on series with an initial bucket can cause render failure

ColinEberhardt:
closes #897 #898 

@rjmcneill @jleft I think this is a better fix for #897 - the underlying issue is that the crosshair snap functions were not filtering undefined points. If a series has defined x values but undefined y values, and is snapping in the x direction, this results in the renderer being exposed to undefined values, causing the problems you observed.

### DIFF
--- a/src/util/minimum.js
+++ b/src/util/minimum.js
@@ -1,3 +1,4 @@
+// returns the item from the given array that returns the least value from the accessor function
 export default function minimum(data, accessor) {
     return data.map(function(dataPoint) {
         return [accessor(dataPoint), dataPoint];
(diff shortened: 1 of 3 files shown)


jleft:
Agreed, looks good - thanks! :+1:

rjmcneill:
Looks good to me Colin, thanks :+1:

PR #904, opened 2016-02-19
Fix site hyperlink helper on Windows

jleft:
As Windows uses backslashes for the directory separator in file paths (e.g. `\\examples\\simple\\index.html`), the hyperlink helper doesn't work as expected as it assumes the use `/`.

### DIFF
--- a/site/build/handlebars-helpers/hyperlink.js
+++ b/site/build/handlebars-helpers/hyperlink.js
@@ -1,9 +1,15 @@
+var url = require('url');
 
 function register(handlebars) {
     handlebars.registerHelper('hyperlink', function(link, context) {
         var pages = context.data.root.pages;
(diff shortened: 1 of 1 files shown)


ColinEberhardt:
Thanks for the fix @jleft - but I think the problem runs deeper ...

The `pageMetadata.page.destination` property comes from metadata added by [tiny-ssg](https://github.com/ColinEberhardt/tiny-ssg). I think a better fix would be to ensure that tiny-ssg creates paths that are not OS specific - and are use forward slashes (as per the URL specification)

jleft:
No problem. Sure, seems like a good idea. Feel free to close this :)

ColinEberhardt:
Thanks to @bdimitrov-SL - this has now been fixed :-)

PR #905, opened 2016-02-22
Fixed strategies so that they cope better with out of bounds data

ColinEberhardt:
The previous implementation where Infinity / MAX_VALUE was used for out of bounds rectangles was not working with the 'minimum' calculation which also uses Infinity

@WPFerg - would you be OK to review this PR?


WPFerg:
All looks good to me :+1:

ColinEberhardt:
Thanks @WPFerg

Issue #873, opened 2016-02-08
Releases break the website until cdnjs updates


ColinEberhardt:
When performing a release, the website is built, linking to the new version, and the package published to npm. All good.

However, it takes a while for updates to reach cdnjs, the docs quote ~12hrs

https://github.com/cdnjs/cdnjs/blob/master/documents/autoupdate.md

chrisprice:
What do we use cdnjs for? If it's just the quick start stuff, would a comment along those lines suffice?

chrisprice:
Implication being if it's anything else we shouldn't be using cdnjs...

ColinEberhardt:
CDNJS is used by the quick start and codepen. My plan is to resolve this by hosting our latest distribution builds on GitHub pages, using these for quick start and codepen, but also have a comment that points people at the cdn:

https://cdnjs.com/libraries/d3fc

chrisprice:
As of #906 we no longer directly reference the cdnjs hosted versions.

ColinEberhardt:
Another alternative, should we ever need it, is https://npmcdn.com/#/ - which updates more quickly than cdnjs.

PR #906, opened 2016-02-22
No longer use cdnjs for codepen

ColinEberhardt:
see #873

@chrisprice - this uses gh-pages for codepen and our getting started guide.


chrisprice:
Agree it's better than what we've got, at least codepen isn't borked when we're trying to promote releases!

PR #908, opened 2016-02-23
Fixes vertical baseline issues with Safari

### LINKED ISSUE #877: cartesian chart x-axis label cropped in Safari
screen shot 2016-02-10 at 22 12 42

ColinEberhardt:
Fixes #877 

Safari and Chrome disagree over ideographic and text-after-edge baselines.

This fix uses a 25% baseline shift, which positions the text above the baseline:

```
<svg width="1200">
  <line x1='0' y1='20' x2='1200' y2='20'
        stroke="#999" 
        stroke-width="0.5"/>
  <text y="20" baseline-shift="25%">This is a quick test</text>
</svg>
```

screen shot 2016-02-23 at 09 51 58

While this is not perfect (it assumes that the font extends beneath the baseline by no more than 25%), it is probably fine for most languages based around the roman alphabet.

### DIFF
--- a/src/chart/cartesian.js
+++ b/src/chart/cartesian.js
@@ -79,7 +79,7 @@ export default function(xScale, yScale) {
                 </g> \
                 <g class="x-axis label-container"> \
                     <g layout-style="height: 0; width: 0"> \
-                        <text class="label"/> \
+                        <text class="label" dy="-0.5em"/> \
                     </g> \
                 </g> \
                 <g class="y-axis label-container"> \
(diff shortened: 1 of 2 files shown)


ColinEberhardt:
@chrisprice - updated based on our discussion in slack.

From the commit message in https://github.com/ScottLogic/d3fc/commit/399e56ca8168ef790d9105b930081a3f4b927e5e:

--- 
Browsers have different interpretation of ideographic and text-after-edge dominant-baseline:

http://stackoverflow.com/questions/35565063/svg-dominant-baseline-not-working-in-safari

The fix is to use 'middle', then offset by 0.5 em to position the text so that it sits directly on the bottom boundary of the chart

PR #910, opened 2016-02-24
Rectangles layout refactor

ColinEberhardt:
I've been playing around with this layout component quite a bit. And as you can imagine, through experimentation I've found the need to modify / add / remove features.

I've used it to label bubble series, where overlapping labels are removed:

screen shot 2016-02-24 at 17 32 28

And for mapping, where you don't have scales:

screen shot 2016-02-24 at 17 32 55

Here's a summary of the changes:

- Removed the 'anchor' concept, the API was a bit odd, also, in practice you have access to this information already, all it was really doing was proving a simple coordinate system transformation.
- Removed the array functor, in most cases it is easier to provide a function that returns the array / tuple
- Changed `containerWidth` and `containerHeight` to a single `bounds` property - much more concise
- A new `removeCollisions` property that removes rectangles that have an overlap, this is performed iteratively in the order of decreasing overlap area.

TODO:
- [x] update the website docs


ColinEberhardt:
@WPFerg - it would be great if you could review this as the original author of this layout!

WPFerg:
Particularly like the removal of `anchor`, that never really sat right, and I'd never considered using a point series (obvious in hindsight!).

It all looks good, fairly straight forward fixes/additions

ColinEberhardt:
@chrisprice - I've made the algorithm that removed overlaps pluggable:

        var rectangles = fc.layout.rectangles(fc.layout.strategy.local())
            .size([10, 10])
            .filter(fc.layout.strategy.removeOverlaps())
            .position(function(d) { return [d.x, d.y]; });

This provides an extension point for adding a weight - either by providing a custom filter function, or we could eventually extend `removeOverlaps` to add weighting.

PR #913, opened 2016-02-25
fc.util.extent.include now accepts an array of values. Also updated t…

### LINKED ISSUE #911: extent only allows a single point
util.extent could maintain an _array_ of points which are always included in the extent.

Typically (and in my case) this will only contain 0 and a higher value required to give visual context to the graph.

```
#fc.util.extent.js
...
var includePoints = []
...
extent.include(singlePoint) -> add value to includePoints
extent.include([array of points]) -> append array to includePoints
extent.include() -> return includePoints
extent.include([]) -> empty includePoints
```

In the current implementation a second call to extent.include replaces the previous point.
This change would break extent if it is ever used with arrays as values.

jdhodges:
Closes #911.

### DIFF
--- a/src/util/extent.js
+++ b/src/util/extent.js
@@ -11,7 +11,7 @@ import d3 from 'd3';
 export default function() {
 
     var fields = [],
-        extraPoint = null,
+        extraPoints = [],
         padUnit = 'percent',
         pad = 0,
         symmetricalAbout = null;
@@ -90,13 +90,9 @@ export default function() {
             }
         }
 
-        // Include the specified point in the range
-        if (extraPoint !== null) {
-            if (extraPoint < min) {
-                min = extraPoint;
-            } else if (extraPoint > max) {
-                max = extraPoint;
-            }
+        if (extraPoints.length) {
+            min = Math.min(min, d3.min(extraPoints));
+            max = Math.max(max, d3.max(extraPoints));
         }
 
         if (dateExtent) {
@@ -128,9 +124,9 @@ export default function() {
 
     extents.include = function(x) {
         if (!arguments.length) {
-            return extraPoint;
+            return extraPoints;
         }
-        extraPoint = x;
+        extraPoints = x;
         return extents;
     };
 
--- a/tests/util/extentSpec.js
+++ b/tests/util/extentSpec.js
@@ -61,7 +61,7 @@ describe('fc.util.extent', function() {
 
         var extents = fc.util.extent()
             .fields('high')
-            .include(10)(data);
+            .include([10])(data);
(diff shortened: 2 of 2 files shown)


src/util/extent.js
```diff
@@ -128,9 +128,9 @@ export default function() {
 
     extents.include = function(x) {
         if (!arguments.length) {
-            return extraPoint;
+            return extraPoints;
         }
-        extraPoint = x;
+        extraPoints = x;
```
bjedrzejewski: Do you want to add to already existing points, or override?

bjedrzejewski: Something more like `extraPoints = extraPoints.concat(x);`

jdhodges: @chrisprice mentioned this earlier https://github.com/ScottLogic/d3fc/issues/911#issuecomment-188778371

> If a value is stored internally as an array then I think it should be set as an array.

src/util/extent.js
```diff
@@ -90,14 +90,14 @@ export default function() {
...
         }
 
-        // Include the specified point in the range
-        if (extraPoint !== null) {
-            if (extraPoint < min) {
-                min = extraPoint;
-            } else if (extraPoint > max) {
-                max = extraPoint;
+        // Include the specified point(s) in the range
+        extraPoints.forEach(function(point) {
```
chrisprice: Suggestion -
```js
min = Math.min(min, d3.min(extraPoints));
max = Math.max(max, d3.max(extraPoints));
```

src/util/extent.js
bjedrzejewski: I thought this would make the interface even smoother, if you are calling include twice on two separate arrays.

jdhodges: Yeah, that would be nice - just wondering if there'd be a scenario where you'd want to empty it though, this was mentioned in the original comment in https://github.com/ScottLogic/d3fc/issues/911#issue-136182304

jdhodges: I can add that behaviour in too perhaps?

chrisprice: I don't think any of the components do or should support additive operations. It used to be the case that data.random.financial had mutable internal state but that's now been encapsulated in the stream instance.  

Can you give a common example where it would be useful to have the additive support?

ColinEberhardt:
This LGTM - @chrisprice are you happy to merge?

chrisprice:
Thanks @jdhodges

ColinEberhardt:
Just realised, this PR did not include the website changes. @jdhodges - any chance you could make he minor tweaks required to this page:

https://d3fc.io/components/util/extent.html

PR #916, opened 2016-02-26
fc.util.extent.include documentation updated.

jdhodges:
https://github.com/ScottLogic/d3fc/pull/913#issuecomment-189154461


ColinEberhardt:
Thanks for the docs :-)

PR #919, opened 2016-02-29
Replaced harcoded svg paths with d3-path (v4)

alisd23:
We should take advantage of the new d3-path module in d3 version 4, instead of hardcoding paths. Part of the modularisation of d3fc.


src/svg/bar.js
```diff
@@ -50,17 +51,14 @@ export default function() {
...
+            var barPath = path();
+            barPath.rect(
+                xValue - horizontalOffset,
+                yValue - verticalOffset,
+                barWidth,
+                barHeight
+            );
+            return barPath.toString();
         }, this)
         .join('');
```
chrisprice: Does this duplicate the functionality of path.toString?

tests/annotation/bandSpec.js
```diff
@@ -21,7 +21,7 @@ describe('band', function() {
             .call(band);
 
         var path = element.children[0].children[0];
-        expect(path.getAttribute('d')).toEqual('M0,0h20v20h-20z');
+        expect(path.getAttribute('d')).toEqual('M0,0h20v20h-20Z');
```
ColinEberhardt: I looked up the difference between `Z` and `z`, MDN says:

> Apparently there is no difference between the uppercase and lowercase command.

I love the lack of certainty!

package.json
```diff
@@ -8,7 +8,9 @@
   "dependencies": {
     "css-layout": "^1.1.0",
     "d3": "^3.5.4",
+    "d3-path": "^0.1.5",
```
ColinEberhardt: Do we need to depend on `d3-path`, isn't the transitive dependency via `d3fc-path` sufficient?

alisd23: **d3fc-path** only depends on **d3-path** as a dev dependency (used for testing). **d3fc-path** just takes in the context you give it, which from the **d3fc/svg** stuff is **d3-path**.
Wow that's a lot of paths...

chrisprice: I think it's correct as a devdep.

We use an interface that both d3-path and Context2D implement. Therefore we
don't have a direct dependency on either. However an easy way to write the
tests is to use d3 -path to serialise to an svg path string for assertions.

Let me know if that's still not clear.

ColinEberhardt: No, that makes perfect sense. Thanks for the explanation.

src/svg/bar.js
```diff
@@ -1,112 +1,13 @@
-import d3 from 'd3';
+import { rebindAll } from '../util/rebind';
```
chrisprice: Is this import still required?

src/svg/bar.js
```diff
@@ -1,112 +1,13 @@
...
+import { path as svgPath } from 'd3-path';
+import { bar as barPath } from 'd3fc-path';
 
-// Renders a bar series as an SVG path based on the given array of datapoints. Each
-// bar has a fixed width, whilst the x, y and height are obtained from each data
-// point via the supplied accessor functions.
+// Renders an OHLC as an SVG path based on the given array of datapoints. Each
+// OHLC has a fixed width, whilst the x, open, high, low and close positions are
+// obtained from each point via the supplied accessor functions.
 export default function() {
```
chrisprice: I think this is one of the rare occasions that expressing it as a one-liner is actually simpler to understand? E.g.
```js
return barPath(svgPath)
```

chrisprice:
Outstanding -
* [ ] Address [comment](https://github.com/ScottLogic/d3fc/pull/919#discussion_r54857784)
* [ ] Update dependency to npm module

Once these are done, LGTM

chrisprice:
@alisd23 Build's failing :disappointed:

PR #920, opened 2016-02-29
Cleared RSI calculator state after each use. Added test to verify fix.

### LINKED ISSUE #917: Subsequent calls to RSI calculator causing erroneous output.
https://github.com/ScottLogic/BitFlux/issues/664

The RSI calculator holds on to some internal state that is being retained and used in subsequent calls.

Currently working on a fix.

jdhodges:
Fixes #917.

### DIFF
--- a/src/indicator/algorithm/calculator/relativeStrengthIndex.js
+++ b/src/indicator/algorithm/calculator/relativeStrengthIndex.js
@@ -3,53 +3,52 @@ import _slidingWindow from './slidingWindow';
 
 export default function() {
 
-    var closeValue = function(d, i) { return d.close; },
+    var windowSize = 14,
+        closeValue = function(d, i) { return d.close; },
         wildersSmoothing = function(values, prevAvg) {
-            var result = prevAvg + ((values[values.length - 1] - prevAvg) / values.length);
-            return result;
+            return prevAvg + ((values[values.length - 1] - prevAvg) / values.length);
         },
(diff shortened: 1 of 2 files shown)


ColinEberhardt:
Thanks :-) - are there any other indicators that suffer from this issue?

jdhodges:
No problem :) None of the other indicators seem to have this problem so we should be OK. I'll come back with a more concrete answer when I have run some more tests.

I think I'm stretching the capabilities of the sliding window with the current implementation. I'm thinking a better option would be similar to how the EMA calculator works at the moment (since it needs to remember previous values). I'll push the new implementation now.

ColinEberhardt:
Actually, isn't there a simpler way of removing the retained state - by changing its scope so that it is restricted to the component function itself ...

Regarding the change you've made in this commit:

https://github.com/jdhodges/d3fc/commit/ad9a2c4c0b79585f68498b8ea78c4c134622ed01

How about changing it so that the sliding window component is created within the rsi function scope as follows:


```
var rsi = function(data) {

  var prevClose, prevDownChangesAvg, prevUpChangesAvg;

  var slidingWindow = _slidingWindow()
         .windowSize(14)
         .accumulator(function(values) {
             ...
         });

  return slidingWindow(data);
}
```

jdhodges:
Aha! That's what I'm looking for. Cheers Colin.

ColinEberhardt:
That's better - thanks for the fix.

PR #921, opened 2016-02-29
[WIP] measureText component

ColinEberhardt:
@chrisprice this is what I have in mind for the two components. You can see how they are used here:

https://github.com/ColinEberhardt/d3-financial-components/blob/dbad6bf721264eb5f1e23951093e4b4c778c4a84/site/src/components/layout/rectangles-example-measure.js#L115

TODO:
 - [ ] unit / visual test if appropriate
 - [ ] website documentation
 - [ ] decorate?


chrisprice:
(On phone)
In the margin expansion top is assigned a value based on bottom, is that
correct?
There seems to be some (functionality) overlap with container, is it worth
trying to decorate one to create the label?
Is there a use case for margin on measure text?
At a guess i would say that appending a label, measuring it and then
removing it many times is less efficient than adding a bunch, measuring
them all and then removing them all. Is it worth a quick profile? It might
even be beneficial to cache the temporary elements...

ColinEberhardt:
I like the idea of decorating layout, I wasn't too keen on the dependency between the label component and flexbox.

ColinEberhardt:
Actually, I think you're suggesting something different, by its given me an idea all the same ...

ColinEberhardt:
@chrisprice - I've removed the label component for now, you're right, container works well for this. 

Regarding performance, on a map with ~40 labels measureText was around 32ms. I modified measureText to cache the svg and text elements, removing that via setTimout, so that they are only added / removed once. This brought the time down marginally to 25ms. Not really worth it in my opinion (I am guessing a setTimeout(removeSVG, 0) would actually result in the SVG being rendered?)

I am sure it would be faster to perform a bulk measure, i.e. add a bunch of text, then measure them all, as the timeline currently shows many reflows. However, this would require an asynchronous API, for batching, or a change in the rectangles layout API in order to request all the sizes in one go.

chrisprice:
Glad at least part of it went to plan!

I would expect the timeout value to be insignificant for any value of 16
and below (including 0), at 25 you're probably making it slightly slower
than it needs to be but I'm guessing the reflow it's still the bottleneck.

While we're trying other things, what about using canvas for this?
https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/measureText

ColinEberhardt:
replaced by #924

PR #924, opened 2016-03-07
Rectangles includes a measure then layout pass

ColinEberhardt:
@chrisprice - this seems to work quite nicely, by having `rectangles` render the component before invoking `size` and `position` you can measure the added elements directly. I like the way this gives a two pass rendering, measure then layout.

TODO:
- [x] update docs


src/layout/rectangles.js
```diff
@@ -33,10 +32,15 @@ export default function(layoutStrategy) {
         }
 
         selection.each(function(data, index) {
+
+            var g = dataJoin(this, data);
+            g.call(component);
```
chrisprice: Could chain onto previous line?

src/layout/rectangles.js
```diff
@@ -58,7 +62,7 @@ export default function(layoutStrategy) {
             // filter the data to include only those remaining after collision removal
             var filteredData = filteredLayout.map(function(d) { return data[d.dataIndex]; });
 
-            var g = dataJoin(this, filteredData);
+            g = dataJoin(this, filteredData);
```
chrisprice: If the filtering only affected `display`/`visibility`/`opacity` of the rectangle, then I don't think there wouldn't be a need for a second join. Do you think that would be a reasonable compromise?

src/layout/rectangles.js
```diff
@@ -58,7 +62,7 @@ export default function(layoutStrategy) {
             // filter the data to include only those remaining after collision removal
             var filteredData = filteredLayout.map(function(d) { return data[d.dataIndex]; });
```
chrisprice: I know this is pre-existing but adding the indicies seems a bit weird.

src/layout/rectangles.js
```diff
@@ -33,10 +32,15 @@ export default function(layoutStrategy) {
...
         selection.each(function(data, index) {
+
+            var g = dataJoin(this, data);
+            g.call(component);
+
             // obtain the rectangular bounding boxes for each child
-            var childRects = data.map(function(d, i) {
-                var childPos = position(d, i);
-                var childSize = size(d, i);
+            var childRects = g[0].map(function(node, i) {
```
chrisprice: Is this not the equivalent of an `each`?

chrisprice: No it's not...

chrisprice: It's just the [0] looks odd

site/src/components/layout/rectangles-example-measure.js
```diff
@@ -112,35 +112,20 @@ var data = [
...
-            .attr({
-                // vertically centre the text in the container
-                'dominant-baseline': 'middle',
-                'dy': '-0.5em'
-            });
-    selection.layout();
-}
+var label = fc.tool.container()
+    .padding(labelMargin)
+    .component(function(sel) {
```
chrisprice: Did you try decorate instead of component?

ColinEberhardt: Yes, I tried decorate - the implementation is pretty similar, although you are piggy-backing the container components data-join, which makes it a bit simpler. However, by using decorate you are becoming tightly coupled to the container components internal structure, you have to know the correct selector to place the `text` in the correct location. This implementation, although slightly more complex, felt more correct.

src/layout/rectangles.js
ColinEberhardt: That's a good point, I was meaning to re-visit this. I can't think of a better way to map a d3 selection to an array. Have you seen any alternatives?

src/layout/rectangles.js
ColinEberhardt: That's a good idea, by hiding filtered labels I can avoid a second join - this also means that components that are rendered via this layout can be a little simpler, i.e. they don't need to respect enter / update.

site/src/components/layout/rectangles-example-measure.js
chrisprice: It seems clearer to me, unless I'm missing something?
```js
.decorate(function(g) {
  g.enter()
    .append('text')
    .attr('dy', '0.7em')
    .text(function(d) { return d.language; });
})
```

ColinEberhardt: You're missing one small step ...


```
.decorate(function(g) {
  g.enter()
    .select('g')
    .append('text')
    .attr('dy', '0.7em')
    .text(function(d) { return d.language; });
})
```

You need to select the inner `g` element that has the padding. It's a small thing, but couples this code to the internal structure.

chrisprice: Ah fair enough.

src/layout/rectangles.js
chrisprice: I don't think I have. I suppose in d3 it's unusual to want to do a projection of nodes into data, it's normally the other way around.

ColinEberhardt:
@chrisprice I've updated the component to hide rectangles when filtered, rather than perform the data join twice. Docs updates.

site/src/components/layout/rectangles-example-measure.js
```diff
@@ -158,7 +140,14 @@ var chart = fc.chart.cartesian(
 
 // create the layout that positions the labels
 var labels = fc.layout.rectangles(fc.layout.strategy.greedy())
-        .size(function(d) { return boundingBox(d.language); })
+        .size(function(d) {
+            // measure the label and add the required padding
+            var textSize = d3.select(this)
+                    .select('text')
```
chrisprice: Indentation

src/layout/rectangles.js
```diff
@@ -49,33 +53,34 @@ export default function(layoutStrategy) {
             // or number of rectangles.
             var layout = strategy(childRects);
 
-            // add indices to the layouts
+            // add indices to the layouts - this is require in order to determine which
```
chrisprice: After discussion -
* [ ] add optional `hidden` flag to objects returned from layout
* [ ] remove `filter` property
* [ ] consolidate `g.attr` calls

ColinEberhardt:
@chrisprice - I've made those changes. I really like the removal of `filter`, which means `removeOverlaps` is a strategy like any other, which means it can now be used on its own:

https://github.com/ScottLogic/d3fc/commit/88dff166cd8e77dce4189f670e6850cfccc40551#diff-dc4b01aaa354fb8a4d962bee3a6d280dR27

chrisprice:
One unrelated comment but otherwise now LGTM.

PR #929, opened 2016-03-10
Remove unbundled from quick start

chrisprice:
As we break up d3fc the unbundled section is going to grow beyond a useful size. It is also implicitly referencing the old npm folder layout.


ColinEberhardt:
:+1: - I guess originally we didn't add the bundle to npm?

PR #930, opened 2016-03-10
Extent no longer mutates the fields property when used

### LINKED ISSUE #813: extent - mutates the fields property when used
If fields is specified as an array of strings, after the first invocation the array will be populated with functions.

jdhodges:
Closes #813.

### DIFF
--- a/site/src/components/annotation/line-example.js
+++ b/site/src/components/annotation/line-example.js
@@ -8,7 +8,7 @@ var dataGenerator = fc.data.random.financial()
 var data = dataGenerator(50);
 
 var xScale = fc.scale.dateTime()
-    .domain(fc.util.extent().fields('date')(data))
+    .domain(fc.util.extent().fields(['date'])(data))
     .range([0, width]);
 
 var yScale = d3.scale.linear()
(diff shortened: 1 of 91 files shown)


chrisprice:
One more, nice one!

chrisprice:
Just waiting on @alisd23 to fix the build... :trollface:

ColinEberhardt:
Thanks @jdhodges - while you are fixing this, I wonder if it is worth addressing the other side-effect in the `fields` property setter:

https://github.com/jdhodges/d3fc/blob/bug/extent-mutates-fields-when-used/src/util/extent.js#L117

We've been actively moving away from this pattern. As a result, I think `fields` should always expect to be set with an array.

jdhodges:
No problemo, I'll push the changes soon.

jdhodges:
Looks like we don't need to make any changes to the website for this.

ColinEberhardt:
Thanks for the update @jdhodges - although I think there are changes required for both the website and visual tests ...

screen shot 2016-03-11 at 07 22 48

Hopefully a regex find and replace can easily update all these instances :-)

jdhodges:
> 227 results

A _slight_ oversight on my part. I have made the necessary changes :smiley:

ColinEberhardt:
Nice work. Our CI checks for console log errors on both the website and visual tests, which should 'protect' us if this does introduce any errors.

PR #932, opened 2016-03-10
Generalised the grouped bar to grouped series

### LINKED ISSUE #915: groupedBar series should allow grouping of other series types
See this example:

This should be a pretty trivial change.

rjmcneill:
Closes #915 

There might be a few too many examples for the visual tests, but I wrote them while testing the different scenarios so thought I might as well let you decide if you want them.


src/series/groupedSeries.js
```diff
@@ -1,31 +1,31 @@
...
-    var bar = barSeries(),
-        barWidth = fractionalBarWidth(0.75),
+    var barWidth = fractionalBarWidth(0.75),
         decorate = noop,
         xScale = d3.scale.linear(),
         offsetScale = d3.scale.linear(),
-        values = function(d) { return d.values; };
+        values = function(d) { return d.values; },
+        x = function(d, i) { return xScale(value(d, i)); },
+        value = function(d) { return d.x; };
```
ColinEberhardt: I don't think you should be introducing a new accessor for the x-value. Have you tried re-binding the `xValue` property of the series that this adapts?

src/series/groupedSeries.js
```diff
@@ -1,31 +1,31 @@
...
 
-    var bar = barSeries(),
-        barWidth = fractionalBarWidth(0.75),
+    var barWidth = fractionalBarWidth(0.75),
         decorate = noop,
         xScale = d3.scale.linear(),
         offsetScale = d3.scale.linear(),
-        values = function(d) { return d.values; };
+        values = function(d) { return d.values; },
+        x = function(d, i) { return xScale(value(d, i)); },
```
ColinEberhardt: As the `x` function is only used once, I do wonder about the value of having it pulled up into this outer scope? Typically the variables found here are exposed as properties.

src/series/groupedSeries.js
```diff
@@ -1,31 +1,31 @@
-import barSeries from './bar';
 import d3 from 'd3';
```
ColinEberhardt: Can we call this `grouped` rather than `groupedSeries` - nice and concise ;-)

src/series/groupedSeries.js
```diff
@@ -37,55 +37,69 @@ export default function() {
 
             g.select('g')
                 .datum(values)
-                .each(function(series, index) {
+                .each(function(_series, index) {
```
ColinEberhardt: A common naming convention for function arguments that are not used is simply underscore, `_`

src/series/groupedSeries.js
```diff
@@ -37,55 +37,69 @@ export default function() {
...
+    groupedSeries.offsetScale = function(_x) {
+        if (!arguments.length) {
+            return offsetScale;
+        }
+        offsetScale = _x;
+        return groupedSeries;
     };
 
-    d3.rebind(groupedBar, bar, 'yValue', 'xValue', 'yScale');
+    d3.rebind(groupedSeries, series, 'yScale');
```
ColinEberhardt: How about using `rebindAll` that way it will dynamically rebind the series properties. Not sure if this might cause name collisions? Definitely worth considering.

visual-tests/index.html
```diff
@@ -200,9 +200,9 @@
               <p>Data is loaded from a CSV file.</p>
             </article>
 
-            <article data-href="series/grouped-bar.html">
-              <h3>Grouped Bar</h3>
-              <p>It's a bar ... but grouped. Oh yes.</p>
+            <article data-href="series/grouped-series.html">
+              <h3>Grouped Series</h3>
+              <p>It's a series ... but grouped. Oh yes.</p>
```
ColinEberhardt: Very thorough ;-)

ColinEberhardt:
Nice job, and thanks for the thorough testing. It certainly makes this series type more versatile and useful.

When the review comments have been addressed it is probably worth trimming the visual tests a little. Does it add much value to demonstrate on so many series types? I think as long as you include one with a bar-width and one without, that would be fine. Also, please try to keep it within the 'frame' that the index page constructs. You might want to consider adding a simple set of radio controls to select series types. See the other visual tests for examples.

Anyhow, this change looks almost complete, a few tweaks, then update the website docs and its done :-)

src/series/groupedSeries.js
```diff
@@ -37,55 +37,69 @@ export default function() {
...
+        return groupedSeries;
+    };
+    groupedSeries.value = function(_x) {
+        if (!arguments.length) {
+            return value;
+        }
+        value = _x;
+        return groupedSeries;
+    };
+    groupedSeries.offsetScale = function(_x) {
```
ColinEberhardt: I'm not sure there is any value in exposing this scale? The example in the visual tests produces an effect that looks no different from rendering two candlestick series via a multi-series.

Although if you can find a use-case, I'm happy to keep it :-)

src/series/groupedSeries.js
rjmcneill: I had that originally, but for the candlestick series that returns ```d.date``` so it doesn't work.
I could reduce the scope of it though and remove the accessor.  I haven't changed it in the examples.

ColinEberhardt: I *think* the use of `rebindAll` would solve this?

rjmcneill:
Website needs updating still, but I'll get that sorted on Monday!

ColinEberhardt:
Sorry @rjmcneill - I'd missed that you've pushed some updates. You don't get any notification on PR updated other than comments. Is this ready for another review?

rjmcneill:
@ColinEberhardt Sorry, I should have commented!
Yup, ready for review now

ColinEberhardt:
merged via #943 (updating to include the new d3fc-rebind functions) - many thank @rjmcneill for all your hard work :-)

PR #939, opened 2016-03-15
refactor: replace util/rebind with d3fc/d3fc-rebind

chrisprice:
I've intentionally not replaced any instances of d3.rebind, that feels like an upgrade to v4 task. 

Here's the link to [`d3fc-rebind`](https://github.com/d3fc/d3fc-rebind) for reference.


src/chart/cartesian.js
```diff
@@ -177,36 +177,30 @@ export default function(xScale, yScale) {
...
-            rebind(cartesian, scale, mappings);
-        }
-
-        rebindAll(cartesian, scale, prefix, scaleExclusions);
+    var scaleExclusions = exclude(
+        /range\w*/,   // the scale range is set via the component layout
+        /tickFormat/  // use axis.tickFormat instead (only present on linear scales)
+    );
+    rebindAll(cartesian, xScale, scaleExclusions, prefix('x'));
+    rebindAll(cartesian, yScale, scaleExclusions, prefix('y'));
```
ColinEberhardt: I think the above `rebindAll` will result in the `ticks` method being rebounds for ordinal and linear scales. The current implementation performs a `scaleExclusions.push('ticks');` in order to exclude the `ticks` method from being rebound for ordinal scales.

src/chart/sparkline.js
```diff
@@ -91,16 +91,9 @@ export default function() {
...
-        xDiscontinuityProvider: 'discontinuityProvider',
-        xDomain: 'domain'
-    });
-
-    rebind(sparkline, yScale, {
-        yDomain: 'domain'
-    });
-
-    rebind(sparkline, line, 'xValue', 'yValue');
+    rebindAll(sparkline, xScale, include('discontinuityProvider', 'domain'), prefix('x'));
```
ColinEberhardt: Nice, I like the way that include and prefix combine here.

src/indicator/algorithm/calculator/stochasticOscillator.js
```diff
@@ -56,13 +56,8 @@ export default function() {
...
 
-    rebind(stochastic, kWindow, {
-        kWindowSize: 'windowSize'
-    });
-
-    rebind(stochastic, dWindow, {
-        dWindowSize: 'windowSize'
-    });
+    rebindAll(stochastic, kWindow, map({'windowSize': 'kWindowSize'}));
+    rebindAll(stochastic, dWindow, map({'windowSize': 'dWindowSize'}));
```
ColinEberhardt: `kWindow` and `dWindow` both have an `accumulator` property, now that you `rebindAll` for both, what happens with this collision? (Similar comment for other components in this namespace)

ColinEberhardt: Actually, I presume that a `map` is an implicit `include`? That's not immediately obvious.

ColinEberhardt:
Looks really cool, a couple of comments here and PR / issue on `d3fc-rebind`

src/chart/cartesian.js
chrisprice: I hadn't spotted that but I think it's still ok....

The check is `!isOrdinal(scale)` which results in `ticks` being excluded i.e. ordinal scales do get `[x/y]Ticks` but non-ordinal ones do not. 

With this change, it is correct that as of this line, `cartesian` has an `[x/y]Ticks` method irregardless of axis type. However I think, as the order of calls has been reversed, this method is then overwritten by the axis rebind.

src/indicator/algorithm/calculator/stochasticOscillator.js
chrisprice: That's correct, `map` has an implicit `include` and from what I've seen in d3fc, that's what's required. If I remove the implicit `include` I reckon there'll be a lot of duplication of property names. Do you think more of a warning in the docs would help or maybe a rename to `onlyMap`/`includeMap` or something else?

src/chart/cartesian.js
ColinEberhardt: From reviewing the code, I think you're right, the net effect is the same. However, it would be all to easy to move the order of the methods around and change this behaviour.

src/indicator/algorithm/calculator/stochasticOscillator.js
ColinEberhardt: I like `includeMap` - much more obvious what it does.

src/chart/cartesian.js
chrisprice: That's a fair point. As the new exclude function does not validate the parameters before excluding (it no longer can), we can safely always exclude `ticks`.

chrisprice:
Changes pushed, please can you review the `exclude('ticks')` bit?

ColinEberhardt:
:+1:

PR #942, opened 2016-03-16
feat: added integration with d3fc-label-layout

ColinEberhardt:
BREAKING CHANGE: this removes the 'rectangles' layout component replacing it with the equivalent 'label' layout component from d3fc-label-layout

This integrated the  d3fc-label-layout module using our existing d3fc.io website structure. As we break things up we are going to have to review the 'component' navigation again. We could do something similar to assemble (http://assemble.io) (sorry for mentioning that SSG again!), where it shows a brief summary 'card' of the various sub-projects.


site/src/components/layout/rectangles-example-measure.js
```diff
@@ -114,15 +114,10 @@ var data = [
 //START
 var labelPadding = 2;
 
-// the container component is used to add padding around a text label
-var label = fc.tool.container()
+// a component which renders the labels
+var label = fc.layout.textLabel()
```
chrisprice: The naming seems forced to fit in with `fc.layout.label`, not that I can think of a better name...

src/layout/label.js
```diff
@@ -0,0 +1,46 @@
+import { range } from '../util/scale';
+import { rebindAll } from 'd3fc-rebind';
+import { label as labelLayout } from 'd3fc-label-layout';
+
+export default function(strategy) {
+
+    var adaptee = labelLayout(strategy);
+
+    var xScale = null,
```
chrisprice: Normally we'd default the scales to e.g. `d3.scale.identitiy`.

src/layout/label.js
```diff
@@ -0,0 +1,46 @@
...
+        yScale = null;
+
+
+    var label = function(selection) {
+        // automatically set the bounds of the strategy based on the scale range
+        if (strategy && strategy.bounds && xScale && yScale) {
```
chrisprice: Does it ever make sense for this component to be used without setting scales?

src/layout/label.js
```diff
@@ -0,0 +1,46 @@
...
+            var xRange = range(xScale),
+                yRange = range(yScale);
+            strategy.bounds([
```
chrisprice: Just noticed that bounds only accepts two arguments, I assume that means it's assuming zero as the origin? Maybe that should be changed (separately!)

site/src/components/layout/rectangles-example-measure.js
ColinEberhardt: Yeah, I know - I didn't like it much either. I find myself writing code like this:

```
var labelLayout = fc.layout.label()
var label = fc.layout.textLabel()
```

Feels a bit confused.

chrisprice:
A few minor comments but LGTM. Nice to see so much code deleted (or moved)!

src/layout/label.js
ColinEberhardt: That's a good question, and why I defaulted to `null` above. I think the answer is 'no', it doesn't make sense to use it without scale - in that case you just use the non-adapted layout. I'll simplify this code a little ...

ColinEberhardt:
Thanks for the feedback - I've addressed these comments.

PR #951, opened 2016-03-29
Include bundled files in GitHub Releases

### LINKED ISSUE #950: GitHub releases do not include dist bundled files
I think it would be useful to also include the bundled d3fc files in GitHub releases (assuming GitHub releases is still relevant?).

Currently, only the following files are including:

GitHub Releases

jleft:
Resolves #950


ColinEberhardt:
Thanks @jleft

PR #955, opened 2016-04-13
fix: axis tick text positioning in IE

### LINKED ISSUE #954: Axis tick text placement is incorrect in IE
I believe this a problem with `dominant-baseline` being interpreted differently in different browsers (similar to #877).

Example (I think Chrome looks right and IE does not):

| IE 11  | Chrome |
| ------------- | ------------- |
| axis-broken-ie | axis-original-chrome |

jleft:
Fixes #954.

This is a similar fix to that in #908 - using `dy` to align the text elements, rather than `dominant-baseline`. It's also consistent with how [D3's axis component positions the tick text elements](https://github.com/mbostock/d3/blob/c36befc7361cd61538294048c387a8a407b65dd1/src/svg/axis.js#L54-L62).

This is how it looks with the fix (see #954 to see the original problem):

| IE 11  | Chrome |
| ------------- | ------------- |
| axis-fixed-ie | axis-fixed-chrome |

### DIFF
--- a/src/svg/axis.js
+++ b/src/svg/axis.js
@@ -123,6 +123,15 @@ export default function() {
 
             g.select('text')
                .attr('transform', translate(0, labelOffset))
+               .attr('dy', function() {
+                   var offset = '0em';
+                   if (isVertical()) {
+                       offset = '0.32em';
+                   } else if (orient === 'bottom') {
+                       offset = '0.71em';
+                   }
+                   return offset;
+               })
                .text(tickFormatter);
 
             // exit - for non ordinal scales, exit by animating the tick to its new location
--- a/src/fc.css
+++ b/src/fc.css
@@ -106,21 +106,12 @@ text {
   text-anchor: middle;
 }
 
-.tick.orient-left text,
-.tick.orient-right text {
-  dominant-baseline: middle;
-}
-
 .tick.orient-left text {
   text-anchor: end;
 }
 
 .tick.orient-right text {
   text-anchor: start;
-}
-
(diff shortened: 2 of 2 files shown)


src/svg/axis.js
```diff
@@ -123,6 +123,15 @@ export default function() {
 
             g.select('text')
                .attr('transform', translate(0, labelOffset))
+               .attr('dy', function() {
+                   var offset = '0em';
```
chrisprice: Normally I'm against any magic numbers but as we've gone round on this multiple times I'm willing to let it pass in this instance. Before I do though, can I check where these numbers came from because they don't appear to be symmetrical  (e.g. 0.32/0.68)?

jleft: The same values are also used by D3's axis ([`.71em`](https://github.com/mbostock/d3/blob/c36befc7361cd61538294048c387a8a407b65dd1/src/svg/axis.js#L56) and [`.32em`](https://github.com/mbostock/d3/blob/c36befc7361cd61538294048c387a8a407b65dd1/src/svg/axis.js#L60)). As D3 uses them, I thought it was a fairly safe option!


Also, through trial and error, I tried various combinations of `dominant-baseline` and `dy` values, and these seem to work.

For example, in Chrome, with the axis orient set to bottom: `dominant-baseline: auto` and `dy="0.71em"` renders the same as `dominant-baseline: hanging` and `dy="0em"` (`dy="0.68em"` was slightly too close to the axis).

chrisprice:
Can't argue with that, thanks @jleft

PR #957, opened 2016-04-26
feat: add ability to set small multiples domain for each row or column

### LINKED ISSUE #947: Small multiples chart doesn't allow different y axis domains
Our small multiple uses x and y scale for each chart. However, in some examples different domains are required for each multiple. See the following example:

http://bl.ocks.org/mbostock/1157787

jdhodges:
Closes #947


ColinEberhardt:
Thanks for this pull request, really appreciate it. Here's some general bit of feedback:

 1. I'm not keen on having the `xDomainAccessor` property as well as `xDomain` (and the same on y of course!). Could you make it so that the `xDomain` property can either accept a regular domain (i.e. an array of values), or a function which returns a domain? You'd have to modify your code so that domain is no longer re-bound from the underlying scale.
 2. I'm not keen on the logic that creates component-scoped state variables `xDomains` / `yDomains`, which are then use by component-scoped functions. If any temporary state is required, my feeling is it should be scoped to the rendering function. Also, do you need to cache the domains? the domain accessors described in (1) will likely have very simple switch / case type logic, so I don't think you need to cache their output.

Finally, the example feels a bit weak, it doesn't look terribly different from the one above it! What about re-creating this one:

http://bl.ocks.org/mbostock/1157787

If you want to discuss any of the above, give me a shout. Definitely moving in the right direction.

jdhodges:
Cheers Colin, I've removed the `xDomains` and `yDomains` from the component scope and updated the example. I'm unsure how I can avoid caching the domains since they're being used in a couple of different places (to generate each multiple and each axis).

ColinEberhardt:
Cool - thanks for the update. I'll take a look at this tomorrow :-)

ColinEberhardt:
Thanks for the updates, looking good :-)

I've made some additional changes in #963. Rather than caching the domains, I've made it easier to obtain the domain for a specific cell, allowing the removal of transient state.

PR #960, opened 2016-04-28
fix: correct y domain calculation for small multiples visual test

### LINKED ISSUE #959: Small multiples visual test y domain doesn't take into account both series

jdhodges:
Fixes #959


ColinEberhardt:
👍

PR #963, opened 2016-05-03
Updates to #957

ColinEberhardt:
A small refactor, removing the caching of domains in favour of making it simple to obtain the domain for a specific cell


ColinEberhardt:
See #957

PR #966, opened 2016-05-11
refactor: import Quandl and Coinbase feeds from d3fc-financial-feed

tlsim:


src/data/data.js
```diff
@@ -1,10 +1,10 @@
-import feed from './feed/feed';
+import { coinbase, quandl } from 'd3fc-financial-feed';
```
chrisprice: Not a big deal but I think you can `import * as feed from 'd3fc-financial-feed'`

ColinEberhardt:
This looks good to me :-)

Although I think the website should be updated as part of this change, removing the information which is how duplicated.

tlsim:
@ColinEberhardt site now updated.

ColinEberhardt:
Nice job ... +10 / -325 good diff statistics :-)

PR #972, opened 2016-05-24
docs: fix site index page streaming chart resize

### LINKED ISSUE #970: Site index page streaming chart breaks on resize
Intervals created in `renderStreaming` are not cleared, so the chart is repeatedly redrawn using old data.

tlsim:
closes #970


chrisprice:
Any chance we could move all the interval code into the `index.html` rather than having it split across files? I assume the split across the files is what caused the mismatch in the first place.

chrisprice:
Superseded by #974

PR #956, opened 2016-04-25
feat: allow functional usage of merge

jdhodges:
Closes #814

PR #973, opened 2016-05-24
Integrate d3fc-technical-indicator

tlsim:


src/indicator/algorithm/bollingerBands.js
```diff
@@ -19,8 +19,8 @@ export default function() {
         return d.bollingerBands;
     };
 
-    d3.rebind(bollingerBands, mergedAlgorithm, 'merge');
-    d3.rebind(bollingerBands, bollingerAlgorithm, 'windowSize', 'value', 'multiplier');
+    rebind(bollingerBands, mergedAlgorithm, 'merge');
```
tlsim: Kept the 'merge' rebind - the visual tests use this to apply multiple indicators of the same type to the input data. Is it worth removing?

visual-tests/series/comparison.js
```diff
@@ -1,88 +0,0 @@
-(function(d3, fc) {
-    'use strict';
-
-    var data = [
-        fc.data.random.financial().startDate(new Date(2014, 1, 1))(50),
-        fc.data.random.financial().startDate(new Date(2013, 12, 15))(50)
-    ];
-
-    var percentageChange = fc.indicator.algorithm.calculator.percentageChange()
```
tlsim: This is soon to be removed from d3fc-technical-indicator. Could apply the calculation here, but seems simpler to remove the test completely (given that the scale and series components are both visually-tested elsewhere)?

src/indicator/algorithm/calculator/undefinedInputAdapter.js
```diff
@@ -1,54 +0,0 @@
-import d3 from 'd3';
-import slidingWindow from './slidingWindow';
-
-// Indicator algorithms are not designed to accomodate leading 'undefined' value.
-// This adapter adds that functionality by adding a corresponding number
-// of 'undefined' values to the output.
-export default function() {
```
tlsim: Previously used by envelope algorithm, with extra null check added in envelope renderer to compensate removal. Should this stay / move into d3fc-technical-indicator?

src/indicator/algorithm/merge.js
```diff
@@ -1,16 +1,15 @@
-import calculator from './calculator/slidingWindow';
-import d3 from 'd3';
-import {noop} from '../../util/fn';
+import { zip } from 'd3-array';
```
chrisprice: I think we're sticking with d3 in d3fc for now, any chance you could change this over to `d3.zip`?

package.json
```diff
@@ -9,11 +9,13 @@
     "babel-polyfill": "^6.7.2",
     "css-layout": "^1.1.0",
     "d3": "^3.5.4",
+    "d3-array": "^0.7.1",
```
chrisprice: I don't think this should be a dependency

src/indicator/algorithm/algorithm.js
```diff
@@ -1,8 +1,6 @@
 import bollingerBands from './bollingerBands';
-import calculator from './calculator/calculator';
 import exponentialMovingAverage from './exponentialMovingAverage';
 import macd from './macd';
-import merge from './merge';
```
chrisprice: Why have we removed merge?

src/indicator/algorithm/bollingerBands.js
chrisprice: Let's remove it when we split out the algorithms and renderers, that'll solve the visual test problem as well.

src/indicator/algorithm/bollingerBands.js
```diff
@@ -1,15 +1,15 @@
-import calculator from './calculator/bollingerBands';
-import d3 from 'd3';
+import { bollingerBands as _bollingerBands } from 'd3fc-technical-indicator';
```
chrisprice: Is it worth sticking with `calculator`? I wouldn't object to the naming if it was new, it just feels like we're changing it for the sake of changing it.

src/indicator/algorithm/merge.js
```diff
@@ -1,16 +1,15 @@
...
 export default function() {
 
     var merge = noop,
-        algorithm = calculator();
+        algorithm = identity;
 
     var mergeCompute = function(data) {
-        return d3.zip(data, algorithm(data))
+        return zip(data, algorithm(data))
             .forEach(function(tuple) {
```
chrisprice: FYI #956

visual-tests/series/comparison.js
chrisprice: Why are we removing it?

src/indicator/algorithm/calculator/undefinedInputAdapter.js
chrisprice: I'd prefer we keep this and move it into d3fc-technical-indicator. It's much easier to deal with the results if they're in a consistent *shape*, undefined guards on traversals quickly get unwieldy.

visual-tests/series/comparison.js
tlsim: https://github.com/d3fc/d3fc-technical-indicator/issues/8

chrisprice: In that case it sounds like we should just stick it in util for now then, I'm reluctant to remove it completely.

src/indicator/algorithm/calculator/undefinedInputAdapter.js
tlsim: Sounds good to me - I think there's more to discuss here given https://github.com/d3fc/d3fc-technical-indicator/issues/10 - but perhaps that's about configuration rather than consistency of the shape of the output?

src/indicator/renderer/envelope.js
```diff
@@ -13,23 +13,25 @@ export default function() {
         root = function(d) { return d.envelope; },
         decorate = noop;
 
+    var getUpperValue = function(d) {
+        var _root = root(d);
+        return _root && _root.upper ? _root.upper : undefined;
```
chrisprice: What about `_root.upper == 0` ?

src/indicator/renderer/envelope.js
```diff
@@ -13,23 +13,25 @@ export default function() {
...
+    };
+
+    var getLowerValue = function(d) {
+        var _root = root(d);
+        return _root && _root.lower ? _root.lower : undefined;
```
chrisprice: As above

tlsim:
@chrisprice Updated for d3fc-technical-indicator 4.0.1

PR #978, opened 2016-05-28
fix: tooltip text alignment in IE

### LINKED ISSUE #977: Tooltip text is not aligned correctly in IE
For example see the tooltip in [low barrel](https://d3fc.io/examples/low-barrel/index.html).

ie-tooltip-text-alignment

jleft:
Fixes #977

This is a similar to the axis tick text alignment problem and solution (800bd27b969b9d15b285c2f91b537db8d9083ab1).


chrisprice:
👍

PR #979, opened 2016-05-29
fix: improve skip weekend discontinuity provider DST support

### LINKED ISSUE #936: Skip weekend discontinuity provider does not account for DST
Something we're looking into fixing with BitFlux.

jleft:
Fixes #936


ColinEberhardt:
👍

PR #980, opened 2016-05-29
fix: add support for negative offsets in skip weekend discontinuity provider

### LINKED ISSUE #922: Skip weekend discontinuity provider offset does not work as expected with negative values
For example:

```js
var msPerDay = 24 * 60 * 60 * 1000;

var monday = new Date(2015, 0, 19); // Mon Jan 19 2015 00:00:00 GMT+0000

fc.scale.discontinuity.skipWeekends()
    .offset(monday, -msPerDay); // returns Sun Jan 18 2015 00:00:00 GMT+0000

// Would expect this to be returned (skipping the weekend):
// Fri Jan 16 2015 00:00:00 GMT+0000
```

I already have a fix and set of tests for this in BitFlux - I'll propose the changes here shortly.

jleft:
Fixes #922


ColinEberhardt:
👍

PR #983, opened 2016-06-08
IE 11, Polymer and Cartesian Chart

bmazzarol:
Hi Guys,

First off great library, we are using it with a lot of success.

We discovered an issue with using Shady DOM (Polymer shadow DOM in IE) and the svg-innerhtml library in IE 11.

Unfortunately any usage of innerHTML on an instance of a Shady DOM node will fail. This is tricky because it is not really possible (As far as I can tell) within the life cycle of a polymer element to get a handle on the root unwrapped SVG element (used in the Polymer template) without it resolving into a Shady DOM wrapped instance in IE.

Therefore the change replaces the html call (avoiding the innerHTML assignment) with manual d3 append calls. 

The changes are functionally equivalent and resolve the issue in this case.

However I can understand if the change is rejected, because it was required through no fault of d3fc and eventuates due to our choice of Polymer. I however thought it would be worth a shot.

Cheers
Ben

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -72,7 +72,6 @@ module.exports = function(grunt) {
                     'node_modules/d3/d3.js',
                     'node_modules/css-layout/dist/css-layout.js',
                     'node_modules/d3-svg-legend/d3-legend.js',
-                    'node_modules/svg-innerhtml/svg-innerhtml.js',
                     'node_modules/babel-polyfill/dist/polyfill.js',
                     'dist/d3fc.js'
                 ],
@@ -278,7 +277,7 @@ module.exports = function(grunt) {
                     'dist/d3fc.js': ['src/fc.js']
                 },
                 options: {
-                    external: ['css-layout', 'd3', 'd3-svg-legend', 'svg-innerhtml'],
+                    external: ['css-layout', 'd3', 'd3-svg-legend'],
                     format: 'umd',
                     moduleName: 'fc',
                     plugins: [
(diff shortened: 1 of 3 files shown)


chrisprice:
Thanks Ben, glad you're finding it useful 👍 

The initial reason behind using the `html` approach was to try and improve the legibility of the HTML snippet in the source. However, our original approach had been to `data-join` each of the elements as they were added and so the improvement in legibility that came with moving to `html` was significant. Your novel approach of using `enter().call` is also a significant step up in legibility and, had we known about it at the time, would definitely have been up for consideration.

I personally think we should make this change but what do you think @ColinEberhardt? 

If we choose to go with this approach then I think we should fully adopt it and remove the `svg-innerhtml` package. Also it would be good to squash it down to a single semantic commit.

ColinEberhardt:
I'm very happy with this change - thanks @bmazzarol 

Also, agree the dependency on `svg-innerhtml` should also be removed.

bmazzarol:
Thanks guys.

chrisprice:
@bmazzarol are you up for making the above changes? Not sure if you've closed this in anticipation of opening a new one? I'm happy to pick it up if not :smiley:

bmazzarol:
Sorry did not mean to close. I'm not all that clued in with github, I'm getting there.

chrisprice:
Sounds good to me. If you need any help with git[hub] stuff give us a shout.

bmazzarol:
Removed svg-innerhtml as a dependency.

I could not find any additional usages of the html function. I'm also am not aware of any other d3 functions which make use of innerHTML.

Let me know if there is anything else.

chrisprice:
Thanks again @bmazzarol 👍

Issue #987, opened 2016-06-16
Axis decorate doesn't work as expected


chrisprice:
See this [codepen](https://codepen.io/anon/pen/gMMXvy) which is a slightly modified version of the [example from the site](https://d3fc.io/components/svg/axis.html)

chrisprice:
Actually that might be a different issue due to overloading the `transform` attribute. However, this demonstrates exactly the issue I'm seeing -
```js
axis.decorate(g =>
  g.enter()
    .selectAll('path')
    .style({
      stroke: 'none',
    });
})
```

ColinEberhardt:
This is a combination of a couple of issues ...

 1. In the codepen the decorate function sets an attribute, `transform`, which is also set in the axis update selection. This explains the off behaviour where the axis rotation applied in the decorate / enter is overwritten by the second datajoin
 2. The issue with the path stroke not being set when new ticks 'enter', is due to cartesian exposing the axis series decorate, rather than the axis decorate. Hence the enter selection is the one that constructs the axis container, not the axis ticks. 

Fixed in #989

PR #989, opened 2016-06-16
Axis series now rebinds the underlying axis decorate

ColinEberhardt:
Fixes: #987


chrisprice:
That looks much better, I can't see a use-case for decorating the container.

Issue #991, opened 2016-06-17
How to draw custom disconnected lines


gargleblaster:
I am trying to draw a line for each closed trade, from the opening date and price to the closing date and price.  Ideally, I would like to be able to decorate each line based on direction (long/short), quantity, gain/loss, etc.

I am taking advantage of the zoom and pan capabilities in low barrel [where does that name come from??], so I need to draw these lines using d3fc's scale settings to have them move with the candles.  But I'm not sure how to add simple d3 lines (with points x1, y1, and x2, y2) to the d3fc framework.  If I use fc.series.line(), it won't let me draw distinct lines.

Each line would be based off of the following data:

`                                account: t.account,
                                direction: t.direction,
                                openDate: moment(t.openDate).tz('America/New_York').format('YY-MM-DD HH:mm:ss'),
                                closeDate: moment(t.closeDate).tz('America/New_York').format('YY-MM-DD HH:mm:ss'),
                                openQty: t.openQty,
                                closeQty: t.closeQty,
                                openPrice: t.openPrice,
                                closePrice: t.closePrice,
                                gain: t.gain
`

What is the right way to do this in d3fc?

BTW, thanks for a very cool framework!

chrisprice:
Maybe you could draw a quick sketch of what you're after to make it clear? But... it sounds like you want to draw n individual lines on the chart?

If so, I'd use `fc.series.multi` to wrap n `fc.series.line` instances. Whilst this modified low barrel (high stock...) example isn't exactly what you're after hopefully you can follow the gist of it -

https://codepen.io/anon/pen/VjKZgW?editors=0010

gargleblaster:
Thanks for your reply, Chris.  This isn't a great sketch, but you can see that I have four disconnected lines with arbitrary endpoints, along with some annotations.  I suspect from your example that `fc.series.line` forces you to connect your lines.  Also, it is quite possible that I would want multiple lines that overlap partially or completely, so I'm thinking that `fc.series.line` will not allow that.

Would it work to make each individual line its own series?  I'm hungry for documentation.  I find the wiki pages and web site whets my appetite but does not satisfy my need for details on how to use the APIs.

![image](https://cloud.githubusercontent.com/assets/728163/16164778/02a5ec28-34ae-11e6-901c-b8739e62035e.png)

chrisprice:
Ah that was just my laziness coming up with the example! It can render
arbitrary lines.

In my example the data rendered by a given line series is the item
[...]

PR #992, opened 2016-06-19
Add site examples

chrisprice:
Inspired by #991


site/src/examples/area-line-point/index.md
```diff
@@ -0,0 +1,27 @@
...
+</style>
+
+{{{area-line-point-html}}}
+
+<script>
+{{{area-line-point-js}}}
+</script>
+
+In this example we combine a {{{ hyperlink 'series/area.html' title='area series' }}}, {{{ hyperlink 'series/line.html' title='line series' }}} and {{{ hyperlink 'series/point.html' title='point series' }}} using a {{{ hyperlink 'multi.html' title='multi series' }}}. This combination allows for flexibility when styling the series to achieve e.g. gradient fills. It also demonstrates {{{ hyperlink 'decorate-pattern.html' title='decorating' }}} the point series to easily add labels to specific points on the series.
+
```
ColinEberhardt: I think one other very powerful connect it demonstrates which is worth mentioning is the multi-series `mapping` concept. In most examples we just use it to simple wire-up each series. However, in this case you are using it as a way to create a projection, which is more like a traditional `map`.

ColinEberhardt:
Nice examples 👍 

One minor thing, rather than using:

```
var yExtent = fc.util.extent()
     .fields([
         function(d) { return d.y; }
     ])
     .pad(0.2);
```

You could do the following:

```
var yExtent = fc.util.extent()
     .fields(['y'])
     .pad(0.2);
```

Although I am pretty sure you know that already, and are using the functional form because you don't like the more 'magic' version!

site/src/examples/area-line-point/index.md
chrisprice: Yea good point. I wanted to use annotated snippets of the js code but couldn't remember if that was possible?

chrisprice:
I'd really like to split out extent and eliminate a few of its "features" -
* Remove support for field names
* Remove support for nested arrays
* Split out date functionality e.g. import {linear, date} from 'd3fc-extent'. 
* Fix pad as being a tuple.
One for later in the week.

site/src/examples/area-line-point/index.md
ColinEberhardt: The `code block` helper outputs the contents of a file between `\\START` and `\\END` comments:

https://github.com/ScottLogic/d3fc/blob/master/site/build/handlebars-helpers/codeblock.js

It's a bit cumbersome, but does the trick.

ColinEberhardt:
Updated looks good - 👍 to merging

PR #1001, opened 2016-06-27
fixed bar series decorate origin

### LINKED ISSUE #1000: Bar series origin is not in the correct location
From the decorate docs:

> The d3fc series are designed so that the containing g element is translated to the location of each datum, as a result any elements that you add need only have a relative transform applied.

Bar series do not obey this. Their containing `g` element is translated to the baseline location

ColinEberhardt:
fixes #1000 :-D

Decorate visual test before:

screen shot 2016-06-27 at 10 11 49

(dots at the base of the bars)

After:

screen shot 2016-06-27 at 10 12 15

### DIFF
--- a/src/series/bar.js
+++ b/src/series/bar.js
@@ -23,7 +23,9 @@ export default function() {
         .selector('g.bar')
         .element('g');
 
-    function containerTranslation(d, i) {
+    // in order that new bars 'grow' from the baseline when transitioning, the origin
+    // is initially set to the base, transitioning to the correct location
+    function initialContainerTranslation(d, i) {
         if (orient === 'vertical') {
             return 'translate(' + base.x1(d, i) + ', ' + base.y0(d, i) + ')';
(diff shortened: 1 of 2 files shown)


src/series/bar.js
```diff
@@ -23,19 +23,31 @@ export default function() {
...
+    // is initially set to the base, transitioning to the correct location
+    function initialContainerTranslation(d, i) {
         if (orient === 'vertical') {
             return 'translate(' + base.x1(d, i) + ', ' + base.y0(d, i) + ')';
         } else {
             return 'translate(' + base.x0(d, i) + ', ' + base.y1(d, i) + ')';
         }
     }
 
+    // the container translation sets the origin to the 'tip'
```
chrisprice: Can be simplified and inlined.

ColinEberhardt:
@chrisprice good spot - updated

PR #1100, opened 2018-03-02
Monorepo

chrisprice:
Such trend :+1: 

Whilst this looks like a whooper, the last 2 commits are the significant ones.

The rest of the commits were imported using `lerna import`

Problems -

* [ ] Looks like d3fc-label-layout could cause some issues as it has its own site
* [x] Travis seems to have found a problem with the yarn workspaces config -
> yarn install v1.3.2
> error Workspaces can only be enabled in private projects
> info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
* [ ] `d3fc-technical-indicator` is causing a cyclic dependency


ColinEberhardt:
We should make this one a priority ... we have just about finished a new release of d3fc, and semantic-release is breaking for a number of packages. Firstly due to the revoked npm tokens, and secondly, lots of our GH tokens are not working either - possibly manually revoked by whoever originally set up semantic-release for these packages: https://github.com/d3fc/d3fc/pull/1103

PR #1180, opened 2018-08-23
canvas shapes now fill then stroke

chrisprice:
Not sure why we were doing it the other way around in the first place? I guess in most cases the shapes don't render a combination of fill and stroke?


ColinEberhardt:
Makes sense - does this fix https://github.com/d3fc/d3fc/issues/1161?

ColinEberhardt (approved):

PR #1184, opened 2018-08-24
fix: propagate falsely axis tick values

### LINKED ISSUE #1148: Falsy values not propagated to axis
_From @robatwilliams on January 9, 2018 12:38_

I want to pass `xTickSizeOuter(0)` in order to hide the end ticks on the axis (so I don't have an unlabeled tick at the start of my axis).

This doesn't work because d3fc uses a truthy check to decide if it should pass values through:
```
if (xTickSizeOuter) {
  xAxis.tickSizeOuter(xTickSizeOuter);
}
```

Workaround: pass zero wrapped in an array: `[0]`

_Copied from original issue: d3fc/d3fc-chart#35_

chrisprice:
Fixes #1148


chrisprice:
Accidentally pushed this to upstream...

PR #1185, opened 2018-08-28
fix: incorrect gridline position on update

### LINKED ISSUE #1087: odd gridline behavior when redrawing chart (non-idempotent?)
I'm trying to create a candlestick chart that can be scrolled by keyboard, and I redraw the chart by repeating the original drawing code with different array of data.

On some dates when there are large jumps in range I get strange non-even spacing of grid lines:

screen shot 2017-10-12 at 4 32 58 pm

screen shot 2017-10-12 at 4 28 53 pm

I've posted a replication case on codepen: [https://codepen.io/eponymous/pen/mBjozV](https://codepen.io/eponymous/pen/mBjozV) - click on the lower pane where the graph is, then hit the 'left' key twice. Hitting 'right' and 'left' to scroll the lowest (green) candlestick on and off the screen will reliably replicate the issue.

 

It seems to happen only when drawing using a div that is not empty due to previous call. The problem goes away if I delete all the children of the div before redrawing the chart, but this slows things down and causes flicker.

chrisprice:
Fixes #1087

### DIFF
--- a/packages/d3fc-annotation/src/svg/gridline.js
+++ b/packages/d3fc-annotation/src/svg/gridline.js
@@ -41,18 +41,18 @@ export default () => {
             const xData = xTicks();
             const xLines = xJoin(container, xData);
 
+            xLines.enter()
+                .attr('x1', xScaleOld)
+                .attr('x2', xScaleOld)
+                .attr('y1', yScale.range()[0])
+                .attr('y2', yScale.range()[1]);
+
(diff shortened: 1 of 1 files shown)


chrisprice:
Build failing because of new `d3-array` package release. We either need to move over to fixed dependency versions or change the bundle spec to do something smarter.

PR #1193, opened 2018-09-06
feat: add chartCartesian - a CSS grid-based chart

chrisprice:
I've not marked this as a breaking change because it should be backwards compatible unless there's heavy decoration going on


packages/d3fc-chart/README.md
```diff
@@ -78,33 +78,167 @@ The chart is constructed using a pair of scales. The scale configuration propert
...
+
+  // when the chart is added to the DOM
+  selection.enter()
+    // additionally add a div element for the label
+    .append('div')
+    .attr('class', 'chart-label')
+    // move the element into the top-label cell
+    .style('grid-column', 3)
+    .style('grid-row', 1)
+    // add some styling to the label
```
ColinEberhardt, line 148: as you're not including a screenshot, perhaps remove the lines relating to styling for the sake of brevity?

packages/d3fc-chart/README.md
```diff
@@ -43,26 +43,26 @@ var yExtent = fc.extentLinear()
 // gridlines (from d3fc-annotation)
 var gridlines = fc.annotationSvgGridline();
 // series (from d3fc-series)
-var line = fc.seriesSvgLine();
-var area = fc.seriesSvgArea()
+var line = fc.seriesCanvasLine();
```
ColinEberhardt: is it worth explicitly calling out that this example is rendering SVG gridlines over a canvas series? it's quite an interesting feature of the component.

i.e. it has a canvas plot area and an avg plot area, you can use either, or both.

packages/d3fc-chart/src/cartesian.js
```diff
@@ -0,0 +1,201 @@
+import { select, event } from 'd3-selection';
+import { scaleIdentity } from 'd3-scale';
+import { seriesSvgMulti, seriesCanvasMulti } from '@d3fc/d3fc-series';
```
ColinEberhardt, line 3: the default plot area for the current cartesian charts is a line series - not sure whether that is any better / worse than a multi?

ColinEberhardt (approved):
A few really minor points, but generally looks really good - I like the way that the grid-based component is has a much simpler 'render' function.

packages/d3fc-chart/README.md
chrisprice: I figured these would end up being copy/pasted so went for it looking half decent. Happy to remove it if you're keen though?

packages/d3fc-chart/src/cartesian.js
chrisprice: I opted for multi so that you wouldn't have to explicitly remove the plot area you're not using

packages/d3fc-chart/README.md
chrisprice: Good point, I've added a note to the comments

ColinEberhardt:
Thanks for the updates, looks good to me 👍

PR #1208, opened 2018-10-31
Add readme file to packages/d3fc

rlax:
When using lerna monorepo you need to provide README.md for all relevant folders inside `/packages`.
For example: https://github.com/facebook/create-react-app/tree/master/packages/create-react-app
results
https://www.npmjs.com/package/create-react-app
I have copied README.md contents from the root README.md.
Resolves: https://github.com/d3fc/d3fc/issues/1197


chrisprice:
Thanks @rlax!

Issue #1202, opened 2018-10-07
Quandl data feed no longer maintained


chrisprice:
From https://www.quandl.com/databases/WIKIP/documentation/about -

> As of April 11, 2018 this data feed is no longer actively supported by the Quandl community. We will continue to host this data feed on Quandl, but we do not recommend using it for investment or analysis.

cprice-scottlogic:
Remove it from d3fc

chrisprice:
We could look to replace it with something like the IEX feed but they're in the processing of switching their API versions. One to revisit in the future.

![image](https://user-images.githubusercontent.com/433457/50896674-afe42c80-1401-11e9-9215-37fe413d6dc2.png)

PR #1219, opened 2019-01-15
fix:  remove Quandl feed

dbarkey-scottlogic:
BREAKING CHANGE this is removing all references to the Quandl feed as it is no longer actively supported

Closes [#1202](https://github.com/d3fc/d3fc/issues/1202)


chrisprice (changes requested):
It doesn't look like the commit contains the BREAKING CHANGE blah blah blah

chrisprice (approved):

PR #1223, opened 2019-01-31
Fixed display of bubble-sort transition example

andy-lee-eng:
The bars weren't visible. Now that d3fc outputs a canvas element as well as svg, the white background on .plot-area meant that the svg element was hidden.

Also a couple of other minor edits.


ColinEberhardt:
Thanks for fixing that - much appreciated!

PR #1226, opened 2019-02-14
Export cartesian/element css variables as part of package

andy-lee-eng:
Exports `cssChartCartesian` and `cssElement` variables as strings containing the necessary css styles that should be applied.

When showing the d3fc chart inside a shadow-dom element, this allows us to do:

````javascript
import css from "../styles.css";
import {cssChartCartesian, cssElement} from "d3fc";

const styleWithD3FC = `${css}${cssChartCartesian}${cssElement}`;

// Append `style` block with content `styleWithD3FC` to the shadow-dom root.

````

### DIFF
--- a/packages/d3fc-chart/index.js
+++ b/packages/d3fc-chart/index.js
@@ -1,4 +1,3 @@
 export { default as chartSvgCartesian } from './src/svg/cartesian';
 export { default as chartCanvasCartesian } from './src/canvas/cartesian';
 export { default as chartCartesian } from './src/cartesian';
-export { default as cssChartCartesian } from './src/css';
--- a/packages/d3fc-chart/index.js
+++ b/packages/d3fc-chart/index.js
@@ -1,3 +1,4 @@
 export { default as chartSvgCartesian } from './src/svg/cartesian';
 export { default as chartCanvasCartesian } from './src/canvas/cartesian';
 export { default as chartCartesian } from './src/cartesian';
+export { default as cssChartCartesian } from './src/css';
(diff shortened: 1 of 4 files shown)


chrisprice (changes requested):
A minor thing and a perhaps more controversial one:

* The commit message should follow the semantic commit message format. Have a quick gander through the commit log for examples, is expect this would be a non breaking feature (`feat`). 
* I wasn't expecting this to be (re-)exported at the top level as it's a bit of a hack. Any objection to exporting it at the sub-package level? We also get to skip the what to name it discussion if we do that!

andy-lee-eng:
I don't actually know what you're asking for. It sounds like you have a particular implementation in mind?

chrisprice:
Sorry for being cryptic. All I meant was prefixing the following lines with `export` -

https://github.com/d3fc/d3fc/blob/master/packages/d3fc-element/src/css.js#L2

https://github.com/d3fc/d3fc/blob/master/packages/d3fc-chart/src/css.js#L2

andy-lee-eng:
So we'd have to do:

````javascript
import css from "../styles.css";
import {css as cssChartCartesian} from "@d3fc/d3fc-chart/src/css";
import {css as cssElement} from "@d3fc/d3fc-element/src/css";

const styleWithD3FC = `${css}${cssChartCartesian}${cssElement}`;

// Append `style` block with content `styleWithD3FC` to the shadow-dom root.
````

I assumed this wouldn't work, because we have only npm-installed "d3fc" (which only has a /build folder, not /src), but it turns out that you also get all the "@d3fc/d3fc-x" packages installed even though you haven't asked for them (along with all the source folders). Is that right?

Still, I'm not sure I'm comfortable with the above code, having not installed "@d3fc/d3fc-chart" or "@d3fc/d3fc-element"...

ColinEberhardt:
TL;DR; it is working as I'd expect!

> I assumed this wouldn't work, because we have only npm-installed "d3fc" (which only has a /build folder, not /src), but it turns out that you also get all the "@d3fc/d3fc-x" packages installed even though you haven't asked for them (along with all the source folders). Is that right?

d3fc-x packages are dependencies of d3fc, so I'd expect them to also get installed via the usual npm resolution process.

You can consume d3fc in various different ways ...

For those that just want to consume it as a simple / single bundled dependency, this is available in the build folder of d3fc - https://unpkg.com/d3fc@14.0.26/build/

However, for those that want to consume it as separate modules (which would allow them to perform a more optimised build / bundle), potentially as part of a webpack build, we also distribute the source with the [module property pointing at the ES2015 'entry point'](https://github.com/d3fc/d3fc/blob/master/packages/d3fc-axis/package.json#L13) of each module.

I must admit, I've not kept up with the [myriad ways](https://github.com/stereobooster/package.json#package-bundlers) you could / should bundle libraries. We did discuss this around a year ago https://github.com/d3fc/d3fc/issues/1060

andy-lee-eng:
Fair enough. I assumed that if you wanted to consume the separate modules you'd also have to install the modules separately, since the otherwise the minimised file in the d3fc/build folder has everything you need.

chrisprice:
My understanding (again potentially out of date) is that by distributing the source, a tree-shaking optimisation pass would have an easier job of throwing away the bits of the library you're not using (without you having to explicitly import the bits you are).

> Still, I'm not sure I'm comfortable with the above code, having not installed "@d3fc/d3fc-chart" or "@d3fc/d3fc-element"...

Agreed but at this point I consider this a relatively niche use-case so I'm not too worried about it being ugly, just so long as it's technically possible. A more involved but nicer solution would be to detect the document (from `element.ownerDocument`) in which the `d3fc-chart` or `d3fc-element` was being inserted and ensure that it contained the relevant stylesheet.

andy-lee-eng:
Would you like me to explore the `element.ownerDocument` idea? I'm happy to, but I guess there's a risk of breaking something else...

If we insert the style blocks with a reasonably obfuscated `id`, it should then be simple to check that it's there when rendering a chart/element.

andy-lee-eng:
This PR is to resolve https://github.com/d3fc/d3fc/issues/1224

chrisprice (approved):

PR #1227, opened 2019-02-15
Fixed issue with horizontal seriesGrouped offset in wrong direction

andy-lee-eng:
When the seriesSvgGrouped is used with a seriesSvgBar in "horizontal" orientation (i.e. bars instead of columns), it still offsets the individual bars in the X direction, so they overlap each other, instead of in the Y direction.

### DIFF
--- a/packages/d3fc-series/examples/grouped.html
+++ b/packages/d3fc-series/examples/grouped.html
@@ -19,7 +19,9 @@
     <svg id="grouped-svg-autobandwidth-bandscale"></svg>
     <svg id="grouped-svg-autobandwidth-linearscale"></svg>
     <svg id="grouped-svg-variable-bandwidth"></svg>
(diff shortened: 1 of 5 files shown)


andy-lee-eng:
Before
image

After
image

ColinEberhardt:
Thanks for fixing this issue @DevAndyLee - much appreciated!

Can you please add an example of horizontal orientation to the examples page:

https://github.com/d3fc/d3fc/blob/master/packages/d3fc-series/examples/grouped.js

We don't have much test coverage, instead we tend to rely on visual inspection using these 'example' pages. Would be good to get this added too.

andy-lee-eng:
Added examples of SVG and Canvas horizontal grouped series

ColinEberhardt:
great job - thanks @DevAndyLee

PR #1228, opened 2019-02-18
Added multi-level grouping ability to chart axis labelling

andy-lee-eng:
After chatting with Colin on Friday, about whether it was feasible to replace the Cartesian chart's axis with a custom or extended version of the axis component, he said he'd be interested in whether we could improve the capabilities of the existing axis component...

Extended cartesian chart and axis so that:
* x/y axis size can be specified (allows for extra room for rotated axis or for multi-levels as below)
* axis has a "tickGrouping" function that allows you to group the axis labelling into multiple levels

image

### DIFF
--- a/packages/d3fc-axis/src/axis.js
+++ b/packages/d3fc-axis/src/axis.js
@@ -13,6 +13,7 @@ const axis = (orient, scale) => {
     let tickSizeOuter = 6;
     let tickSizeInner = 6;
     let tickPadding = 3;
+    let tickGrouping = null;
 
     const svgDomainLine = line();
 
@@ -20,6 +21,8 @@ const axis = (orient, scale) => {
         .key(identity);
 
     const domainPathDataJoin = _dataJoin('path', 'domain');
+
+    const groupDataJoin = _dataJoin('g', 'group');
 
     // returns a function that creates a translation based on
     // the bound data
@@ -31,7 +34,11 @@ const axis = (orient, scale) => {
                 offset = Math.round(offset);
             }
         }
-        return d => trans(scale(d) + offset, 0);
+        return d => trans((d.domain ? centerOf(scale, d.domain) : scale(d)) + offset, 0);
+    };
+
+    const centerOf = (scale, values) => {
+        return values.reduce((s, v) => s + scale(v), 0) / values.length;
     };
 
     const translate = (x, y) =>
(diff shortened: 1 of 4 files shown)


chrisprice (changes requested):
As the grouping changes are a bit more involved it'll take me a bit longer to review those. However, the size change sounds like a good one on its own. Maybe you could split that into its own PR?

Comments specifically on that change:

* Can we confirm the naming matches similar properties? I've no reason to think it doesn't but we've been caught out by this before
* As this value is ultimately applied as a CSS size, should we also allow the units to be specified?
* I'm not a fan of 0 being used as a magic value because I can think of legitimate uses for a 0 height, could we use`null` instead?

ColinEberhardt:
Thanks @DevAndyLee this is a great contribution. I've had a quick chat with @chrisprice about this - here is some general feedback.

This PR contains 3 distinct changes, would it be possible to separate them out? In order of complexity:

1. The change that allows you to set the width / height of cartesian chart axes. I think Chris has some more minor feedback, but this should be quite easy to complete (as an aside, can we use width / height instead of size?)
2. When grouping, the axis centre aligns ticks - this is a really useful feature which [we have discussed before](https://github.com/d3fc/d3fc/issues/541). I'd really like to see this as a fully supported option of the axis, e.g. a boolean property. 
3. The grouping functionality itself works really well, I experimented with it a little and like the results. Regarding the implementation, rather than modifying the axis, would it be possible to implement this via a component that uses axes internally? You might have to make a few changes to axis, e.g. expose the ability to 'turn off' the domain line. If this implementation is possible it will keep the axis component nice and simple. 

For (3) I also think we need to look at some more test cases, how does it work with continuous scales? For example, can we render a time scale with two levels of labels (e.g. quarters and years) for a time series chart.

With (3) can we use this to render an axis with different tick lengths? e.g. a bit like a ruler where the 10s are longer than the unit ticks. This might not be straightforward, but worth considering what the overall scope is. (in this case, it might be easier to just decorate a regular axis).

andy-lee-eng:
I'll create PRs for (1) and (2).

(3) looks a bit more challenging, but in the meantime I've figured out how to implement a local extended version of `chartSvgCartesian`, overriding the `draw` event for the axis to provide a customised component to do the actual drawing.

Would you like to close this PR or keep it here for now for reference?

ColinEberhardt:
Thanks @DevAndyLee, PRs for (1) and (2) would be great. Happy to leave this PR open for future reference.

andy-lee-eng:
For (2), any suggestions for the property name? And does "center align ticks" refer to putting the text "in the center between the ticks" or to putting the tick "in the center of the text"?!

e.g.
`tickCenterAlign`: true/false
`tickPosition`: "middle"/"between"
`tickAlign`: "center"/"surround"

ColinEberhardt:
For (2) this relates to the case where tick labels relate to a continuum of values rather than a discrete instance. 

A good example is 'Monday', which relates to all the time within that given day. Currently that would be labelled as follows:

```
    |          |         |
  Monday    Tuesday  Wednesday
```

But the following would be better:

```
    |          |         |
    |  Monday  | Tuesday | Wednesday
```

I guess the issue here is that it isn't 'alignment' as such.

Anyhow, names are easy to change. Let's start with `centerAlignTicks`, but have a bit more of a think before merging.

PR #1231, opened 2019-02-25
feat: Option to center axis tick labels between tick lines

andy-lee-eng:
- `centerAlignTicks` boolean property (defaults to false)
- `tickOffset` function gives a default implementation based on the scale's `step`, but can be overridden to customise behaviour - e.g. for non-evenly-spaced custom scale.
- `labelOffset` should not include `tickSizeInner` when centered between tick lines

image


packages/d3fc-axis/src/axis.js
```diff
@@ -112,10 +114,14 @@ const axis = (orient, scale) => {
 
             // update
             g.select('path')
+                .attr('visibility', (d, i) => (i === ticksArray.length - 1 && centerAlignTicks) ? 'hidden' : '')
                 .attr('d',
-                    (d) => svgDomainLine(pathTranspose([
-                        [0, 0], [0, sign * tickSizeInner]
-                    ]))
+                    (d) => {
+                        const offset = sign * tickOffset(d);
```
chrisprice, line 120: Should we pass `i` in as well?

ColinEberhardt:
Thanks for the PR @DevAndyLee looks good. One general question, I'm wondering about the need for `tickOffset`? As illustrated by your example you are using this for situations where the scale step is non-uniform. 

In order to render the ticks / labels it [obtains the array of ticks](https://github.com/d3fc/d3fc/pull/1231/files#diff-0763d5d6d06cd6337117c5c19d980950R79) - can you just use this to determine the location of each label instead?

packages/d3fc-chart/README.md
```diff
@@ -271,4 +271,4 @@ The Cartesian chart exposes the scale properties with either an `x` or `y` prefi
 <a name="cartesian_yDecorate" href="#cartesian_yDecorate">#</a> *cartesian*.**yDecorate**(...)  
 ...
 
-The Cartesian chart exposes the [d3fc-axis](https://github.com/d3fc/d3fc/tree/master/packages/d3fc-axis#d3fc-axis) *ticks*, *tickSize*, *tickValue*, *tickFormat* and *decorate* properties with either an `x` or `y` prefix.
+The Cartesian chart exposes the [d3fc-axis](https://github.com/d3fc/d3fc/tree/master/packages/d3fc-axis#d3fc-axis) *ticks*, *tickSize*, *tickValue*, *tickFormat*, *tickArguments*, *tickSizeInner*, *tickSizeOuter*, *tickPadding*, *centerAlignTicks* and *decorate* properties with either an `x` or `y` prefix.
```
ColinEberhardt, line 274: Thanks for expanding on this list - well spotted!

andy-lee-eng:
> Thanks for the PR @DevAndyLee looks good. One general question, I'm wondering about the need for `tickOffset`? As illustrated by your example you are using this for situations where the scale step is non-uniform.
> 
> In order to render the ticks / labels it [obtains the array of ticks](https://github.com/d3fc/d3fc/pull/1231/files#diff-0763d5d6d06cd6337117c5c19d980950R79) - can you just use this to determine the location of each label instead?

Unfortunately not, or at least I couldn't find a way. The `scale` gives the centre of each tick but not its size, which we need so we can draw the line on the edge. I considered passing `d` to the `step()` function (it would be ignored by `scaleBand` but could be used by a custom scale to give the correct size for each tick). However, it didn't feel right.

packages/d3fc-axis/src/axis.js
```diff
@@ -13,6 +13,8 @@ const axis = (orient, scale) => {
     let tickSizeOuter = 6;
     let tickSizeInner = 6;
     let tickPadding = 3;
+    let centerAlignTicks = false;
```
chrisprice, line 16: I'm not sure about the naming. The convention seems to be a `tick` prefix - maybe we could we go for `tickCenter`/`tickAlign` (with `'center'` as a value)?

packages/d3fc-axis/src/axis.js
chrisprice: Actually picking up on Colin's question, could we access `ticksArray[i+1]`?

chrisprice (changes requested):

ColinEberhardt:
> The `scale` gives the centre of each tick but not its size

This is what @chrisprice is referring to when he mentions `ticksArray[i+1]`, we can use this to compute the distance between each tick and position the label.

I've found the D3 [pairs](https://github.com/d3/d3-array#pairs) function quite useful for this sort of thing before.

andy-lee-eng:
image

chrisprice:
image

(I like this new way of communicating 😄)

ColinEberhardt:
Oh, I see your point @DevAndyLee - for continuous scales the calculation based on `tickArray` should be fine, but for ordinal / band scales we actually want to shift the location of the tick.

Solving both is going to require a bit more thought!

chrisprice:
If my suggestion want too cryptic, does it not solve both?

andy-lee-eng:
The point I was trying to make is that `ticksArray[]` tells you where to put the center of the text, but it doesn't tell you where to put the tick line.

If the distance between the ticks is always the same, then you can just position it half-way between. If the distance changes, then you need another way to work out where to put it.

There are actually lots of ways we could do this...

1. Use a function that can be overridden (as in this PR).

2. Assuming we're using a custom `scale` anyway, add some function to the `scale` that can do this (the `step` property is already specific to point and band scales). e.g. We could use `step(d)` to let the scale vary the step size, or else create a new function.

3. Use a data structure that associates the grouped labels with the underlying domain (as per my original PR - e.g. "Fruit" becomes `{ name: "Fruit", domain: ["Apples", "Bananas"] }`). This is actually a bit more complicated here, because in trying to keep d3fc-axis simple, it no longer has access to the original domain and scale.

ColinEberhardt:
> If my suggestion want too cryptic, does it not solve both?

I must admit, I didn't quite get it!

Perhaps we should get together and have a chat about it. Definitely more complicated than I had originally anticipated.

PR #1232, opened 2019-02-28
feat: Dynamically insert css into head or shadow-dom if missing

andy-lee-eng:
`d3fc-element` and `d3fc-chart` dynamically load their required css into either the document head, or else the root of the shadow dom, rather than always loading the css info the document head as a one-off.

The style element is created with an id, and we check if it exists before adding it again.

Added an example of a chart hosted in a shadow-dom.

### DIFF
--- a/packages/d3fc-element/src/css.js
+++ b/packages/d3fc-element/src/css.js
@@ -1,16 +1,22 @@
 // Adapted from https://github.com/substack/insert-css
-const css = `d3fc-canvas,d3fc-svg{position:relative;display:block}\
+export const css = `d3fc-canvas,d3fc-svg{position:relative;display:block}\
 d3fc-canvas>canvas,d3fc-svg>svg{position:absolute;height:100%;width:100%}\
 d3fc-svg>svg{overflow:visible}`;
 
-const styleElement = document.createElement('style');
-styleElement.setAttribute('type', 'text/css');
+export const insertCss = (node, css, id) => {
+    const root = node.getRootNode();
+    const head = root.querySelector('head') || root;
 
-document.querySelector('head')
-  .appendChild(styleElement);
+    if (!head.querySelector(`#${id}`)) {
+        var styleElement = document.createElement('style');
+        styleElement.setAttribute('type', 'text/css');
+        styleElement.setAttribute('id', id);
+        if (styleElement.styleSheet) {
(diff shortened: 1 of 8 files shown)


andy-lee-eng:
Presented as a possible alternative to https://github.com/d3fc/d3fc/pull/1226

This avoids having to workaround the css problem when using shadow-dom, because it always gets inserted in the correct place.

I made `applyCss` a method on the custom HtmlElement simply because importing `insertCss` from `d3fc-element` caused issues. I guess it could be pulled out into a new package instead?

packages/d3fc-element/src/css.js
```diff
@@ -1,16 +1,22 @@
 // Adapted from https://github.com/substack/insert-css
-const css = `d3fc-canvas,d3fc-svg{position:relative;display:block}\
+export const css = `d3fc-canvas,d3fc-svg{position:relative;display:block}\
 d3fc-canvas>canvas,d3fc-svg>svg{position:absolute;height:100%;width:100%}\
 d3fc-svg>svg{overflow:visible}`;
 
-const styleElement = document.createElement('style');
-styleElement.setAttribute('type', 'text/css');
+export const insertCss = (node, css, id) => {
+    const root = node.getRootNode();
```
chrisprice, line 7: I think we might have compatibility problems using this -https://developer.mozilla.org/en-US/docs/Web/API/Node/getRootNode

chrisprice (commented):
Thanks for taking a shot at this. Like yourself, I'm not keen on the additional method on D3FC element. I also have some concerns about browser compatibility.

At this stage, I'm not sure it's worth investing the time in fixing the above problems. I reckon anyone using shadow DOM is likely to be using a build chain that would allow them to ingest a third-party CSS file. 

That seems like the "proper" fix to me but again I'm not sure it's worth changing our build toolchain to support that. We've only had the one report of this being an issue and #1226 offers a reasonable mitigation for that.

packages/d3fc-element/src/css.js
andy-lee-eng: I'll leave it up to your judgement.

We still can't ingest the CSS at build time, only insert it at runtime. I suspect we won't use #1226 for Perspective because importing from "@d3fc/d3fc-chart" instead of from "d3fc" causes errors in our webpack build output.

Issue #1233, opened 2019-03-06
seriesCanvasPoint implementation of size function different to svg


Ro4052:
There is a difference in the implementation of the `size` function between `seriesCanvasPoint` and `seriesSvgPoint`. When calling `size` in the SVG version, the function is provided individual data points, i.e. `symbol(d, i)`. In the canvas case, the function is called as `symbol(data)`, meaning the whole data array is passed. In our case, we want the point size based off individual data points and so the SVG implementation works, but not the canvas one.

ColinEberhardt:
Thanks @Ro4052 that sounds very much like a bug! Would you be up for creating a fix?

Ro4052:
That's fine, we have a fix, I'm just looking at making an example for it.

PR #1234, opened 2019-03-06
Canvas symbol functionality to match SVG

Ro4052:
#1233 Fixed canvas functionality so that it matches that of the SVG implementation and added a new example.


ColinEberhardt:
Thank you 👍

PR #1235, opened 2019-03-07
fix: Canvas multi series checks if series decorate exists before invoking

Ro4052:
Unlike the SVG implementation, `seriesCanvasMulti` invokes `decorate` on individual series. This is breaking in the cases where series does not have a `decorate` function. `seriesCanvasMulti` now checks if the series has a `decorate` before trying to call it.


packages/d3fc-series/src/canvas/multi.js
```diff
@@ -18,15 +18,19 @@ export default () => {
...
-                adaptedDecorate(c, d, i);
-            });
-
-            dataSeries(seriesData);
-
-            dataSeries.decorate(adaptedDecorate);
+            if (dataSeries.decorate) {
+                const adaptedDecorate = dataSeries.decorate();
+                dataSeries.decorate((c, d, i) => {
+                    base.decorate()(c, data, index);
```
chrisprice, line 25: I'm thinking we should still call `base.decorate` in the case that `dataSeries.decorate` is not defined. What do you think?

packages/d3fc-series/src/canvas/multi.js
```diff
@@ -18,15 +18,19 @@ export default () => {
...
+                    adaptedDecorate(c, d, i);
+                });
+
+                dataSeries(seriesData);
+
+                dataSeries.decorate(adaptedDecorate);
+            } else {
+                dataSeries(seriesData);
```
chrisprice: Minor stylistic nit pic - I'm not a fan of duplicating this line.

packages/d3fc-series/src/canvas/multi.js
Ro4052: I don't see any issues with this

packages/d3fc-series/src/canvas/multi.js
Ro4052: I agree with this point but I wasn't sure how to refactor. I'll take another look.

Ro4052:
@chrisprice, I've made a couple of changes based on your comments. Let me know if there's anything else you think needs doing.

packages/d3fc-series/src/canvas/multi.js
```diff
@@ -18,15 +18,21 @@ export default () => {
...
+            if (dataSeries.decorate) {
+                adaptedDecorate = dataSeries.decorate();
+                dataSeries.decorate((c, d, i) => {
+                    base.decorate()(c, data, index);
+                    adaptedDecorate(c, d, i);
+                });
+            }
 
             dataSeries(seriesData);
+            base.decorate()(context, data, index);
```
chrisprice: Sorry it's taken a little while to look at this. It all looks good to me apart from this line, I'm not sure why it's there?

Ro4052: That was in reference to your comment about calling `base.decorate` even if `dataSeries.decorate` is undefined. Perhaps I misunderstood? I can remove the line if you'd prefer.

chrisprice:
Would this implementation mean we'd double call it when dataSeries.decorate
is defined?

Ro4052:
Ah, OK, I see the problem. I'll get that sorted.

chrisprice (approved):
LGTM 👍

PR #1236, opened 2019-03-07
feat: Allow custom axis components in cartesian chart

andy-lee-eng:
This gives us the ability to use a custom implementation of an axis render, or enhance the existing one.
Example included


ColinEberhardt:
Thanks @DevAndyLee - had a chat with @chrisprice and have a few suggestions.

1. I think we should allow different x and y axis components - this ties in with our discussion about creating two different axis types, one for continuous, one for sequential. 
2. Rather than adding a third arg to the constructor function, can we instead provide an 'options' object? It's much more readable. Here's an example:

```
import { axisBottom, axisRight, axisLeft, axisTop } from '@d3fc/d3fc-axis';

const axisFactory = {
  left: axisLeft,
  right: axisRight,
  top: axisTop,
  bottom: axisBottom
};

const chart = fc.chartSvgCartesian({
    xScale: d3.scaleLinear(),
    yScale: d3.scaleLinear(),
    xAxis: axisFactory,
    yAxis: axisFactory
})
```

It would be good to make this a non breaking change, i.e. if they supply 2 args, use the current behaviour.

Does that make sense?

packages/d3fc-chart/examples/custom-axis.html
```diff
@@ -0,0 +1,19 @@
+<html>
```
ColinEberhardt, line 1: Thanks for adding an additional example 👍

ColinEberhardt (changes requested):
as per comments above.

ColinEberhardt:
> It would be good to make this a non breaking change, i.e. if they supply 2 args, use the current behaviour.

Actually, just thought - checking number of args doesn't work as we support defaults. Better to check if the first arg is a scale or a config object.

D3 components do have a name property, `d3.scaleLinear().name` - not seen this documented anywhere though!

andy-lee-eng:
Ok, I'll try the `{settings}` approach.

What about if I made the axisFactory parameter more factory-function like?

i.e. to create an instance you'd do this

````javascript
   const xAxis = axisFactory('left')(xScale);
````

instead of 

````javascript
   const xAxis = axisFactory.axisLeft(xScale);
````

Then maybe it would be unnecessary to provide two parameters for xAxis and yAxis?

**Comment by andy-lee-eng** [comment]
Also, to get different components for X and Y you can do this (without requiring 2 params):

````javascript
import { axisBottom,  axisTop } from '@d3fc/d3fc-axis';
import { customAxisRight, customAxisLeft } from './myCustomAxis';

const axisFactory = {
  left: customAxisLeft,
  right: customAxisRight,
  top: axisTop,
  bottom: axisBottom
};

const chart = fc.chartSvgCartesian({
    xScale: d3.scaleLinear(),
    yScale: d3.scaleLinear(),
    axis: axisFactory
});
````

**Comment by andy-lee-eng** [comment]
I've ammended it so usage is like this:

```javascript
chart = chartSvgCartesian({
  xScale: scaleLinear(),
  yScale: scaleLinear(),
  xAxis: {
    bottom: myCustomAxis
  },
  yAxis: {
    left: d3.axis
  }
})
```

packages/d3fc-chart/README.md
```diff
@@ -157,6 +157,29 @@ chart.decorate(selection => {
 });
 ```
 
+#### Using custom axis components
+
+By default the chart uses d3fc-axis for the axis comonent. This example shows how to specify which axis components to use for the X and Y axis.
```
ColinEberhardt: com *p* onent

packages/d3fc-chart/README.md
```diff
@@ -157,6 +157,29 @@ chart.decorate(selection => {
...
+
+```javascript
+var chart = fc.chartCartesian({
```
ColinEberhardt, line 165: const

ColinEberhardt (changes requested):
A couple of very minor documentation changes. Really like the implementation, very clean and understandable 👍

packages/d3fc-chart/README.md
andy-lee-eng: I used `var` because that's what the rest of the ReadMe uses. Would you like me to change them all to `const`?

ColinEberhardt: Oh, hadn't realised that. Just stick with `var` it's not a bit deal.

ColinEberhardt (approved):

ColinEberhardt:
Thanks @DevAndyLee 👍

PR #1238, opened 2019-03-08
feat: Two types of axis components, with specialisations

andy-lee-eng:
The default `axis` is backwards compatible (works for all scales), and has an additional `tickCenterLabel` property for positioning the label between ticks. This can be useful for continuous scales, but is less useful for ordinal scales (example included)

The alternative `axisOrdinal` will also work for all scales, and includes an optional `tickLineAlign` to control positioning of the tick line. It defaults to "center", putting the tick line in the middle of the label, but can be "left" or "right". This is more useful for ordinal scales (example included)


andy-lee-eng:
This is an alternative to https://github.com/d3fc/d3fc/pull/1231

ColinEberhardt:
Thanks @DevAndyLee this looks really good. I'll give this ago over the next couple of days.

packages/d3fc-chart/src/cartesian.js
```diff
@@ -20,9 +20,9 @@ export default (xScale = scaleIdentity(), yScale = scaleIdentity()) => {
     let xOrient = functor('bottom');
     let canvasPlotArea = seriesCanvasMulti();
     let svgPlotArea = seriesSvgMulti();
-    let xAxisStore = store('tickFormat', 'ticks', 'tickArguments', 'tickSize', 'tickSizeInner', 'tickSizeOuter', 'tickValues', 'tickPadding');
+    let xAxisStore = store('tickFormat', 'ticks', 'tickArguments', 'tickSize', 'tickSizeInner', 'tickSizeOuter', 'tickValues', 'tickPadding', 'tickCenterLabel', 'tickLineAlign');
```
andy-lee-eng: I'm not sure about this, because `tickCenterLabel` and `tickLineAlign` are not necessarily available on the chosen axis. It works, because `store` only tries to set the property on the axis if it has been given a value for it (i.e. it will throw an error if you try to set `tickLineAlign` on a non-ordinal axis).

If we don't do this, it makes things a bit harder for the developer - i.e. you'd need to do something like:

```javascript
  fc.chartSvgCartesian({
    xScale: d3.scaleBand(),
    yScale: d3.scaleLinear(),
    xAxis: {
      bottom: scale => fc.axisOrdinalBottom(scale).tickLineAlign('right')
    }
  })
  .xTickSize(10)
  //...
```

instead of 

```javascript
  fc.chartSvgCartesian({
    xScale: d3.scaleBand(),
    yScale: d3.scaleLinear(),
    xAxis: {
      bottom: fc.axisOrdinalBottom
    }
  })
  .xTickLineAlign('right')
  .xTickSize(10)
  //...
```

ColinEberhardt:
Thanks @DevAndyLee - I've just had a go with this, playing around with various options, testing with discontinuous scale, etc ... and it works perfectly 👍 

The only thing that threw me was the ordinalAxis API, specifically the `tickLineAlign` property. I had assumed that as it takes `left` and `center` values it might also accept `right`. I'm not entirely sure what I was expecting it to do with that, but it does feel a little odd.

Considering that `center` alignment renders in exactly the same way as the d3fc and d3 axis, do we need it? Perhaps the ordinalAxis doesn't need this property and should always render as per `tickLineAlign=left`?

andy-lee-eng:
Hi @ColinEberhardt . It does actually accept `right`, but it isn't mentioned in the code because it does `if ('left') {...} else {...}` - if it's 'left', it reverses the direction and hides the first one, and otherwise defaults to hiding the last one (for 'right').

In fact, I used `right` in my example and in the ReadMe.

However, I agree with your other point. The `tickLineAlign` values `left` and `right` are visually identical anyway, and `center` is the same as the regular axis. The only thing I had in mind was that we might add more ordinal-like functionality in the future, so you may want `tickLineAlign='center'` combined with some other feature of `axisOrdinal`.

For now though, I'd be happy to remove the property and default it to what we want.

andy-lee-eng:
I've removed `tickLineAlign` now

ColinEberhardt:
Hi @DevAndyLee looks great - I've rebased and made some doc changes, see the new PR

ColinEberhardt:
Great job - this has landed in https://github.com/d3fc/d3fc/pull/1245

PR #1239, opened 2019-03-11
feat: [For Preliminary Review - Do Not Merge] added new padding strategy component to extent package

### LINKED ISSUE #1229: Padding is crossing 0 when datapoints are exclusively negative or exclusively positive.
image

image

image

These three images illustrate the problem. I'd like to make a pull request that adds a new function to linearExtent - something to the effect of "padAcrossZero(i take a boolean)" - which allows a user to prevent padding from taking an axis' extent across zero in such instances.

I guess it would be a breaking change to make this the default behaviour, but I think it might be a good idea to try to move in that direction if we can.

redbearsam:
- examples
- tests (incomplete set for demo purposes)

_Not ready for merge._ A new padding strategy component has been added in this commit. 

For the time being, the tests (and potentially examples) are an incomplete set, as I didn't want to sink too much into them before running the approach through review.

The alternative approach that may be preferred would be effectively overloading the pad() function rather than adding a new paddingStrategy() function. The pad() function then detects whether it's passed a function, in which case that function overrides the defaultPadding(). In the case that it's passed only a number, it behaves in the pre-existing manner.

Alternatively, you might have an entirely different approach in mind, or have cooled on the feature. I hope the examples illustrate the use case for this feature enough to justify it's inclusion. 

This commit resolves #1229

### DIFF
--- a/packages/d3fc-extent/examples/.eslintrc.json
+++ b/packages/d3fc-extent/examples/.eslintrc.json
@@ -1,0 +1,10 @@
+{
+  "extends": "defaults/configurations/eslint",
+  "env": {
+    "browser": true
+  },
+  "globals": {
+    "fc": true,
+    "d3": true
+  }
+}
--- a/packages/d3fc-extent/examples/hardLimitZeroPadding.html
+++ b/packages/d3fc-extent/examples/hardLimitZeroPadding.html
@@ -1,0 +1,38 @@
+<!doctype html>
+<html>
+<head>
+    <script src="../node_modules/babel-polyfill/dist/polyfill.js"></script>
+    <script src="../node_modules/d3/build/d3.js"></script>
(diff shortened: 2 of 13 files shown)


chrisprice:
I like the general idea as discussed it on #1229. The approach in the PR makes the padding algorithm pluggable which is complicating the implementation. Could we instead have extent hard-coded to use the new padding component with its options rebound? 

I know we use the strategy concept in other places and I appreciate that you're trying to follow the existing pattern. However, I think the strategy concept makes sense when it is being applied multiple times to partial results within an algorithm. In this case it's only being used to post-process a single result, which is something that I think a consumer can easily wire up themselves.

N.B. I think with the above change the `hardLimitZero` implementation should simplify quite a bit.

Issue #1247, opened 2019-03-19
d3fc-axis site build is failing


ColinEberhardt:
See: https://d3fc.io/api/axis-api.html

ColinEberhardt:
fixed in #1248

PR #1248, opened 2019-03-19
fix: Fix problem generating axis documentation

andy-lee-eng:
It seems to fail because of the missing new-line after the final image link

(note the GitHub diff doesn't show the difference)!


andy-lee-eng:
Fix for #1247

andy-lee-eng:
If it's a fix to a bug with the documentation, should it be "fix:" or "docs:"?

ColinEberhardt:
> It seems to fail because of the missing new-line after the final image link

Great - thanks for fixing that. Out of interest, how did you uncover this issue? And is it something we can make easier to catch next time?

>  should it be "fix:" or "docs:"?

It should be `docs` - but don't worry about it, I can change the message when I merge

andy-lee-eng:
I removed all the recent changes and spent ages trying to add it back a bit at a time to try to find out what caused the problem, until eventually I'd added it all back and the only thing left was the line feed at the end :-(

However, your comment prompted me to go and have a look at what the build is actually doing...

In `/builder/parseReadmes.js`, there's a `createBlockPattern` regex that ends in "\\n$", which means it only matches documents that end in a new-line. If I simply remove the "\\n" from this, then it's able to build the axis page without changes.

To catch this sort of error, we could add a check after calling `parseReadme`:

```javascript
        if (readme.structure.length === 0) {
          throw new Exception(`No content for Readme - ${readme.name}`);
        }
```

ColinEberhardt (approved):

ColinEberhardt:
> To catch this sort of error, we could add a check after calling parseReadme:

It would be great if you could add a more long-term fix. I have a feeling that this will trip us up again in the future!

Thanks :-)

andy-lee-eng:
I suppose that changing the regex could potentially introduce errors elsewhere, but I've had a quick scan through and couldn't find any. That would imply that we have "blocks" of ReadMe files that we intentionally exclude on the basis that they don't end in "\n". That seems unlikely but I can't rule it out, because the same regex is used against all the parts of a file as well as the whole file itself.

Introducing that error check would mean we get a build failure if any of the ReadMe files produce no content, which might help catch problems like this, but wouldn't tell you what was wrong. I guess regex isn't exactly well known for providing good error feedback!

On a related note, I noticed that the d3fc-scripts package produces a "Scripts" entry in the side-nav, but it is empty when you click on it. Perhaps we should filter out packages that produce no content?

ColinEberhardt:
> Perhaps we should filter out packages that produce no content?

Good idea - I hadn't spotted that

PR #1253, opened 2019-03-25
docs: switch default export to named

Ro4052:
I think I noticed an issue with the d3fc extent docs. `extentLinear` should be a named import.


ColinEberhardt (approved):

ColinEberhardt:
You're absolutely right - thanks for the fix

Issue #1160, opened 2018-08-09
Canvas equivalents?


ColinEberhardt:
_From @chrisprice on August 18, 2016 14:51_

Currently all components are implemented as SVG only components.

_Copied from original issue: d3fc/d3fc-annotation#2_

ColinEberhardt:
_From @ezekg on August 2, 2017 18:19_

What's the status on this? Anything I can do to help out?

ColinEberhardt:
The status is pretty simple, we have't got around to it yet!

> Anything I can do to help out?

Any help would be appreciated. If you just want to try your hand at implementing one of the annotation types that would be great. The canvas implementations within d3fc-series might provide some good patterns to follow.

ColinEberhardt:
_From @utdrmac on January 24, 2018 3:6_

@ColinEberhardt is this still work in progress? I just spent hours designing my app using fc.seriesCanvas* only to get black-flagged on the last lap?

ColinEberhardt:
Hi @utdrmac good point, this has been on the 'to do' list for a little too long. I'll see if I can get this moving. Just out of interest, which specific annotations are you using?

ColinEberhardt:
_From @utdrmac on January 24, 2018 16:37_

gridlines, x/y axis, crosshairs. financial-style app using CanvasCandlesticks().

ColinEberhardt:
_From @ezekg on January 24, 2018 16:44_

Sorry I never got to contributing to this issue. I ended up drawing my annotations manually using the native canvas API, due to time constraints.

Ro4052:
When writing a canvas implementation for the crosshair annotation, I've noticed a fundemental difference with `decorate` for the canvas version. The only way to achieve the same flexibility as the SVG version would be to have a `decorate` function for each line and one for the point. Would @chrisprice or @ColinEberhardt have any opposition to something like this? The current idea would be to have `xDecorate` for the vertical line, `yDecorate` for the horizontal line and `decorate` for the point. I'll move forward with this approach unless anyone has any objections.

PR #1255, opened 2019-04-03
feat: canvas band annotation

Ro4052:
In reference to #1160. This PR implements a canvas version of the band annotation.

### DIFF
--- a/packages/d3fc-annotation/src/canvas/band.js
+++ b/packages/d3fc-annotation/src/canvas/band.js
@@ -8,6 +8,7 @@ export default () => {
     let orient = 'horizontal';
     let fromValue = d => d.from;
     let toValue = d => d.to;
+    let decorate = () => {};
     let context = null;
 
     var instance = (data) => {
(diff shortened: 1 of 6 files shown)


packages/d3fc-annotation/src/canvas/band.js
```diff
@@ -0,0 +1,101 @@
...
+        const valueScale = horizontal ? yScale : xScale;
+        const crossScaleRange = crossScale.range();
+
+        data.forEach((d, i) => {
+            context.save();
+            context.beginPath();
+            context.strokeStyle = 'transparent';
+
+            decorate(context, d, i);
+            const x = horizontal ? crossScaleRange[0] : valueScale(fromValue(d));
```
andy-lee-eng: As discussed, Ronan will look into whether it makes sense to use the `band` shape from d3fc-shapes.

andy-lee-eng (approved):
Looks good, and feels more in line with the other components this way.

ColinEberhardt (approved):

PR #1257, opened 2019-04-09
Allow chart axis to set how much width / height it needs

andy-lee-eng:
Introduces a new "initialise" event to the d3fc-element, allowing it to allocate the amount of space it needs.

"initialise" is called a second time on any for any element that has had it's available space adjusted by other components in the group (e.g. for X and Y axis affecting each other).

Implemented a new property `labelRotate` of the `axisOrdinal` component. Can be set to "auto" to automatically rotate the labels based on the available space, or else to a number to fix the rotation angle (in degrees).

Examples included: `d3fc-chart/examples/ordinal.html` and `d3fc-axis/examples/rotate.html`

### DIFF
--- a/packages/d3fc-chart/examples/ordinal.html
+++ b/packages/d3fc-chart/examples/ordinal.html
@@ -11,6 +11,12 @@
     <script src="../../d3fc-shape/build/d3fc-shape.js"></script>
     <script src="../../d3fc-annotation/build/d3fc-annotation.js"></script>
     <script src="../build/d3fc-chart.js"></script>
+    <style>
+        .x-axis svg text {
+            font-size: 20px;
+            fill: red;
+        }
+    </style>
 </head>
 <body>
     <div id="ordinal" style="width: 100%; height: 400px"></div>
--- a/packages/d3fc-chart/examples/ordinal.html
+++ b/packages/d3fc-chart/examples/ordinal.html
@@ -13,7 +13,7 @@
     <script src="../build/d3fc-chart.js"></script>
 </head>
 <body>
-    <div id="ordinal" style="width: 500px; height: 250px"></div>
+    <div id="ordinal" style="width: 100%; height: 400px"></div>
     <script src="ordinal.js"></script>
 </body>
 </html>
--- a/packages/d3fc-chart/examples/ordinal.js
+++ b/packages/d3fc-chart/examples/ordinal.js
@@ -60,7 +60,7 @@ const customAxis = (scale) => {
             s.select('text')
                 .style('text-anchor', rotate ? 'end' : 'middle')
                 .attr('transform', rotate ? `translate(0, 3) rotate(-${Math.floor(90 * rotate)} 3 ${Math.floor(labelHeight / 2)})` : 'translate(0, 8)');
-            decorate();
+            decorate(s);
         });
         base(selection);
     }
--- a/packages/d3fc-chart/examples/ordinal.js
+++ b/packages/d3fc-chart/examples/ordinal.js
@@ -1,18 +1,18 @@
 var data = [
     {
-        'month': 'Jan',
+        'month': 'January',
         'sales': 1
     },
     {
-        'month': 'Feb',
+        'month': 'February',
         'sales': 1.5332793661950717
     },
     {
-        'month': 'Mar',
+        'month': 'March',
         'sales': 2.0486834288742597
     },
     {
-        'month': 'Apr',
+        'month': 'April',
         'sales': 2.556310832331535
     },
     {
@@ -20,34 +20,67 @@ var data = [
         'sales': 3.029535759511747
     },
     {
-        'month': 'Jun',
+        'month': 'June',
         'sales': 3.507418002703505
     },
     {
-        'month': 'Jul',
+        'month': 'July',
         'sales': 4.02130992651795
     },
     {
-        'month': 'Aug',
+        'month': 'August',
         'sales': 4.482485234741706
     },
     {
-        'month': 'Sep',
+        'month': 'September',
         'sales': 4.957935275183866
     },
     {
-        'month': 'Oct',
+        'month': 'October',
         'sales': 5.427273488256043
     },
     {
-        'month': 'Nov',
+        'month': 'November',
         'sales': 5.943007604008045
     },
     {
-        'month': 'Dec',
+        'month': 'December',
         'sales': 6.454464059891373
     }
 ];
+
+const customAxis = (scale) => {
+    const base = fc.axisOrdinalBottom(scale);
+    let rotate = false;
+    let decorate = () => {};
+
+    function axis(selection) {
+        base.decorate(function(s) {
+            s.select('text')
+                .style('text-anchor', rotate ? 'end' : 'middle')
+                .attr('transform', rotate ? 'rotate(-45 5 5)' : 'translate(0, 3)');
+            decorate();
+        });
+        base(selection);
+    }
+
+    axis.height = () => {
+        const width = scale.range()[1];
+        rotate = width < 600;
+        return rotate ? '60px': '1em';
+    }
+
+    axis.decorate = (...args) => {
+        if (!args.length) {
+            return decorate;
+        }
+        decorate = args[0];
+        return axis;
+    };
+
+    fc.rebindAll(axis, base, fc.exclude("decorate"));
+    return axis;
+};
 
 var yExtent = fc.extentLinear()
     .accessors([d => d.sales])
@@ -57,10 +90,13 @@ var bar = fc.seriesSvgBar()
     .crossValue(d => d.month)
     .mainValue(d => d.sales);
 
-var chart = fc.chartCartesian(
-        d3.scalePoint().padding(0.5),
-        d3.scaleLinear()
-    )
+var chart = fc.chartCartesian({
+        xScale: d3.scalePoint().padding(0.5),
+        yScale: d3.scaleLinear(),
+        xAxis: {
+            bottom: customAxis
+        }
+    })
     .xLabel('Value')
     .yLabel('Sine / Cosine')
     .yOrient('left')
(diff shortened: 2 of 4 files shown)


packages/d3fc-axis/src/axisOrdinal.js
```diff
@@ -37,10 +40,117 @@ const axisOrdinal = (orient, scale) => {
         return { offset: [0, base.tickPadding()] };
     };
 
+    const decorateRotation = s => {
```
ColinEberhardt: this logic is currently specific to ordinal axes, it would be good to have the option to rotate linear / continuous axes

packages/d3fc-axis/src/axisOrdinal.js
```diff
@@ -37,10 +40,117 @@ const axisOrdinal = (orient, scale) => {
...
+            const offset = sign() * Math.floor(labelHeight / 2);
+            const offsetTransform = isVertical() ? `translate(${offset}, 0)` : `translate(0, ${offset})`;
+
+            text.style('text-anchor', anchor)
+                .attr('transform', `${existingTransform} ${offsetTransform} rotate(${rotate} 0 0)`);
+        }
+    };
+
+    const measureLabels = s => {
+        const labels = scale.domain();
```
ColinEberhardt: `scale` is a global variable

packages/d3fc-axis/src/axisOrdinal.js
```diff
@@ -37,10 +40,117 @@ const axisOrdinal = (orient, scale) => {
...
+
+        // Use a test element to measure the text in the axis SVG container
+        s.attr('font-size', 10).attr('font-family', 'sans-serif');
```
ColinEberhardt: is this styling necessary?

packages/d3fc-axis/src/axisOrdinal.js
```diff
@@ -37,10 +40,117 @@ const axisOrdinal = (orient, scale) => {
...
     const base = axisBase(orient, scale, {labelOffset, tickPath});
 
+    base.decorate(s => {
+        decorateRotation(s);
+        decorate(s);
+    });
+
     const axis = (selection) => {
-        return base(selection);
+        if (!rotation) calculateRotation(selection);
```
ColinEberhardt: what's the reason behind intercepting the 'render' function to cache the rotation value, rather than just decorating?

ColinEberhardt (changes requested):
Thanks @DevAndyLee - functionally this works very well. I've added a few comments to the code, and also experimented with taking your 'core' logic and refactoring a little. 

I think this works better as an adapter (similar to `seriesAutoBarWidth`). For example:

```
fc.axisLabelRotate(fc.axisOrdinalTop(foodScale))
```

My refactored version is here:
https://github.com/ColinEberhardt/d3fc/commit/723c64f1eca454cdea80bcc4ee05e267331ba5e4#diff-77615f2bd2745c8ced56f511942aa36f

ColinEberhardt:
Just spotted that SVG elements end up with negative heights / widths if you make the browser window very small.

packages/d3fc-axis/src/axisOrdinal.js
```diff
@@ -37,10 +40,117 @@ const axisOrdinal = (orient, scale) => {
         return { offset: [0, base.tickPadding()] };
     };
 
+    const decorateRotation = s => {
+        if (labelRotate) {
+            const { rotate, labelHeight, anchor } = rotation;
+
+            const text = s.select('text');
+            const existingTransform = text.attr('transform');
```
ColinEberhardt: just spotted this - grabbing an existing attribute from the DOM, and modifying / supplementing it feels like risky business. When re-applying this below, I guess this is via the D3 update selection? should it also be applied to enter selections?

packages/d3fc-axis/src/axisOrdinal.js
ColinEberhardt: This is an ordinal-specific way of obtaining the labels. Is there a generic approach that works for all scale types?

andy-lee-eng:
I have your branch now, and it's working well. Here are a few replies and notes:

**shaded backgrounds in the example**
The reason I had that is that I wanted to test that the axis was correctly reporting its desired width (when vertical) or height (when horizontal), so that it would play nicely with my enhancements to the cartesian chart, which allows it to give more width/height to those axes if they ask for it.
At the moment, the adapter doesn't implement width or height properties like my modified ordinalAxis did, so this doesn't work. I'll try adding it and see how that looks.

**caching the rotation value rather than just decorating**
This links to the above one. If we're doing an initial measure to set the axis sizes prior to doing a render, you end up having to measure the text sizes twice. I was trying to avoid that, since I thought it might be costly if there were a large number of them.

**font-size / font-family styling**
This was necessary to give the test element the correct font size so that it measures the text correctly.
`axisBase` sets the font styles in code when it renders: [axisBase.js#74](https://github.com/ColinEberhardt/d3fc/blob/axis-rotation/packages/d3fc-axis/src/axisBase.js#L74)
It would seem better to have those style rules in css, but the axis component doesn't have its own css (the cartesian chart does). Alternatively, perhaps expose the styling as a property that the adapter can use (like you did with `orient()`)?

andy-lee-eng:
Shall I begin be re-organising this PR so that it's just the chart/element "initialise" event, that allows the chart to get the desired width/height of the vertical/horizontal axis?

i.e. remove the changes to the ordinalAxis so that can be a separate PR?

andy-lee-eng:
I've changes this PR so that it is just:

- new "initialise" event in `d3fc-element`, allowing elements in a group a chance to negotiate what size they want to be. It gets called a second time for elements that have had their available space changed after the first time (e.g. x-axis gets less room because y-axis asked for more).
- changes to `cartesianChart` so that it asks the vertical axis what `width` it wants to be and the horizontal axis what `height` it wants to be (if implemented).
- add an example custom horizontal axis that rotates labels, taking up additional more height.

andy-lee-eng:
Question:

In my current implementation, `cartesianChart` checks for a function `width()` on the vertical axis and calls it if it exists, and does the same for `height()` on the horizontal axis.

I did consider having a single function called `measure()` that would measure the width for vertical axis and the height for a horizontal axis. That would simplify the implementation of the axis, but seemed a little confusing thanks to the existing "measure" event, which actually does something different (in the "measure" event, it's being told how much space it has, rather than being asked how much space it needs).

Do you think `measure()` would be better than `width()`/`height()`? Or perhaps make it explicit that they're not properties, e.g. `getWidth()`/`getHeight()` or `measureWidth()`/`measureHeight()`...

chrisprice:
Really like the example, that's very cool. Are we allowed a proper version of that?!

I took a look over the `d3fc-element` changes and I think there might be some confusion about how that's meant to work (which probably means I've done a bad job of documenting it). I'll have a go at knocking together an example of an alternative that avoids adding any new phases to the event lifecycle.

EndilWayfare:
Hey, is there any desire to keep moving on this? I've been having my own fights with `d3-element` with exactly the same problem: what if a component wants to change its width/height? I've got a custom legend component that sits below the x-axis, so this feature definitely generalizes beyond axes; there's a genuine use case.

I've got an (admittedly hacky) solution that uses an `onResize` callback into the component with a property. It just says "hey, I want to be this size now" and allows the chart author to decide what to do with that info. It never occurred to me to actually augment the event model of `d3-element`, but that's clearly a better strategy.

EndilWayfare:
> **caching the rotation value rather than just decorating**
> This links to the above one. If we're doing an initial measure to set the axis sizes prior to doing a render, you end up having to measure the text sizes twice. I was trying to avoid that, since I thought it might be costly if there were a large number of them.

Had the same thought. In my case, the legend does manual wrapping layout by measuring labels and figuring out how many of them will fit on each line, given the fixed width from `measure`. Doing the full layout calculation is necessary to determine height. Redoing the same calculations just to then `dataJoin` and update properties feels wrong to me. But maybe the general measure/update cycle is more important and this is a necessary concession to the model.

EndilWayfare:
> **font-size / font-family styling**
> This was necessary to give the test element the correct font size so that it measures the text correctly.
> `axisBase` sets the font styles in code when it renders: [axisBase.js#74](https://github.com/ColinEberhardt/d3fc/blob/axis-rotation/packages/d3fc-axis/src/axisBase.js#L74)
> It would seem better to have those style rules in css, but the axis component doesn't have its own css (the cartesian chart does). Alternatively, perhaps expose the styling as a property that the adapter can use (like you did with `orient()`)?

I came across a similar question of styling from a slightly different angle. I wanted a method to measure text that didn't touch the DOM (to avoid layout thrashing). My naive `textMeasurer` component:
```js
const textMeasurer = () => {
  const canvas = create("canvas").node();
  const context = canvas.getContext("2d");

  const instance = {};

  instance.font = (...args) => {
    if (!args.length) {
      return context.font;
    }
    context.font = args[0];
    return instance;
  };

  instance.measureText = (text) => {
    return context.measureText(text);
  };

  return instance;
};
```
I'm sure there are edge cases that I don't have the experience to identify (e.g. perhaps `devicePixelRatio` needs to be accounted for).

Since style can come from so many different sources, to make sure the measurement uses the actual style in use, what if we just asked the element what font style it has?
```js
// Measurer can live in "outer closure" of component
const measurer = textMeasurer()

...

const instance = (selection) => {
...
  const selectionNode = selection.node();
  const document = selectionNode.ownerDocument;
  const window = document.parentWindow || document.defaultView;
  // Node not added to document, so shouldn't trigger reflow
  const dummy = create("div").node();
  dummy.style.setProperty(
    "font",
    // Captures style cascaded from external sources
    window.getComputedStyle(selectionNode).font
  );
  // Example of applying local style overrides. Here, the hypothetical component has a `fontSize()` property
  dummy.style.setProperty("font-size", fontSize + "px");
  measurer.font(dummy.style.font);
...
}
```

Issue #1262, opened 2019-05-02
Discussion about chartCartesian


Ro4052:
#### Styling Differences

 - [x] check that with cartesian charts it is still relatively easy to rotate a y axis label vertically.
 - [x] migrate the various package / dev examples to `chartCartesian`
 - [x] migrate the website examples to `chartCartesian`
 - [x] migrate the various D3 blocks examples to `chartCartesian`
 - [x] re-add the chart label (based on creating examples) - see related issue comment https://github.com/d3fc/d3fc/issues/1259#issuecomment-490574556
 - [x] check whether we can solve much of the above by capitalising on CSS grid, if so, can we just add a few examples?

#### Canvas plot area issues

 - [x] adopt the approach of defaulting the canvas / svg plot area components to `null`.
 - [x] double check that a chart without any canvas / svg series still renders OK
 - [x] ensure we add these components via a data-join, not just the enter selection
 - [x] move the canvas to the back?
 - [ ] enhance the cartesian chart docs, we really need to illustrate some more of the common use cases

There are couple of points that @DevAndyLee and I believe are worth discussing about the implementation of the `chartCartesian` component. The first point regards styling differences between `chartCartesian` and the legacy components and the second concerns the canvas plot area still being present in a purely SVG chart.

### Styling Differences

The legacy components use `cartesianBase` which has a fair amout of code that is not applied to the charts produced by `chartCartesian`. This seems to be mostly styling that is handled in css but there is one more noticeable difference in that the legacy components rotate the y labels. This means they are parallel to the y axis and take up less space, the same is not applied to `chartCartesian`.

### Canvas Plot Area

This issue is a little trickier. In many cases, users will want to draw a chart using only SVG. One of the benefits of SVG is that it is very easy to interact with. However, with `chartCartesian`, a canvas plot area is placed over the top, meaning it must be hidden/removed by the user in order to interact with the SVG plot. This isn't necessarily a huge issue but is less obvious with the legacy `chartSvgCartesian` and `chartCanvasCartesian` components. As the user specifies what plot area they want, it would be reasonable to expect that the other plot area would be hidden/removed automatically.

[...]

PR #1266, opened 2019-05-14
chore: migrate examples to chartCartesian

Ro4052:
In relation to #1262, migrated examples to use `chartCartesian`.

### DIFF
--- a/packages/d3fc-brush/examples/brush.js
+++ b/packages/d3fc-brush/examples/brush.js
@@ -49,11 +49,11 @@ var multi = fc.seriesSvgMulti()
         }
     });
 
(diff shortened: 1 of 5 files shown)


ColinEberhardt (commented):
Thanks @Ro4052 - can you also migrate the website examples too? For example: https://github.com/d3fc/d3fc/blob/master/packages/d3fc-site/src/examples/simple/simple.js

ColinEberhardt:
> Thanks @Ro4052 - can you also migrate the website examples too?

Actually, before moving on this, we probably need to resolve some of the issues around #1262 involving the y label placement and chart label

PR #1267, opened 2019-05-16
feat: add y label container to chartCartesian

Ro4052:
I'm sure this needs some refactoring but this PR moves the `y-label-container` div from `cartesianBase` to `chartCartesian`, in reference to #1262, for easier styling of the label.

### DIFF
--- a/packages/d3fc-chart/src/cartesian.js
+++ b/packages/d3fc-chart/src/cartesian.js
@@ -34,6 +34,7 @@ export default (...args) => {
         .key(d => d);
     const xLabelDataJoin = dataJoin('div', 'x-label')
         .key(d => d);
+    const yContainerDataJoin = dataJoin('div', 'y-label-container');
     const yLabelDataJoin = dataJoin('div', 'y-label')
         .key(d => d);
 
@@ -58,8 +59,16 @@ export default (...args) => {
                 .attr('class', d => `x-label ${d}-label`)
                 .text(xLabel(data));
 
-            yLabelDataJoin(container, [yOrient(data)])
-                .attr('class', d => `y-label ${d}-label`)
+            const yOrientValue = yOrient(data);
+            const yLabelContainer = yContainerDataJoin(container, [yOrientValue])
+                .attr('class', d => `y-label-container ${d}-label`)
+                .style('grid-column', yOrientValue === 'left' ? 1 : 5)
+                .style('-ms-grid-column', yOrientValue === 'left' ? 1 : 5)
+                .style('grid-row', 3)
+                .style('-ms-grid-row', 3);
+
+            yLabelDataJoin(yLabelContainer, [yOrientValue])
+                .attr('class', 'y-label')
                 .text(yLabel(data));
 
             xAxisDataJoin(container, [xOrient(data)])
--- a/packages/d3fc-chart/src/cartesian.js
+++ b/packages/d3fc-chart/src/cartesian.js
@@ -34,8 +34,7 @@ export default (...args) => {
         .key(d => d);
     const xLabelDataJoin = dataJoin('div', 'x-label')
         .key(d => d);
-    const yContainerDataJoin = dataJoin('div', 'y-label-container');
-    const yLabelDataJoin = dataJoin('div', 'y-label')
+    const yContainerDataJoin = dataJoin('div', 'y-label-container')
         .key(d => d);
 
     const propagateTransition = maybeTransition => selection =>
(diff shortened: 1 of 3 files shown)


ColinEberhardt:
Can you also rotate the y axis label as per default? My preference is to render the most conventional chart when using the default settings. A thoroughly unscientific review of defaults, by [Google Image Search](https://www.google.com/search?safe=strict&biw=1427&bih=774&tbm=isch&sa=1&ei=a2jdXIWnJ6G5gwep47PoCw&q=chart&oq=chart&gs_l=img.3..0l10.4188.4188..4547...0.0..0.135.135.0j1......1....1..gws-wiz-img.X_ZwhT0Litw) shows that people typically render y axis labels vertically

Ro4052:
> Can you also rotate the y axis label as per default? My preference is to render the most conventional chart when using the default settings. A thoroughly unscientific review of defaults, by [Google Image Search](https://www.google.com/search?safe=strict&biw=1427&bih=774&tbm=isch&sa=1&ei=a2jdXIWnJ6G5gwep47PoCw&q=chart&oq=chart&gs_l=img.3..0l10.4188.4188..4547...0.0..0.135.135.0j1......1....1..gws-wiz-img.X_ZwhT0Litw) shows that people typically render y axis labels vertically

You definitely can. The only reason I hadn't done so far was to keep the styling the same as before.

packages/d3fc-chart/src/cartesian.js
```diff
@@ -58,8 +59,16 @@ export default (...args) => {
...
-                .attr('class', d => `y-label ${d}-label`)
+            const yOrientValue = yOrient(data);
+            const yLabelContainer = yContainerDataJoin(container, [yOrientValue])
+                .attr('class', d => `y-label-container ${d}-label`)
+                .style('grid-column', yOrientValue === 'left' ? 1 : 5)
+                .style('-ms-grid-column', yOrientValue === 'left' ? 1 : 5)
+                .style('grid-row', 3)
+                .style('-ms-grid-row', 3);
+
+            yLabelDataJoin(yLabelContainer, [yOrientValue])
```
andy-lee-eng: It feels like we shouldn't need to do both a yContainerDataJoin and then yLabelDataJoin?

For example, could you do:
```javascript
  // add the .y-label to new y-label containers
  yLabelContainer.enter().append('div').class('y-label');
  // Alter the text on the y-label
  yLabelContainer.select('.y-label').text(yLabel(data));
```

packages/d3fc-chart/src/cartesian.js
```diff
@@ -58,8 +59,16 @@ export default (...args) => {
                 .attr('class', d => `x-label ${d}-label`)
                 .text(xLabel(data));
 
-            yLabelDataJoin(container, [yOrient(data)])
-                .attr('class', d => `y-label ${d}-label`)
+            const yOrientValue = yOrient(data);
+            const yLabelContainer = yContainerDataJoin(container, [yOrientValue])
+                .attr('class', d => `y-label-container ${d}-label`)
+                .style('grid-column', yOrientValue === 'left' ? 1 : 5)
```
andy-lee-eng: I think these grid-column styles shouldn't be needed - the css already has these for the `.left-label` and `.right-label` classes.

packages/d3fc-chart/src/cartesianBase.js
```diff
@@ -24,21 +24,13 @@ export default (setPlotArea, defaultPlotArea) =>
...
-                    .style('width', '1em')
+                container
+                    .select('div.y-label-container')
                     .style('display', 'flex')
                     .style('align-items', 'center')
                     .style('justify-content', 'center')
-                    .append('div')
-                    .attr('class', 'y-label')
+                    .style('width', '1em')
+                    .select('div.y-label')
```
andy-lee-eng: So if we add `transform: rotate(-90deg)` to the css as the default (like Colin suggested), we should be able to avoid any additional styling of .y-label here.

andy-lee-eng (changes requested):

packages/d3fc-chart/src/cartesian.js
Ro4052: I'll take a look, that didn't feel quite right.

packages/d3fc-chart/src/cartesianBase.js
Ro4052: I'll have a look at that too.

Ro4052: The default for `chartCartesian` is now to have the labels vertical (without clipping). It also seems to be very easy to make the label horizontal again using just CSS. I've tested this with axes on both sides of the chart.

andy-lee-eng (approved):

Ro4052:
I've just found this. Using the CSS below allows the user to get the same rotated y-label result with `chartCartesian` without the `y-label-container`.

```
.y-label {
    display: flex;
    transform: rotate(-90deg);
    width: 1em;
    white-space: nowrap;
    justify-content:center;
}
```

Ro4052:
Maybe the `y-container-label` is unecessary and I should open a new PR that applies the above styling by default?

ColinEberhardt:
superseded by #1268

PR #1271, opened 2019-05-21
feat: add chart label to chartCartesian

Ro4052:
This PR adds the chart label to `chartCartesian` as mentioned in #1262 and #1259.

### DIFF
--- a/packages/d3fc-chart/src/cartesianBase.js
+++ b/packages/d3fc-chart/src/cartesianBase.js
@@ -1,4 +1,4 @@
-import { rebindAll, include, rebind } from '@d3fc/d3fc-rebind';
+import { rebindAll, include } from '@d3fc/d3fc-rebind';
 import cartesianChart from './cartesian';
 
 const functor = (v) =>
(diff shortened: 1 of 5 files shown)


packages/d3fc-chart/src/cartesianBase.js
```diff
@@ -45,38 +43,15 @@ export default (setPlotArea, defaultPlotArea) =>
...
-                    .text(chartLabel(data));
-
                 decorate(container, data, index);
             });
 
             selection.call(cartesian);
         };
 
         rebindAll(cartesianBase, cartesian, include(/^x/, /^y/));
+        rebind(cartesianBase, cartesian, 'chartLabel');
```
andy-lee-eng: Can you do it in one line?

```javascript
rebindAll(cartesianBase, cartesian, include(/^x/, /^y/, 'chartLabel'));
```

andy-lee-eng (changes requested):
I think the chart documentation needs updating too - the `chartLabel` property should be added, but there's also a section about using decoration to add a chart label, which can be removed.

packages/d3fc-chart/src/cartesianBase.js
Ro4052: I was wondering about that but I wasn't sure.

andy-lee-eng (approved):

ColinEberhardt (approved):

PR #1275, opened 2019-05-22
fix: moved both axis to give them a higher z-order than the plot area

ColinEberhardt:
If you look at `examples/move-axis.js` it relies on the axes having a higher z-order than the plot area. I can't think of any example where you would want this the other way around.


ColinEberhardt:
@Ro4052 can you please take a look at this change?

Ro4052:
> @Ro4052 can you please take a look at this change?

My git diff makes it look like nothing has changed. It hasn't picked up any changes in `examples/move-axis.js`.

image

Ro4052 (approved):

PR #1282, opened 2019-06-05
feat: Implementation of WebGL based Point series with webglPlotArea

andy-lee-eng:
- Added WebGL setup and helper functions, plus two types of WebGL shader programs.
- Added "Point" series for WebGL rendering, which seems to be the most useful case.
- Added "Multi" and "Repeat" exports for WebGL, but these don't (currently) need to be different to the Canvas version.

I've also added examples and documentation for the new WebGL rendering, including a performance-test sample.

On that last point, `seriesWebglMulti` and `seriesWebglRepeat` allow you to render multiple `seriesWebglPoint` (e.g. with different colors and shapes). However, the actual implementation doesn't need to be any different than the Canvas one (at least, not right now).

I thought it would be important to export "Webgl" variants of these though, to make it clear that you can't mix-and-match Canvas and WebGL series in the same Multi (which wouldn't work). It also means that in the future, if we do need to modify the WebGL version, we can do so without making a breaking change.

Implementation wise, I could have made `seriesWebglMulti` and `seriesWebglRepeat` new components that have their "Canvas" counterparts as the base component. However, it seems to work just as well to simply export the same component twice under different names (and is less code).


andy-lee-eng:
Additional note:

I haven't implemented WebAssembly code for doing the heavy lifting triangle calculations (as described in my [blog article](https://blog.scottlogic.com/2019/05/28/drawing-2d-charts-with-webgl.html)).

However, I have structured the code so that we could easily swap the JavaScript implementation for a WebAssembly one if we want to (for about a 10%-25% performance boost). Basically everything in the `/webgl/points` folder.

I have experimented with building the WebAssembly functions in AssemblyScript (i.e. TypeScript-ish), which can be compiled by an npm task with no additional tooling installed. I've also got it bundling the .wasm into the JavaScript using base64 (33% bigger), so that the consumer doesn't have to worry about hosting the .wasm file somewhere and dynamically loading it at runtime.

This could be a future enhancement if interest. It could be a nice way to showcase some production WebAssembly code as part of D3FC!

MaxGraey:
@DevAndyLee About performance. Not sure how significantly this improve performance but I should note AssemblyScript can't optimize `Math.pow` for constant arguments yet. So better use simple squaring `x * x` [here](https://github.com/DevAndyLee/d3fc-webgl-as/blob/master/src/assembly/edges.ts#L31) instead `Math.pow(x, 2)`. Another suggestion is using `-O3` instead `--optimize` [here](https://github.com/DevAndyLee/d3fc-webgl-as/blob/master/package.json#L15)

andy-lee-eng:
Thanks @MaxGraey, I will try your suggestions.

The [d3fc-webgl-as](https://github.com/DevAndyLee/d3fc-webgl-as) project is experimental. I found about a 25% performance boost when creating the triangles needed for circles, but only about 5% for the simpler shapes (i.e. fewer triangles).

I haven't actually implemented any WebAssembly in this PR (not sure how you got here from there?). It's something I'd like to do though if there's interest.

ColinEberhardt:
Thanks for opening this PR @DevAndyLee this would be a great addition to D3FC.

I had a chat with @chrisprice earlier, he suggested that you should try offloading as much work as possible to the GPU. For example, are you performing the scale calculations in JS, wasm, or OpenGL?

I'm afraid I haven't had a chance to take a detailed look at the code yet, but would be interested in your opinions on whether there are any other changes that would make it even faster?

andy-lee-eng:
Offloading scale calculations is an interesting idea which I did actually consider (but haven't tried).

It might be possible to calculate a model projection matrix that represents the X and Y scales, then we could just pump the raw X/Y data into WebGL without transforming it. However, that assumes that they are both linear scales. I suspect they probably will be most of the time, but not necessarily, so the code would have to be able to work in both ways.

As I mentioned in the blog, I did want to generate all the triangles using a geometry shader (on the GPU), but those just don't exist in WebGL. That's why I first started looking into WebAssembly to try to speed up that process. Even without WebAssembly, the WebGL Points implementation is 5-10 times faster than the 2D canvas one.

I could look into offloading scale calculations, but I think it'll make the code a lot more complex, unless we just ban the use of non-linear scales for WebGL.

We need to do at least one pass of the data in JavaScript anyway, to call the `mainValue()` / `crossValue()` / `size()` properties for each entry prior to calculating the triangles, so I'm not sure that skipping the scale (which should be quite quick for a linear scale) will actually help save much time.

andy-lee-eng:
I've tested this idea (offloading scales) by using data that is already in the (-1 to 1) range then just not applying the scale functions.

In normal examples the difference is too small to measure. If I tailor the example specifically (e.g. high number of the simplest small triangles), I can measure as much as 2fps improvement at around 20fps, but that's a special case.

I'm not sure it's worth it given the compromises (limited to linear scales or else more complicated implementation).

chrisprice:
Sorry I don't have time to look into this properly at the moment but a
couple of suggestions -

Have you tried using one of the line modes rather than triangles?
https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/drawArrays

Whilst extracting the data from the js objects will always necessitate a js
pass when the data changes, I think it's still worth moving the rest to gl
to cover the other ways of interacting e.g. pan/zoom etc.

andy-lee-eng:
I have. The line modes only draw lines of 1 pixel thickness. To get more than 1 pixel you need a triangle
strip, which I have used for drawing a Line series (though a Line series isn't part of this PR so I haven't included it).

For the Point series, the triangle strip mode works for circles and other shapes, but you can only draw one at a time which means it ends up being much much slower.

I'm not sure how you visualise pan/zoom working, but in the current d3 zoom behaviour, it requires you to update the scales then re-render the cartesianChart object, which means you then have to re-process the data anyway (which may have also changed).

I am going to have a play with doing linear-scales-only as a projection matrix, but I don't expect that to bring any noticeable performance gains. Instead, it might mean we can have a special option for the Point series that lets you pass the data in as an already-prepared Float32Array.

andy-lee-eng:
I think it would make sense to add any special-mode optimisations (like Float32Array input) as a separate PR, since it will involve multiple render paths, and thus make this PR more complex.

ColinEberhardt:
@DevAndyLee FYI the build is failing:

> 'npm ERR! Cannot read property \'match\' of undefined\n\nnpm ERR! A complete log of this run can be found in:\nnpm ERR!     /home/travis/.npm/_logs/2019-06-05T10_48_24_231Z-debug.log\n',

I've checked out this branch to have a play / poke-around!

ColinEberhardt:
Random brain-dumped thoughts ...

 - Should the code in `webgl/points/circles.js` and `webgl/points/shapes.js` live in d3fc-shape?
 - Are we bundling gl-matrix? If so, what is the 'cost' of this?
 - Is the number of triangles used to render a circle configurable? Should it be? 
 - As this is a bit more complex than the typical d3fc conventional data-joined components, it would be good to have a few comments in the code regarding the role of the various components in the `points` subfolders

andy-lee-eng:
I've removed all the triangle processing stuff, and it now passes the x/y/size array directly into the GPU buffer and either uses a radius based render (circles) or a texture (other shapes).

I've also removed the gl-matrix reference, and we manually do the (very simple) transform in the vertex shader instead.

andy-lee-eng:
I've implemented `Line`, `Bar` and `Area` series types in a different branch:
https://github.com/DevAndyLee/d3fc/tree/feature/webgl-other-series

I was still hoping we might be able to move this branch forwards before I over-complicate it.

The new branch also comes with a performance-based example "webgl-series.html" that showcases all three series types, with a zoomable date/time x-axis.
200,000 data points are shown by default (which represents 1 per hour in 5 series over 4.5 years).
You can still see the Canvas version of the same to compare performance - though I'd recommend dropping the data size first! :-)

andy-lee-eng:
Additional note: the Area and Bar series can't use the super-speed direct ArrayBuffer to GPU mode (because they always have to create areas from data points).

Therefore, I have implemented a `cacheEnabled(true)` property, which will allow it to cache the generated GPU data during re-render (i.e. pan and zoom - though Bar still has to regenerate when zooming because the bar width changes).

ColinEberhardt:
This is likely going to be superseded with an implementation that re-thinks more aspects of the D3FC design in order to fully capitalise on the potential performance improvements that might be possible with WebGL.

Issue #1284, opened 2019-06-24
top-label keeps adding up when re-rendering chart.


Salketer:
There is a div with class top-label that keeps adding itself when the chart is re-rendered.

Salketer:
See reproduction on this example:

https://bl.ocks.org/ColinEberhardt/3c780088c363d1515403f50a87a87121

ColinEberhardt:
fixed by #1286

PR #1285, opened 2019-06-24
Use classed instead of  attr('class')

Salketer:
This seems that using the attr('class') was removing the "chart-label" applied on the dataJoin. This resulted in recreating the container every time... (Be advised I'm clueless as to why, I just know it works...)

### DIFF
--- a/packages/d3fc-chart/src/cartesian.js
+++ b/packages/d3fc-chart/src/cartesian.js
@@ -54,9 +54,11 @@ export default (...args) => {
             container.enter()
                 .attr('auto-resize', '');
 
-            chartLabelDataJoin(container, [xOrient(data)])
-                .attr('class', d => d === 'top' ? 'bottom-label' : 'top-label')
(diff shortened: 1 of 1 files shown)


Salketer:
Fixes #1284

ColinEberhardt:
Thanks for raising the issue and your proposed fix, much appreciated. I'll look at this shortly :-)

ColinEberhardt:
Nice one, thanks @Salketer - I've worked out what the underlying problem is. The d3fc-data-join component relies on being supplied with a selector that uniquely identifies the elements it creates with the underlying d3 data join. In this case it uses the `chart-label` class. However, this class was accidentally removed with the `attr` code which you modified. As a result, a new div was created on each invocation of the data join.

I've modified your code to create a more efficient fix - see #1286

Salketer:
Cool! Thanks for the follow up. I wasn't confident enough with the code to
find the real problem, I just started with d3.

PR #1294, opened 2019-08-08
Add basic WebGL prototype, supporting point circle series only for now.

tommclaughlan:
The intent here is to put the current state of the WebGL work in the 'webgl' dev branch.


packages/d3fc-series/src/webgl/point.js
```diff
@@ -0,0 +1,83 @@
...
+
+        const xScale = scaleMapper(base.xScale());
+        const yScale = scaleMapper(base.yScale());
+
+        xScale.glScale.coordinate(0);
+        yScale.glScale.coordinate(1);
+
+        const accessor = getAccessors();
+
+        const points = new Float32Array(filteredData.length * 3);
```
hanna-greaves: nit - please extract this magic number to a `const `with a relevant name

packages/d3fc-webgl/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2015-2016 Scott Logic Ltd.
```
hanna-greaves, line 3: we should update the copyright to the current year : )

packages/d3fc-series/src/webgl/point.js
```diff
@@ -0,0 +1,83 @@
...
+        let pi = 0;
```
joelwilliamsmith: `pi` is a slightly strange name for this variable! I was expecting `let pi = 3.141...`

packages/d3fc-webgl/src/buffers/uniformBuilder.js
```diff
@@ -0,0 +1,56 @@
...
+            case 2:
+                gl.uniform2fv(loc, data);
+                break;
+            case 3:
+                gl.uniform3fv(loc, data);
+                break;
+            case 4:
+                gl.uniform4fv(loc, data);
+                break;
+            }
```
hanna-greaves, line 27: Default warning/error case would be nice

packages/d3fc-webgl/src/glDraw.js
```diff
@@ -0,0 +1,91 @@
...
+    let data = null;
+    let context = null;
+    let changed = false;
+    let program = programBuilder();
+    let xScale = glScaleBase();
+    let yScale = glScaleBase();
+    let decorate = () => {};
+
+    const firstDraw = () => {
+        program.buffers().attribute('aVertexPosition', attributeBuilder(data));
```
hanna-greaves: `aVertexPosition` is used twice, we should put it into a const, perhaps in a different file so that we could add any other shader name definitions in the same place

packages/d3fc-webgl/src/scale/scaleMapper.js
```diff
@@ -0,0 +1,32 @@
...
+import glScaleLinear from './glScaleLinear';
+import glScaleLog from './glScaleLog';
+import glScalePow from './glScalePow';
+
+export default (scale) => {
+    let outScale = d3Scale.scaleIdentity();
+    let glScale = null;
+    const scaleCopyString = scale.copy.toString();
+
+    if (scaleCopyString === d3Scale.scaleLinear().copy.toString()) {
```
hanna-greaves, line 13: Could you add a comment here when you get chance, I'm not sure I follow the logic of making all these string copies

hanna-greaves (approved):
Very cool, left a few minor comments. Would also recommend lowering the version number to <= 1.0.0 before we decide this is ready

packages/d3fc-webgl/src/scale/scaleMapper.js
tommclaughlan: Sure, effectively it's a bit of a hacky way of telling which type of d3 scale is passed in, since there's no `scale.name` / `scale.type` to compare against. `scale.copy` gives a function which is specific to the scale type - it's horrible but it works for now.

I'm very much eager to replace this with something nicer.

packages/d3fc-webgl/LICENSE
tommclaughlan: This might be better off done separately on all packages to avoid having different licenses scattered throughout the repo.

packages/d3fc-webgl/src/glDraw.js
tommclaughlan: I'd been toying with the idea of having some kind of metadata for shader snippets - essentially at the moment to use a shader snippet you need to know exactly which attributes/uniform names are required.
This may be a good solution for that.

tommclaughlan: In the meantime I'll just pull the name to a const though.

packages/d3fc-webgl/src/program/programBuilder.js
```diff
@@ -0,0 +1,129 @@
...
+        const vShader = loadShader(gl, vertexShader, gl.VERTEX_SHADER);
+        const fShader = loadShader(gl, fragmentShader, gl.FRAGMENT_SHADER);
+
+        const program = gl.createProgram();
+        gl.attachShader(program, vShader);
+        gl.attachShader(program, fShader);
+        gl.linkProgram(program);
+
+        if (!gl.getProgramParameter(program, gl.LINK_STATUS)) {
+            console.log(gl.getProgramInfoLog(program));
```
joelwilliamsmith: This will probably need to be removed at some point - is it useful to have it in now? If so, I think we should put a `TODO` next to it to flag for merging into master

packages/d3fc-webgl/src/program/programBuilder.js
```diff
@@ -0,0 +1,129 @@
...
+        return program;
+    }
+
+    function loadShader(gl, source, type) {
+        const shader = gl.createShader(type);
+        gl.shaderSource(shader, source);
+        gl.compileShader(shader);
+
+        if (!gl.getShaderParameter(shader, gl.COMPILE_STATUS)) {
+            console.log(gl.getShaderInfoLog(shader));
```
joelwilliamsmith: Same as above comment

joelwilliamsmith (approved):

packages/d3fc-webgl/src/glDraw.js
```diff
@@ -0,0 +1,93 @@
...
+        context = args[0];
+        return draw;
+    };
+
+    draw.xScale = (...args) => {
+        if (!args.length) {
+            return xScale;
+        }
+        xScale = args[0];
+        return xScale;
```
chrisprice: `xScale` -> `draw`

packages/d3fc-webgl/src/glDraw.js
```diff
@@ -0,0 +1,93 @@
...
+    };
+
+    draw.yScale = (...args) => {
+        if (!args.length) {
+            return yScale;
+        }
+        yScale = args[0];
+        return yScale;
```
chrisprice: `yScale` -> `draw`

packages/d3fc-webgl/src/program/programBuilder.js
```diff
@@ -0,0 +1,127 @@
...
+    let program = null;
+    let vertexShader = null;
+    let fragmentShader = null;
+    let numElements = null;
+    let mode = 4; // gl.TRIANGLES
+    let first = 0;
+    let buffers = bufferBuilder();
+
+    const build = (gl) => {
+        if (newProgram(gl, program, vertexShader(), fragmentShader())) {
```
chrisprice: double evaluation of shaders

packages/d3fc-webgl/src/program/programBuilder.js
```diff
@@ -0,0 +1,127 @@
...
+            gl.deleteProgram(program);
+            program = createProgram(gl, vertexShader(), fragmentShader());
```
chrisprice: see above

packages/d3fc-webgl/src/program/programBuilder.js
```diff
@@ -0,0 +1,127 @@
...
+        const fShader = loadShader(gl, fragmentShader, gl.FRAGMENT_SHADER);
+
+        const program = gl.createProgram();
+        gl.attachShader(program, vShader);
+        gl.attachShader(program, fShader);
+        gl.linkProgram(program);
+
+        if (!gl.getProgramParameter(program, gl.LINK_STATUS)) {
+            gl.deleteProgram(program);
+            throw new Error(`Failed to link program : ${gl.getProgramInfoLog(program)}`);
```
chrisprice: You can't call `getProgramInfoLog` on a deleted program. Also can we include the source in the thrown error

packages/d3fc-webgl/src/program/programBuilder.js
```diff
@@ -0,0 +1,127 @@
...
+        return program;
+    }
+
+    function loadShader(gl, source, type) {
+        const shader = gl.createShader(type);
+        gl.shaderSource(shader, source);
+        gl.compileShader(shader);
+
+        if (!gl.getShaderParameter(shader, gl.COMPILE_STATUS)) {
+            gl.deleteShader(shader);
```
chrisprice: as above

packages/d3fc-webgl/src/glDraw.js
```diff
@@ -0,0 +1,93 @@
...
+            program.buffers().attribute(vertexAttribute).data(data);
+        }
+
+        program(context);
+    };
+
+    let drawFn = firstDraw;
+    const draw = () => {
+        context.viewport(0, 0, context.canvas.width, context.canvas.height);
+        program.apply(xScale);
```
chrisprice: Could we apply the co-ordinate to the scale at this point?

packages/d3fc-webgl/src/buffers/attributeBuilder.js
```diff
@@ -0,0 +1,86 @@
...
+        const dataType = getGLType(data.constructor, gl);
+
+        buffer = buffer || gl.createBuffer();
+
+        const location = gl.getAttribLocation(program, name);
+        gl.bindBuffer(gl.ARRAY_BUFFER, buffer);
+
+        gl.vertexAttribPointer(
+            location,
+            numComponents || data.length / numElements,
```
chrisprice, line 33: If we remove the concept of `numElements` here, would that mean we could get rid of `numElements` globally?

I've just been caught out by the above calculation. Specifically I was creating a buffer like this -

```
var buffer = new Float32Array(data.length);
```

Rather than -

```
var buffer = new Float32Array(data.length * 4);
```

Combined with -

```
draw.program().numElements(data.length);
```

packages/d3fc-webgl/src/scale/scaleMapper.js
```diff
@@ -0,0 +1,34 @@
+import d3Scale from 'd3-scale';
+import glScaleLinear from './glScaleLinear';
+import glScaleLog from './glScaleLog';
+import glScalePow from './glScalePow';
+
+export default (scale) => {
```
chrisprice, line 6: These will always be full range. Not sure what the alternative is but it feels like we need to record that somewhere

packages/d3fc-webgl/src/shaders/vertexShaderSnippets.js
```diff
@@ -0,0 +1,22 @@
...
+
+export const multiColor = {
+    header: `attribute vec4 aColor;
+             varying vec4 vColor;`,
+    body: `vColor = aColor;`
+};
+
+export const point = {
+    header: `varying float vSize;`,
+    body: `vSize = aVertexPosition.z;
```
chrisprice: Is there an advantage to using a 4-component vector to store x, y & size?

packages/d3fc-webgl/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "license": "MIT",
+  "keywords": [
+    "d3",
+    "d3fc",
+    "webgl"
+  ],
+  "main": "build/d3fc-webgl.js",
+  "module": "index",
+  "scripts": {
+    "bundle": "d3fc-scripts bundle",
```
joelwilliamsmith, line 14: Flagging this now - will need to change once master is webgl branch is rebased off master

joelwilliamsmith (approved):

PR #1295, opened 2019-08-09
Repository Caretaking

hanna-greaves:
# Monorepo Cleanup PR

This PR attempts to clean up some of the miscellaneous leftovers from when the repository was combined into a monorepo and to normalize the procedures for building, bundling and testing the packages within.

_Unfortunately this has become quite a large PR so I will provide a summary of changes below:_

1. __Remove `d3fc-scripts`.__
    This package was unusual in that it was a package of script files which generated the build and perform tests. Given that we are now in a monorepo, there was no particular need for this. It has been removed.
      * The `bundle.js` script has been replaced with using the `rollup` cli.
      * The `required` folder has not been replaced, I would suggest any file requirements be enforced by travis.
      * The `lint.js` script has been replaced with using the `eslint` and `markdownlint` cli
      * The `screenshot.js` script has been moved to a `util` folder.

2. __Add `script` folder__
   A folder has been added to the project root (as opposed to the packages directory) which contains key scripts and utilities used by all projects. This includes 
      * `jasmine.json` test config file 
      * `rollup.config.js` bundle config file 
      * The aforementioend `screenshot.js` screenshotting tool

3. __Remove duplicate files__
   The files `.npmignore`, `.gitignore`, `.eslintrc.json`, and `LICENSE` were duplicated with the same content in each package. A copy of each of these files has been moved to the project root, and their duplicates removed. 

4. __Enable Hot Module Replacement and Realtime Dev Mode__
   This PR also enables a better developer environment. After running `npm i` from the project root, you can call `npm start` which will bundle the packages and launch a developer sandbox in your default web browser which will automatically refresh whenever you save a change to any of the packages. See below:
demo


util/screenshot.js
```diff
@@ -0,0 +1,79 @@
...
+
+const server = http.createServer(function(req, res) {
+    const done = finalhandler(req, res);
+    serve(req, res, done);
+});
+
+server.listen(port, () =>
+  takeScreenshots()
+    .then(() => {
+        console.error('Screenshots complete 📷');
```
tommclaughlan: Should this be `console.log` rather than error?

hanna-greaves: It should! Thanks for spotting that :)

package.json
```diff
@@ -3,9 +3,14 @@
   "private": true,
   "description": "A collection of components that make it easy to build interactive charts with D3",
   "scripts": {
-    "postinstall": "lerna bootstrap",
-    "publish": "lerna publish --conventional-commits --message \"chore: publish\" --yes",
-    "test": "lerna run test --stream --concurrency 1"
+    "postinstall": "./node_modules/.bin/lerna bootstrap",
+    "publish": "./node_modules/.bin/lerna publish --conventional-commits --message \"chore: publish\" --yes",
+    "bundle": "./node_modules/.bin/lerna run bundle --stream --concurrency 1",
+    "pretest": "./node_modules/.bin/lerna run bundle --stream --concurrency 1",
```
tommclaughlan: If this is always going to be the same as the `bundle` script, would it make sense to have `"pretest": "npm run bundle"` ?

package.json
```diff
@@ -3,9 +3,14 @@
   "private": true,
   "description": "A collection of components that make it easy to build interactive charts with D3",
   "scripts": {
-    "postinstall": "lerna bootstrap",
-    "publish": "lerna publish --conventional-commits --message \"chore: publish\" --yes",
-    "test": "lerna run test --stream --concurrency 1"
+    "postinstall": "./node_modules/.bin/lerna bootstrap",
```
chrisprice: Is the `./noode_modules/.bin` prefix required?

packages/d3fc-brush/package.json
```diff
@@ -16,14 +16,16 @@
     "url": "https://github.com/d3fc/d3fc"
   },
   "scripts": {
-    "bundle": "d3fc-scripts bundle",
-    "test": "d3fc-scripts test"
+    "bundle": "rollup -c ../../util/rollup.config.js --configName=brush",
```
chrisprice: What does `configName=brush` reference?

packages/d3fc-brush/package.json
```diff
@@ -16,14 +16,16 @@
...
+    "test": " jasmine JASMINE_CONFIG_PATH=../../util/jasmine.json",
```
chrisprice: Nitpic... Is the leading whitespace required?

chrisprice: Could we use [`--config`](https://www.npmjs.com/package/jasmine#configuration) to save a few characters?

packages/d3fc/package.json
```diff
@@ -41,9 +42,13 @@
     "@d3fc/d3fc-sample": "^4.0.8",
     "@d3fc/d3fc-series": "^4.0.18",
     "@d3fc/d3fc-shape": "^5.0.8",
-    "@d3fc/d3fc-technical-indicator": "^7.0.24"
+    "@d3fc/d3fc-technical-indicator": "^7.0.24",
+    "nollup": "^0.6.3"
```
chrisprice: Are we using this?

util/rollup.config.js
```diff
@@ -0,0 +1,42 @@
+import babel from 'rollup-plugin-babel';
```
chrisprice: There seems to be a fair amount of commonality between the rollup configs (`pacakges/d3fc/rollup.config.js`, `pacakges/d3fc/dev.rollup.config.js` & `utili/rollup.config.js`). I can imagine we'll have issues if the config drifts. Is there an easy way we can remove some of the duplication?

util/screenshot.js
```diff
@@ -0,0 +1,79 @@
...
+
+const server = http.createServer(function(req, res) {
+    const done = finalhandler(req, res);
+    serve(req, res, done);
+});
+
+server.listen(port, () =>
+  takeScreenshots()
+    .then(() => {
+        console.log('Screenshots complete 📷');
```
chrisprice: Excellent use of emoji.

chrisprice (changes requested):
Thanks @hanna-greaves, this looks great!

> The required folder has not been replaced, I would suggest any file requirements be enforced by travis.

I'm happy to see this dropped. It was mainly used to ensure we kept all the separate repositories up to date. Now that they're all together it's not required.

> Add util folder

* [x] Nitpic... can we call it `scripts`? I think it makes the distinction between library code and build code clearer.

As I've been reviewing this I've noticed a couple of things we also probably want to change but are out of scope for this PR: #1297 & #1296.

chrisprice:
One other thing that's just occurred to me: please could you confirm that you've tested this by comparing the build outputs prior and post this change?

package.json
hanna-greaves: This is me being a bit old fashioned, to avoid people having to -g install lerna I included the local path, but I can change this to npx now that it default ships with node ( and has done for a few years 😅 )

packages/d3fc-brush/package.json
hanna-greaves: This is something I was playing with while I was developing, will revert

util/rollup.config.js
hanna-greaves: I think I can provide a config flag here and merge the two files, will give that a shot today

util/screenshot.js
hanna-greaves: 📷 😉

hanna-greaves:
> One other thing that's just occurred to me: please could you confirm that you've tested this by comparing the build outputs prior and post this change?

I have looked at the build output in beyond compare for the repo as it is now and this branch and the only changes seem to be those caused by upgrading rollup :)

PR #1306, opened 2019-08-19
Add example rendering large number of web gl points

joelwilliamsmith:


tommclaughlan (changes requested):
If I check out this PR, the example doesn't work because `d3` is not an explicit dependency of `d3fc-webgl` and the example is attempting to pull in `packages/d3fc-webgl/node_modules/d3/build/d3.js`

PR #1308, opened 2019-08-20
Add package-lock files

### LINKED ISSUE #1307: package-lock.json files should be checked in to the repo

hanna-greaves:
Fixes #1307, see https://docs.npmjs.com/files/package-lock.json

Also fixes a few package vulnerabilities, primarily `watch-cli` which had hundreds and seemed to be something that was no longer used


packages/d3fc-webgl/package-lock.json
```diff
@@ -0,0 +1,3135 @@
+{
```
joelwilliamsmith, line 1: This file shouldn't be merged into master

joelwilliamsmith (changes requested):
Build has failed.

Additionally it was the presence of package-lock files that caused the Travis problems before (for one, it defaulted to running an `npm ci` rather than an `npm install` although I don't think that was the underlying issue). We should try and make sure this isn't going to do the same - saying that I don't really know how we can do that without merging first!

PR #1310, opened 2019-08-22
Add labeled stars example with SVG overlay

joelwilliamsmith:
A zoomable/pannable example showing use of an SVG overlay to achieve interactivity.


hanna-greaves (approved):
very cool!

PR #1312, opened 2019-08-23
WIP: Line series for WebGL

tommclaughlan:
This adds a line series alongside the existing point series.

It currently supports drawing lines, with breaks in the series, and matches the linear curve type in SVG and canvas.
It doesn't support curve types other than linear (we're not even checking, just linear by default).

It also could do with a bit of cleaning around the shader code doing the line joins at each vertex.

### DIFF
--- a/packages/d3fc-series/examples/line.js
+++ b/packages/d3fc-series/examples/line.js
@@ -38,3 +38,18 @@ var canvasLine = fc.seriesCanvasLine()
     .crossValue(function(_, i) { return i; })
     .mainValue(function(d) { return d; });
 canvasLine(data);
+
+var canvasgl = d3.select('#line-webgl').node();
+canvasgl.width = width;
+canvasgl.height = height;
+var gl = canvasgl.getContext('webgl');
+
+var webglLine = fc.seriesWebglLine()
+    .xScale(xScale)
+    .yScale(yScale)
+    .defined((_, i) => i % 20 !== 0)
+    .context(gl)
+    .crossValue(function(_, i) { return i; })
+    .mainValue(function(d) { return d; });
+
+webglLine(data);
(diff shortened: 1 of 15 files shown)


packages/d3fc-series/examples/line.js
```diff
@@ -38,3 +38,18 @@ var canvasLine = fc.seriesCanvasLine()
     .crossValue(function(_, i) { return i; })
     .mainValue(function(d) { return d; });
 canvasLine(data);
+
+var canvasgl = d3.select('#line-webgl').node();
```
hanna-greaves, line 42: these vars can all be const

packages/d3fc-webgl/src/series/glLine.js
```diff
@@ -0,0 +1,180 @@
...
+
+    const xValueAttrib = 'aXValue';
+    const yValueAttrib = 'aYValue';
+    const xNextValueAttrib = 'aNextXValue';
+    const yNextValueAttrib = 'aNextYValue';
+    const xPrevValueAttrib = 'aPrevXValue';
+    const yPrevValueAttrib = 'aPrevYValue';
+    const cornerValueAttrib = 'aCorner';
+    const widthUniform = 'uWidth';
+    const screenUniform = 'uScreen';
```
hanna-greaves, line 24: Would suggest you move these into an external file and export them so that they can be used in places like line.js:30

packages/d3fc-webgl/src/series/glLine.js
```diff
@@ -0,0 +1,180 @@
...
+
+        decorate(program);
+
+        segments.forEach(segment => {
+            // we're sending every vertex to the shader four times, hence * 4
+            program(segment.numElements * 4, segment.start * 4, segment.bufferSize * 4);
+        });
+    };
+
+    draw.xValues = (...args) => {
```
hanna-greaves, line 87: This line function is understandably quite complicated, do you have a link to any documentation that you followed when you were implementing it that you could add in as a comment to help us to understand it? :)

hanna-greaves (changes requested):
This looks good to me, and I think we should merge it sooner rather than later, my main concern at the moment is that there are quite a few magic strings that should be moved out to consts, but other than that I think we are okay to move forwards

PR #1314, opened 2019-08-30
Update Babel 🐟 and Lerna 🐉

### LINKED ISSUE #1307: package-lock.json files should be checked in to the repo

hanna-greaves:
We were a few years behind, so I've updated us to Babel 7 and Lerna 3

_notes_
* We have a bunch of copies of .babelrc, we can replace these with 1 as soon as we migrate from Jasmine to Jest ( Jasmine expects a .babelrc in the package directory )

* Also converts from using Uglify.JS to @babel/minify. The minified build size is down ~20% (88kb -> 76kb) :)

* Where possible vulnerable packages have been replaced (1 low priority vulnerability remains in `d3fc-site`

* `package-lock.json` files have been generated, resolves #1307


.travis.yml
```diff
@@ -9,6 +9,7 @@ node_js:
   - '10'
 script:
   - npm run bundle
+  - npm run bundle-min
```
joelwilliamsmith, line 12: Do we need to do both?

packages/d3fc-chart/src/cartesianBase.js
```diff
@@ -20,7 +20,7 @@ export default (setPlotArea, defaultPlotArea) =>
                 container.enter()
                     .select('.x-label')
                     .style('height', '1em')
-                    .style('line-height', '1em');
+                    .style('line-height', '4em');
```
joelwilliamsmith: What is this change for?

packages/d3fc-sample/package.json
```diff
@@ -1,40 +0,0 @@
-{
```
joelwilliamsmith: Why deleted?

joelwilliamsmith (changes requested):
Fails to build locally (caused by package-locks again 😡 )

packages/d3fc-chart/src/cartesianBase.js
hanna-greaves: whoops, that was from some feverish webgl testing... undoing.

packages/d3fc-sample/package.json
hanna-greaves: should not have been deleted..

.travis.yml
hanna-greaves: Yep, afaik we deploy both sets of files. An improvement would be to build once and minify the built files rather than doing this, but it would require a bit of a rollup config change that I didn't want to make for this PR

hanna-greaves:
I am undoing the `package-lock.json` changes, I believe that lerna is fighting with npm ci. Someone should follow up with this by checking out [this thread](https://github.com/lerna/lerna/issues/2015) which details the issue

PR #1315, opened 2019-09-02
feat: add Conventional Commit Support

hanna-greaves:
Enables Commitizen integration and a CC lint check which is run on Travis during PRs

* It would be nice to make this more of a warning rather than requirement on Travis but I do not currently have time to write the conditionals and PR commenting code for travis


package.json
```diff
@@ -9,7 +9,18 @@
...
     "lint": "eslint **/*.js -c .eslintrc.json && markdownlint **/*.md --ignore **/node_modules/**",
     "screenshots": "lerna run screenshots --stream --concurrency 1",
-    "start": "lerna run bundle && npm start --prefix ./packages/d3fc"
+    "start": "lerna run bundle && npm start --prefix ./packages/d3fc",
+    "commitlint": "commitlint --from HEAD~1 to HEAD --verbose"
+  },
+  "husky": {
+    "hooks": {
+      "prepare-commit-msg": [
+        "echo +----------------------------------------------+ &&",
```
ColinEberhardt: This is windows specific ...

```
husky > prepare-commit-msg (node v10.7.0)
/bin/sh: -c: line 0: syntax error near unexpected token `&&'
```

package.json
```diff
@@ -9,7 +9,18 @@
...
+        "echo.^|      This repo is Commitzen Friendly!        ^| &&",
+        "echo.^| Use git cz to generate a conventional commit ^| &&",
```
ColinEberhardt: I think `git cz` assumes that they have installed commitizen globally? 

Is it worth recommending `npx git-cz` instead?

package.json
```diff
@@ -9,7 +9,18 @@
     "test": "lerna run test --stream --concurrency 1",
     "lint": "eslint **/*.js -c .eslintrc.json && markdownlint **/*.md --ignore **/node_modules/**",
     "screenshots": "lerna run screenshots --stream --concurrency 1",
-    "start": "lerna run bundle && npm start --prefix ./packages/d3fc"
+    "start": "lerna run bundle && npm start --prefix ./packages/d3fc",
+    "commitlint": "commitlint --from HEAD~1 to HEAD --verbose"
+  },
+  "husky": {
+    "hooks": {
+      "prepare-commit-msg": [
```
ColinEberhardt: I'm not too familiar with git hooks - I guess this is intended to display a notice whenever someone commits? In this case, I think it should (1) indicate that they need to use conventional commits and (2) suggest that they use commitizen to generate them.

The way it is worded at the moment implies that they **must** use `git cz` to generate commits.

ColinEberhardt (changes requested):
Thanks for this, I've just tested it out and it looks really useful. I've not come across Commitizen before, a very handy tool indeed!

Just a few minor comments.

package.json
hanna-greaves: Yes, this executes just before git opens vim (or your usual terminal editor) for you to edit the default commit message. In *nix world there's a nice command that will automatically redirect git commit calls to npx git-cz but unfortunately this doesn't work cross platform so I opted to just give a notice :) 

will update the message to be more clear! :)

package.json
hanna-greaves: That's right, nice catch, thanks

package.json
hanna-greaves: I have moved this into a js file inside [`/scripts`](https://github.com/d3fc/d3fc/blob/06268041641fc22a59aa703744f2c6fb0d1e52e6/scripts/commit-notice.js) to avoid this issue and to declutter the `package.json` file

ColinEberhardt (approved):
Thanks for the updates @hanna-greaves - just tried this out and it works perfectly :-)

I've approved, can you please resolve the `package.json` conflicts then I'd be happy to merge. Thanks again.

ColinEberhardt:
Thanks for tidying up and merging 👍

PR #1322, opened 2019-10-08
Add support and example for WebGL squares for point series

OliverForeman:
- Added shaders and support for drawing squares in point series
- Updated shader names to be general to point instead of circle
- Added a symbol mapper to map d3 symbols to WebGL implementation
- Added a squares point series example with pannable/zoomable area


packages/d3fc-webgl/src/symbolMapper.js
```diff
@@ -0,0 +1,25 @@
...
+  let glSymbol = null;
+
+  switch (symbol) {
+    case d3Shape.symbolCircle:
+      glSymbol = circlePointShader();
+      break;
+    case d3Shape.symbolSquare:
+      glSymbol = squarePointShader();
+      break;
+    default:
```
chrisprice, line 11: Please can you make the default throw an unrecognised error rather than default

packages/d3fc-webgl/src/symbolMapper.js
```diff
@@ -0,0 +1,25 @@
+import d3Shape from 'd3-shape';
+import circlePointShader from './shaders/point/circle/baseShader';
+import squarePointShader from './shaders/point/square/baseShader';
+
+export default (symbol) => {
+  let outSymbol = d3Shape.symbolCircle;
```
chrisprice: I believe the `outSymbol` equivalent in the scale mapper is used in combination with the `glSymbol` equivalent. However, there's no way of combining the two symbols (happy to be proved wrong!) so I think we should remove `outSymbol` and just return `glSymbol` (without the object wrapper).

packages/d3fc-webgl/src/series/glPoint.js
```diff
@@ -87,7 +88,17 @@ export default () => {
...
+        type = functor(args[0]);
+        return draw;
+    };
+
     rebind(draw, program, 'context');
 
     return draw;
 };
+
+const functor = v => typeof v === 'function' ? v : () => v;
```
chrisprice: Please pull this out into its own file.

packages/d3fc-webgl/src/shaders/vertexShaderSnippets.js
```diff
@@ -24,3 +24,14 @@ export const point = {
            gl_PointSize = vSize + uEdgeSize + 1.0;
            gl_Position = vec4(aXValue, aYValue, 0, 1);`
 };
+
+export const square = {
+    header: `attribute float aXValue;
+        attribute float aYValue;
+        attribute float aSize;
+        uniform float uEdgeSize;
```
chrisprice: I appreciate you're just copying the pattern here but I think we call `edgeSize` `strokeWidth` in SVG/canvas. Is it the same thing and if so can we switch over?

packages/d3fc-webgl/examples/squares.js
```diff
@@ -0,0 +1,67 @@
...
+};
+
+const getChart = () => fc.chartCartesian(x, y)
+  .chartLabel(`${numPoints} Points`)
+  .decorate(decorate);
+
+const getWebglChart = () => getChart()
+  .webglPlotArea(getWebglSeries());
+
+const render = () => {
```
chrisprice: Again I think this is following the pattern but if you're using `chartCartesian` or in fact anything based on `d3fc-elements` you can call `requestRedraw` in the `zoom` handler.

packages/d3fc-webgl/src/shaders/point/square/baseShader.js
```diff
@@ -0,0 +1,19 @@
+import shaderBuilder, { vertexShaderBase, fragmentShaderBase } from '../../shaderBuilder';
```
chrisprice: How come these files are called `baseShader`? What are they the base of?

chrisprice (changes requested):

packages/d3fc-webgl/src/shaders/vertexShaderSnippets.js
OliverForeman: I think it is the same thing, I'll change it to `strokeWidth`

packages/d3fc-webgl/src/shaders/point/square/baseShader.js
OliverForeman: I think they are called `baseShader` as they are the base for each shape that can then be added to. I will update the file name to `shader`

packages/d3fc-webgl/src/symbolMapper.js
```diff
@@ -0,0 +1,20 @@
+import d3Shape from 'd3-shape';
+import circlePointShader from './shaders/point/circle/baseShader';
+import squarePointShader from './shaders/point/square/shader';
+
+export default (symbol) => {
```
chrisprice, line 5: This block will simplify down a fair bit if you just return the value in each case rather than `glSymbol = x; return;`

packages/d3fc-webgl/src/shaders/vertexShaderSnippets.js
```diff
@@ -14,13 +14,24 @@ export const multiColor = {
     body: `vColor = aColor;`
 };
 
-export const point = {
+export const circle = {
     header: `attribute float aXValue;
              attribute float aYValue;
              attribute float aSize;
-             uniform float uEdgeSize;
+             uniform float strokeWidth;
```
chrisprice: I think the `a` prefix is still required to be consistent i.e. `aStrokeWidth`

packages/d3fc-webgl/src/series/glPoint.js
```diff
@@ -87,6 +89,14 @@ export default () => {
         return draw;
     };
 
+    draw.type = (...args) => {
+        if (!args.length) {
+            return type;
+        }
+        type = functor(args[0]);
```
chrisprice: Sorry only just spotted that this is where `functor` was being used. Is there any benefit to using it here? Normally it would only be used if it were possible to e.g. change the value per data point.

OliverForeman: This was only here so that the type could be set per data point in the future, as that functionality isn't added in this PR I will change this

chrisprice (approved):

PR #1325, opened 2019-10-18
Add support for WebGL triangles for point series 

mattstobbs:
- Added shaders and support for drawing triangles in point series
- Added a dropdown to point series example with to allow shape to be changed


packages/d3fc-series/examples/point.js
```diff
@@ -48,9 +50,28 @@ var webglPoint = fc.seriesWebglPoint()
...
+    .type(d3.symbolCircle)
     .decorate(function(program) {
         fc.pointFill()(program);
         fc.pointStroke()(program);
         fc.pointAntiAlias()(program);
     });
 webglPoint(data);
+
+d3.select('#point-symbol').on('change', function() {
+  var newSymbol = d3.select(this).property('value');
```
chrisprice: Do the `d3[newSymbol]` lookup once?

packages/d3fc-series/examples/point.js
```diff
@@ -48,9 +50,28 @@ var webglPoint = fc.seriesWebglPoint()
...
+
+  svgPoint.type(d3[newSymbol]);
+  container.selectAll('g').remove();
```
chrisprice:
```
  container.selectAll('g').remove();
  container.append('g')
    .datum(data)
    .call(svgPoint);
```
Can be simplified to -
```
  container.select('g')
    .datum(data)
    .call(svgPoint);
```

Thanks to data join 👍

chrisprice (changes requested):
Just minor changes to the example please. Implementation looks good to me.

chrisprice (approved):

PR #1327, opened 2019-10-24
Webgl OHLC implementation with examples

OliverForeman:
- Added functionality for OHLC charts in Webgl
- Added Webgl OHLC to d3fc-series
- Added d3.scaleTime to Webgl scaleMapper
- Added examples of Webgl OHLC


chrisprice:
@mattstobbs please can you take a first look over this?

packages/d3fc-series/src/webgl/ohlc.js
```diff
@@ -0,0 +1,74 @@
...
+        const lineWidths = new Float32Array(filteredData.length);
+
+        filteredData.forEach((d, i) => {
+            xValues[i] = xScale.scale(accessor.xValues(d, i));
+            open[i] = yScale.scale(accessor.open(d, i));
+            high[i] = yScale.scale(accessor.high(d, i));
+            low[i] = yScale.scale(accessor.low(d, i));
+            close[i] = yScale.scale(accessor.close(d, i));
+            bandwidths[i] = accessor.bandwidth(d, i);
+            lineWidths[i] = lineWidth(d, i);
```
mattstobbs: Looks like this eventually gets loaded in as an attribute. Is there a reason it's not done as a uniform? Then you wouldn't have to repeat all the values for each vertex

packages/d3fc-webgl/src/series/glOhlc.js
```diff
@@ -0,0 +1,245 @@
...
+    program.vertexShader()
+      .appendBody(`
+        gl_Position.x += ((aXLineWidth / 2.0) + (aBandwidth / 2.0)) / uScreen.x;
+        gl_Position.y += (aYLineWidth / 2.0) / uScreen.y;
+      `);
+
+    decorate(program);
+
+    const buffers = buildBuffers(xValues, open, high, low, close, bandwidth, lineWidth, numElements);
+    bindBuffers(program, buffers);
```
mattstobbs: This is what was originally being done with bar but we've changed it to be more consistent with the rest of the code by binding to the buffers when the variable is initially set. I'm assuming we'd want to do that here too?

packages/d3fc-series/src/webgl/ohlc.js
OliverForeman: This was so that each data point could have a different line width, although looking now I don't think this is supported for svg or canvas. I can change this to be a uniform, similar to how `glLine` implements `width`, if we don't want this behaviour just for the webgl implementation?

chrisprice: Yea please swap it out for a uniform. I can't think of a good reason for it.

mattstobbs:
LGTM 👍

packages/d3fc-webgl/src/series/glOhlc.js
```diff
@@ -136,24 +139,42 @@ export default () => {
...
+
+      yDirArray = yDirArray.map((d, i) => indexPositions.includes(i % verticesPerElement) ?
+        yDirections[i % verticesPerElement] :
+        d
+      );
+      program.buffers().attribute(yDirectionAttrib, attributeBuilder(yDirArray).components(1));
+    }
+  };
 
+  const setColors = (numElements) => {
```
chrisprice, line 177: Could we do this in the shader? Maybe pass the colours in as uniforms?

OliverForeman: Each vertex requires some indication as to whether it should be green or red as it doesn't have enough information to calculate it itself. As we discussed this has now been changed to pass in an indicator of -1 (red) or 1 (green) to each vertex.

chrisprice (approved):

PR #1330, opened 2019-11-04
Line refactor to remove segments functionality

OliverForeman:
- Removed segments functionality from line
- Added `defined` as an attribute for drawing line
- Line is now drawn in a single pass


packages/d3fc-webgl/src/series/glLine.js
```diff
@@ -140,6 +131,39 @@ export default () => {
         return draw;
     };
 
+    draw.defined = (...args) => {
+        const builder = program.buffers().attribute(definedAttrib);
+
+        const definedArray = new Float32Array(args[0].length * 4);
+        
+        const length = definedArray.length;
```
chrisprice: Without knowing exactly what this is up to, this seems very confusing. I wonder if it think it might be clearer if it was done as a loop of the `args[0]` array (please give this a name!) instead?

OliverForeman: I've refactored this and used a loop of the args[0] array, hopefully it should be a bit clearer now. Let me know if it is still confusing

packages/d3fc-webgl/src/series/glLine.js
```diff
@@ -140,6 +131,39 @@ export default () => {
...
 
+    draw.defined = (...args) => {
+        const builder = program.buffers().attribute(definedAttrib);
+
+        const definedArray = new Float32Array(args[0].length * 4);
+        const values = args[0];
+
+        for (let i = 0; i < values.length; i += 1) {
+            const val = values[i];
+            const lastVal = i === 0 ? val : values[i - 1];
```
mattstobbs: I think it would be more consistent with the rest of the code if this was called `prevVal`

packages/d3fc-webgl/src/shaders/line/baseShader.js
```diff
@@ -11,11 +11,21 @@ export default () => {
...
+        .appendBody(`vDefined = aDefined;`)
         .appendBody(`vec4 curr = vec4(aXValue, aYValue, 0, 1);`)
         .appendBody(`gl_Position = curr;`)
         .appendBody(`vec4 next = vec4(aNextXValue, aNextYValue, 0, 0);`)
         .appendBody(`vec4 prev = vec4(aPrevXValue, aPrevYValue, 0, 0);`);
 
+    fragmentShader
+        .appendHeader(`varying float vDefined;`)
+        .appendBody(`if (vDefined < 0.5) {
+            discard;
```
chrisprice, line 25: I think discard is a control flow statement in and of itself, don't think the return is required.

packages/d3fc-webgl/src/series/glLine.js
```diff
@@ -140,6 +131,32 @@ export default () => {
         return draw;
     };
 
+    draw.defined = (...args) => {
+        const builder = program.buffers().attribute(definedAttrib);
+
+        const definedArray = new Float32Array(args[0].length * 4);
+        const values = args[0];
+
+        for (let i = 0; i < values.length; i += 1) {
```
OliverForeman, line 140: I've noticed this could also be done using `forEach`, is there a preference one way or another for using a `for` loop here or `values.forEach`?

_For comparison:_
```javascript
values.forEach((value, i) => {
  const previousValue = i === 0 ? value : values[i - 1];
  const nextValue = i === (values.length - 1) ? value : values[i + 1];
  const bufferIndex = i * 4;

  definedArray[bufferIndex] = value ? previousValue : value;
  definedArray[bufferIndex + 1] = value ? previousValue : value;
  definedArray[bufferIndex + 2] = value ? nextValue : value;
  definedArray[bufferIndex + 3] = value ? nextValue : value;
});
```

chrisprice (approved):

PR #1333, opened 2019-11-06
Refactored lineWidth for WebGL OHLC charts

OliverForeman:
Refactored the way `lineWidth` works for OHLC charts to be consistent with the code base


packages/d3fc-series/src/webgl/ohlc.js
```diff
@@ -55,6 +57,14 @@ export default () => {
         };
     }
 
+    ohlc.lineWidth = (...args) => {
```
chrisprice: Could this be rebound?

chrisprice (changes requested):
Not a fan of the naming `ohlcSeries.width()` (especially when we've got `ohlcSeries.bandwidth()`) but we can pick that up in the upcoming mop up.

chrisprice (approved):

PR #1338, opened 2019-11-20
Webgl Area Charts

OliverForeman:
- Added functionality for Area charts with WebGl
- Updated d3fc-series area example to include chart with WebGl


packages/d3fc-series/src/webgl/area.js
```diff
@@ -0,0 +1,50 @@
+import xyBase from '../xyBase';
+import { glArea, scaleMapper } from '@d3fc/d3fc-webgl';
+import { rebindAll, exclude, rebind } from '@d3fc/d3fc-rebind';
+
+export default () => {
+    const base = xyBase();
+
+    let draw = glArea();
```
chrisprice, line 8: `const`

packages/d3fc-series/src/webgl/area.js
```diff
@@ -0,0 +1,50 @@
...
+            .y0Values(y0Values)
+            .defined(defined)
+            .xScale(xScale.glScale)
+            .yScale(yScale.glScale)
+            .decorate((program) => base.decorate()(program, data, 0));
+
+        draw(data.length);
+    };
+
+    const getAccessors = () => ({
```
chrisprice: Can we get rid of this and just inline the calls into lines 22-25?

packages/d3fc-webgl/src/series/glArea.js
```diff
@@ -0,0 +1,163 @@
...
+    program.vertexShader(shaderBuilder.vertex())
+      .fragmentShader(shaderBuilder.fragment())
+      .mode(drawModes.TRIANGLES);
+
+    xScale.coordinate(0);
+    xScale(program);
+    yScale.coordinate(1);
+    yScale(program);
+
+    initCornerArray(numElements);
```
chrisprice: I thought we did this in `xValues` elsewhere?

chrisprice (changes requested):

packages/d3fc-series/src/webgl/area.js
OliverForeman: The other files in `src/webgl` all use `let` here, apart from `bar.js`. Shall I leave this for now to be consistent with the majority of the code base and it can be looked at in the mop up?

packages/d3fc-webgl/src/series/glArea.js
OliverForeman: I had this separate while adding the functionality and forgot to move it

chrisprice (approved):

PR #1341, opened 2019-11-25
feat: add support and example for WebGL cross for point series

mattstobbs:


packages/d3fc-webgl/src/shaders/fragmentShaderSnippets.js
```diff
@@ -22,7 +22,18 @@ export const triangle = {
     if (distance > 1.0) {
         discard;
     }`
-}
+};
+
+export const cross = {
+    header: `varying float vSize;`,
+    body: `vec2 pointCoordTransform = 2.0 * gl_PointCoord - 1.0;
+    float innerCornerDistance = min(abs(pointCoordTransform.x), abs(pointCoordTransform.y)) + 0.56;
```
chrisprice: `0.56` 🤔

chrisprice (changes requested):

packages/d3fc-webgl/src/shaders/fragmentShaderSnippets.js
mattstobbs: It should hopefully be more clear now

chrisprice (approved):

PR #1344, opened 2019-11-25
Move uScreen into program

mattstobbs:


packages/d3fc-webgl/src/program/programBuilder.js
```diff
@@ -9,6 +10,11 @@ export default () => {
     let mode = drawModes.TRIANGLES;
     let buffers = bufferBuilder();
     const build = (numElements, start = 0, bufferSize = numElements) => {
+        buffers.uniform('uScreen', uniformBuilder([
```
chrisprice: I think I would expect to see this down next to the `buffers` invocation?

chrisprice (approved):

PR #1346, opened 2019-11-26
Reinstate eslint - part 1

chrisprice:
It was pointed out that the eslint configuration wasn't being applied universally nor was it very rigorous. This PR starts the ball rolling to fix this. It includes -

* Establishing a set of rules we'd like to follow in the future - `prettier`/`semistandard`*.
* Applying that set of rules to our `/scripts` folder as a starting point.
* Removing all of the floating `.eslint*` files/comments and consolidating everything into the top-level `.eslintrc.json` file.
* Miscellaneous ~no~low-impact changes to bring the code up to our original eslint configuration.

I don't feel strongly about my choice of `prettier`/`semistandard`. I just want consistency and they both seem like popular ways of achieving that.

Currently 2 packages contain `site` folders which are excluded from linting. These will be moved into `examples` folders and brought into line with our standard configuration in my next PR.

I would recommend reviewing commit by commit.


ColinEberhardt (approved):
A good step in the right direction - waffling feedback on slack!

PR #1348, opened 2019-11-28
WebGL matrix calculations for area

OliverForeman:
Replaces the current calculations for the intercept point with matrix calculations


chrisprice:
Blocked on #1402 to allow us to test the performance impact of these changes.

OliverForeman:
After performance testing this change I've found there is no benefit to performance from using matrix calculations. As the current calculations are easier to understand this change will not be made.

PR #1351, opened 2019-11-28
Move shader code into shader snippet files

mattstobbs:


chrisprice:
Pre/post scale naming change

chrisprice (approved):

PR #1360, opened 2020-01-02
Initial WebGL Implementation

chrisprice:
Adds `webgl*Series` to `d3fc-series` and a new package `d3fc-webgl` containing supporting components. This new package is not yet feature-complete and the API is expected to change, therefore users should only consume these new components via `d3fc-series`.

### DIFF
--- a/packages/d3fc-series/README.md
+++ b/packages/d3fc-series/README.md
@@ -1,6 +1,6 @@
 # d3fc-series
 
-A collection of components for rendering data series to SVG and canvas, including line, bar, OHLC, candlestick and more.
+A collection of components for rendering data series to SVG, canvas and WebGL, including line, bar, OHLC, candlestick and more.
 
 <table>
 <tr>
@@ -43,6 +43,7 @@ npm install @d3fc/d3fc-series
 * [General API](#general-api)
  * [SVG Rendering](#svg-rendering)
  * [Canvas Rendering](#canvas-rendering)
+ * [WebGL Rendering](#webgl-rendering)
  * [Decorate Pattern](#decorate-pattern)
  * [Orientation](#orientation)
  * [Multi Series](#multi-series)
@@ -63,7 +64,7 @@ npm install @d3fc/d3fc-series
 * [Grouped](#grouped)
 * [Stacked](#stacked)
 
-This packages contains a number of D3 components that render various standard series types. They all share a common API, with the typical configuration requiring x and y scales together with a number of value accessors. There are SVG and Canvas versions of each series type, sharing the same configuration properties.
+This package contains a number of D3 components that render various standard series types. They all share a common API, with the typical configuration requiring x and y scales together with a number of value accessors. There are SVG, Canvas and WebGL versions of each series type, sharing the same configuration properties.
 
 ### General API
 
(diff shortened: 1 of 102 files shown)


packages/d3fc-series/README.md
```diff
@@ -120,6 +121,34 @@ line(data);
 Because D3 data-joins and data-binding only work on HTML / SVG, the canvas components are invoked directly with the supplied data. This causes the component to render itself to the canvas.
 
 
+#### WebGL rendering
+
+***WebGL rendering is not feature complete. Please check the API documentation against the methods to see what is currently supported.***
```
ColinEberhardt: It might be worth making this sound a bit more optimistic, there are certainly enough features supported to make a whole host of typical visualisations entirely possible.

packages/d3fc-webgl/src/series/glCandlestick.js
```diff
@@ -0,0 +1,252 @@
...
+
+        decorate(program);
+
+        program(numElements * verticesPerElement);
+    };
+
+    draw.xValues = (...args) => {
+        const builder = program.buffers().attribute(xValueAttrib);
+        let xArray = new Float32Array(args[0].length * verticesPerElement);
+        xArray = xArray.map(
```
ColinEberhardt, line 54: Just having a quick poke around from a performance perspective - this method of mapping from the `args[0]` array to `xArray` is a bit slow (also I think it also replaces the array constructed in the line above). By my (crude) performance measurements, the following is approx x10 faster:

~~~
for (let i=0;i < xArray.length;i++) {
  xArray[i] = args[0][Math.floor(i / verticesPerElement)];
}
~~~

although YMMV.

ColinEberhardt: A similar map-based pattern is used elsewhere, so probably worth investigating further.

packages/d3fc-webgl/examples/webgl-overlay.html
```diff
@@ -0,0 +1,13 @@
+<html>
```
ColinEberhardt, line 1: When running this example, if the browser window (and chart) is re-sized the WebGL and SVG  layers become mis-aligned. This might point to an underlying issue in d3fc-chart?

ColinEberhardt (commented):
The [Developing section](https://github.com/d3fc/d3fc#developing) of the README needs updating to add a bundle set between install & test:

~~~
npm install
npm run bundle
npm test
~~~

packages/d3fc-webgl/src/series/glCandlestick.js
chrisprice: Good spot. This should all hopefully be addressed by the changes I'm currently making to introduce 'projectedAttributeBuilder`.

chrisprice: As this is an internal implementation detail of `d3fc-webgl`, are you ok with this change being made subsequently?

ColinEberhardt (approved):

ColinEberhardt:
> As this is an internal implementation detail of d3fc-webgl, are you ok with this change being made subsequently?

Yes, absolutely. I've approved this PR.

packages/d3fc-webgl/examples/webgl-overlay.html
chrisprice: Excellent spot, that wasn't fun to track down! 

I've added a new attribute to `d3fc-canvas` called `set-webgl-viewport` which propagates the surface dimensions down to the WebGL context. This is automatically set on the `webgl-plot-area` by `cartesianChart`. 

I didn't want to make this automatic in case there's a potential performance impact to requesting a WebGL context if you're not using it. As `webgl-plot-area` is only rendered if you supply a value to `webglPlotArea`, this shouldn't be a problem.

ColinEberhardt: great - no problem :-)

Issue #1375, opened 2020-01-10
baseAttributeBuilder.type is not respected


chrisprice:
`projectedAttributeBuilder` does not consider `baseAttribute.type()` when constructing the `projectedData`.

OliverForeman:
Attribute types are restricted to `float`, `vectors`, and `matrices`, do we have any use cases for the types other than `float`

chrisprice:
This ticket should also cover optimising our usage of array types (to reduce their transfer time to the GPU). Anything that is expressed in screen pixels (e.g. line widths, point sizes, etc.) or is a flag (e.g. defined) should be changed to a smaller type (e.g. uint8/16).

jamesmporter:
After experimenting with this, it seems that a `Float32Array` is necessary for `ARRAY_BUFFER` and `Uint16Array` for the `ELEMENT_ARRAY_BUFFER`. However, further investigation may reveal a way to solve this problem.

Issue #1374, opened 2020-01-10
Optimise Float32Array allocation


chrisprice:
In `projectedAttributeBuilder`, the `project` function always allocated a new `Float32Array` even if the old array could be re-used (i.e. the new array is less than or equal in length to the previous one).

Issue #1373, opened 2020-01-10
Comparisons on scaled values


chrisprice:
The following series perform comparisons on scaled rather than raw values -

* `glCandlestick` - in the computation of `aColorIndicator` & `aYDirection`
* `glOhlc` - in the computation of `aColorIndicator`

chrisprice:
`glCandlestick`-`aYDirection` should be fixed by #1371.

I can't think of an example where you would ever want the colour not to match the relative on-screen magnitudes of the values.

OliverForeman:
This is no longer an issue, both `candlestick` and `ohlc` now use the values passed directly to the shaders for their comparisons.

```
vColorIndicator = sign(aCloseValue - aOpenValue);
```

Issue #1372, opened 2020-01-10
Rebind properties in series/glXXX


chrisprice:
Once the specialised attribute builders (#1368, #1369, #1370) exist and the geometry shader work out of JS (#1371) the `projectedAttributeBuilder.data(...)` calls should be rebound.

jamesmporter:
Earmarked for Adil.

PR #1376, opened 2020-01-10
feat: add projectedAttributeBuilder

chrisprice:
Replaces `attributeBuilder` with `projectedAttributeBuilder`. 

Previously `attributeBuilder` was expected to be a dumb conduit for pushing `xxxArray`s to WebGL. It was assumed that the source data could be directly loaded into GL attribute buffers with little manipulation. However, owing to a lack of geometry shaders in WebGL this has not proved possible and we are now transforming the data twice (once in `d3fc-series` and once in `d3fc-webgl`) resulting in lots of boilerplate in the `d3fc-series` implementations.

This PR adds `projectedAttributeBuilder` which is a first step towards making the process smarter and reducing the duplication (of boilerplate, memory & iteration). 

The following issues were identified as part of this work and are expected to be addressed at a later stage -

* #1377 Optimise two-stage generation of attributes
* #1375 - baseAttributeBuilder.type is not respected
* #1374 - Optimise Float32Array allocation
* #1373 - Comparisons on scaled values
* #1372 - Rebind properties in series/glXXX
* #1371 - Move geometry shader work into shaders
* #1370 - Add attributeBuilder optimisation for next/prev use case
* #1369 - Add vertexConstantAttributeBuilder
* #1368 - Add elementConstantAttributeBuilder
* #1367 - projectedAttributeBuilder does not dirty check any base properties
* #1365 - seriesWebglBar bandwidth uses different units


OliverForeman (approved):
LGTM

PR #1386, opened 2020-01-17
feat: add elementConstantAttributeBuilder

josted:
Putting up for early visibility as I'm not sure it's exactly on the money yet, nor am I certain of all of the places it can be used, so feel free to tell me if I'm barking up the wrong tree, or even in the wrong forest entirely.

Tested it with glArea to fill` x/yPreviousValueAttributes`, and can also pass an array to use `vertexAttrib[1234]fv`.

For example, replacing the colorIndicatorAttribute on candlestick with this:
`const colorIndicatorAttribute = elementConstantAttributeBuilder().value(
        [-1]
    );`
will turn every element red. Entering [1] here instead turns them all green.

Linked to https://github.com/d3fc/d3fc/issues/1368

### DIFF
--- a/packages/d3fc-webgl/src/series/glArea.js
+++ b/packages/d3fc-webgl/src/series/glArea.js
@@ -1,4 +1,5 @@
 import projectedAttributeBuilder from '../buffers/projectedAttributeBuilder';
+import elementConstantAttributeBuilder from '../buffers/elementConstantAttributeBuilder';
 import glScaleBase from '../scale/glScaleBase';
 import programBuilder from '../program/programBuilder';
 import drawModes from '../program/drawModes';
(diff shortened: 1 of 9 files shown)


chrisprice:
Please can you rebase this on master?

packages/d3fc-webgl/src/series/glArea.js
```diff
@@ -10,22 +11,22 @@ export default () => {
     let yScale = glScaleBase();
     let decorate = () => {};
 
-    const xValueAttribute = projectedAttributeBuilder().value(
+    const xValueAttribute = elementConstantAttributeBuilder().value(
         (data, element) => data[Math.min(element + 1, data.length - 1)]
     );
-    const xPreviousValueAttribute = projectedAttributeBuilder().value(
+    const xPreviousValueAttribute = elementConstantAttributeBuilder().value(
```
chrisprice: I don't think we need to explicitly set `value` as it's the same as the default. Please could you check for this in the other changes?

chrisprice (changes requested):
Sorry final comment... maybe...

chrisprice (approved):

Issue #1371, opened 2020-01-10
Move geometry shader work into shaders


chrisprice:
For the best possible performance we attempt to minimise the amount of work done in JS. Ideally all series would use a geometry shader to convert each datum into N vertices as required to render each series element. However, as this is not supported in WebGL this needs to be done in JS.

Some series types have abused this to perform additional calculations in JS. These should be moved into the GL shader. The following list tracks which series have the correct implementation -

* [x] glArea.js
* [x] glBar.js
* [x] glBoxPlot.js
* [x] glCandlestick.js
* [x] glErrorBar.js
* [x] glLine.js
* [x] glOhlc.js
* [x] glPoint.js

PR #1391, opened 2020-01-20
refactor: move glOhlc calculations from JavaScript into web-gl

jamesmporter:
Every web-gl attribute, defined in glOhlc.js, now only varies with either element or vertex but not both.

Solves part of: #1371

### DIFF
--- a/packages/d3fc-webgl/src/series/glOhlc.js
+++ b/packages/d3fc-webgl/src/series/glOhlc.js
@@ -1,7 +1,7 @@
 import projectedAttributeBuilder from '../buffers/projectedAttributeBuilder';
 import glScaleBase from '../scale/glScaleBase';
 import programBuilder from '../program/programBuilder';
-import rectShader from '../shaders/rect/shader';
+import ohlcShader from '../shaders/ohlc/shader';
 import lineWidthShader from '../shaders/lineWidth';
 import drawModes from '../program/drawModes';
 import { rebind } from '@d3fc/d3fc-rebind';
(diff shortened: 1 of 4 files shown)


chrisprice:
Please can you rebase and include the changes from #1386.

packages/d3fc-webgl/src/series/glOhlc.js
```diff
@@ -15,48 +15,52 @@ export default () => {
...
+    /*
+     * x-y coordinate to locate the "corners" of the element.
+     * X: -1: LEFT, 0: MIDDLE, 1: RIGHT
+     * Y: -2: HIGH, -1: OPEN, 1: CLOSE, 2: LOW
+     * Z - Follows convention for X/Y (appropriate direction will be selected by the shader): -1: LEFT/TOP, 1: RIGHT/BOTTOM
+     */
+    const cornerAttribute = projectedAttributeBuilder()
+        .size(3)
+        .data([
+            // Triangle 1
```
chrisprice: Please remove the comments. I don't think they help.

chrisprice: I think they might be wrong anyway, should `triangle` be `rectangle`?

packages/d3fc-webgl/src/series/glOhlc.js
```diff
@@ -65,11 +69,12 @@ export default () => {
         program
             .buffers()
             .attribute('aXValue', xValueAttribute)
-            .attribute('aYValue', yValueAttribute)
-            .attribute('aXDirection', xDirectionAttribute)
-            .attribute('aYDirection', yDirectionAttribute)
+            .attribute('aHigh', highAttribute)
```
chrisprice, line 69: I'm not keen on the names but I can see these are copied from candlestick so that horse has bolted. I've raised #1396 to cover improving them.

chrisprice (changes requested):

packages/d3fc-webgl/src/series/glOhlc.js
jamesmporter: Yes, I think you're right.

packages/d3fc-webgl/src/series/glOhlc.js
jamesmporter: What would you have prefered them to be?

packages/d3fc-webgl/src/series/glOhlc.js
jamesmporter: Done.

chrisprice (approved):

PR #1395, opened 2020-01-21
refactor: convert the web-gl bar series into the standard format and move logic from JS into web-gl

### LINKED ISSUE #1365: seriesWebglBar bandwidth uses different units
The following compares the canvas and webgl implementations with `bandwidth(0.5)` -

image

jamesmporter:
Solves part of #1371 

In addition to standardising the web-gl bar by moving logic into GLSL, this fixes #1365.

The second commit is a minor refactor of the error bar in order to make the scaling match the convention of the bar. However, the ohlc and candlestick do not match this convention (there may be further work involved in this beyond a minor refactor so has not been addressed here).

### DIFF
--- a/packages/d3fc-series/examples/bar.js
+++ b/packages/d3fc-series/examples/bar.js
@@ -47,6 +47,5 @@ var webglBar = fc.seriesWebglBar()
     .yScale(yScale)
     .context(gl)
     .crossValue(function(_, i) { return i; })
-    .mainValue(function(d) { return d; })
-    .bandwidth(0.5);
+    .mainValue(function(d) { return d; });
 webglBar(data);
--- a/packages/d3fc-series/src/webgl/bar.js
+++ b/packages/d3fc-series/src/webgl/bar.js
@@ -19,9 +19,9 @@ export default () => {
         const widths = new Float32Array(filteredData.length);
         filteredData.forEach((d, i) => {
             xValues[i] = xScale.scale(base.crossValue()(d, i));
+            widths[i] = xScale.scale(base.bandwidth()(d, i));
             y0Values[i] = yScale.scale(base.baseValue()(d, i));
             yValues[i] = yScale.scale(base.mainValue()(d, i));
-            widths[i] = yScale.scale(base.bandwidth()(d, i));
         });
 
         draw.xValues(xValues)
--- a/packages/d3fc-webgl/src/series/glBar.js
+++ b/packages/d3fc-webgl/src/series/glBar.js
@@ -3,6 +3,7 @@ import drawModes from '../program/drawModes';
 import programBuilder from '../program/programBuilder';
 import barShader from '../shaders/bar/shader';
 import { rebind } from '@d3fc/d3fc-rebind';
+import glScaleBase from '../scale/glScaleBase';
 import elementConstantAttributeBuilder from '../buffers/elementConstantAttributeBuilder';
 
 //     βL            β            βR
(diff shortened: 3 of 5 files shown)


packages/d3fc-webgl/src/series/glBar.js
```diff
@@ -32,22 +33,29 @@ import elementConstantAttributeBuilder from '../buffers/elementConstantAttribute
 
 export default () => {
     const program = programBuilder().verticesPerElement(6);
-    let xScale = null;
```
jamesmporter, line 35: @chrisprice I'm unsure whether this change was neccesary. It matches the convention in the errorBar.

packages/d3fc-series/src/webgl/bar.js
```diff
@@ -7,7 +7,7 @@ export default () => {
 
     const draw = glBar();
 
-    const bar = (data) => {
+    const bar = data => {
```
OliverForeman: In the other WebGL series implementations we've got brackets around the data parameter, probably best that this stays consistent

packages/d3fc-series/src/webgl/bar.js
```diff
@@ -30,7 +30,7 @@ export default () => {
             .widths(widths)
             .xScale(xScale.glScale)
             .yScale(yScale.glScale)
-            .decorate((program) => {
+            .decorate(program => {
```
OliverForeman: Same as above, we have brackets here in the other implementations

OliverForeman (changes requested):
All looks good, just a couple of styling changes

packages/d3fc-series/src/webgl/bar.js
jamesmporter: Do we have a prettier config? This file was auto-formatted in my editor.

packages/d3fc-series/examples/bar.js
```diff
@@ -2,25 +2,32 @@ var width = 500;
 var height = 250;
 var container = d3.select('#bar-svg');
 
-var dataGenerator = fc.randomGeometricBrownianMotion()
-  .steps(50);
+var dataGenerator = fc.randomGeometricBrownianMotion().steps(50);
```
chrisprice: As it stands we have a different code style in these example files. Your formatter seems to have been a bit over-zealous. Any chance you can match the existing style with your changes and revert non-functional formatting changes?

packages/d3fc-series/src/webgl/bar.js
chrisprice: See my previous comment. The configuration is stored in the `eslintrc.json` file.

packages/d3fc-webgl/src/series/glBar.js
chrisprice: Yes it is, good spot. We try to provide sane defaults for all values.

chrisprice (changes requested):

packages/d3fc-series/src/webgl/bar.js
jamesmporter: Done.

packages/d3fc-series/examples/bar.js
jamesmporter: Done.

packages/d3fc-series/src/webgl/bar.js
jamesmporter: Done.

jamesmporter (commented):
I've addressed the formatting issues.

chrisprice (approved):

PR #1399, opened 2020-01-23
fix: move dirty check and baseAttributeBuilder call

### LINKED ISSUE #1367: projectedAttributeBuilder does not dirty check any base properties
The properties from `baseAttributeBuilder` are not considered as part of the dirty check in `projectedAttributeBuilder`.

### LINKED ISSUE #1398: Set vertex attribute metadata on every render
When using `seriesWebglMulti` the first render is correct but all subsequent renders fail to display anything. This happens because the attribute location metadata is not updated for every attribute on every render, if the data hasn't been updated then it's skipped (e.g. for `aCorner` the data isn't updated after the initial render).

The solution to this is to always call the `baseAttributeBuilder`, so that the correct location is set for each attribute, before checking if the data has updated (`validSize >= count`).

Example:
```javascript
if (validSize >= count) {
  return;
}
...
base(gl, program, name);
```
becomes
```javascript
base(gl, program, name);
if (validSize >= count) {
  return;
}
...
```

OliverForeman:
Moves the dirty check variable `validSize` to `baseAttributeBuilder` allowing base properties to be considered, this fixes https://github.com/d3fc/d3fc/issues/1367

Moves the call to `baseAttributeBuilder` to before the dirty check in `projectedAttributeBuilder` and `elementConstantAttributeBuilder`, this fixes https://github.com/d3fc/d3fc/issues/1398

### DIFF
--- a/packages/d3fc-webgl/src/buffers/baseAttributeBuilder.js
+++ b/packages/d3fc-webgl/src/buffers/baseAttributeBuilder.js
@@ -7,11 +7,14 @@ export default () => {
     let normalized = false;
     let stride = 0;
     let offset = 0;
+    let validSize = 0;
 
     const base = (gl, program, name) => {
-        if (buffer == null) {
+        if (buffer == null || !gl.isBuffer(buffer)) {
             buffer = gl.createBuffer();
+            validSize = 0;
         }
+
         gl.bindBuffer(gl.ARRAY_BUFFER, buffer);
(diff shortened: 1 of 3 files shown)


chrisprice:
Changes look good but I'm not sure it fully fixes #1367, what happens if offset changes?

chrisprice (approved):
My bad, didn't spot that changing `offset` reset `validSize`

Issue #1382, opened 2020-01-16
Enable unit testing of shaders


jamesmporter:
Look at a package (similar to the one below) to transpile @GLSL to JS in order to enable unit testing of shaders.

https://www.npmjs.com/package/glsl-transpiler

PR #1400, opened 2020-01-23
Unit test shaders

jamesmporter:
Addresses #1382

### DIFF
--- a/packages/d3fc-webgl/package.json
+++ b/packages/d3fc-webgl/package.json
@@ -22,5 +22,8 @@
   },
   "publishConfig": {
     "access": "public"
+  },
+  "devDependencies": {
+    "glsl-transpiler": "^1.8.5"
   }
 }
--- a/packages/d3fc-webgl/test/util/vertextShaderUtil.js
+++ b/packages/d3fc-webgl/test/util/vertextShaderUtil.js
@@ -1,0 +1,59 @@
+const Compiler = require('glsl-transpiler');
+
+export const getShaders = element => {
+    let _program;
+    try {
+        const bar = element.decorate(program => {
+            _program = program;
+            throw new Error('HACKY_SOLUTION');
+        });
+        bar();
+    } catch (e) {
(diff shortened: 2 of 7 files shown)


packages/d3fc-webgl/test/util/vertextShaderUtil.js
```diff
@@ -0,0 +1,52 @@
...
+export const getShaders = element => {
+    let _program;
+    try {
+        const bar = element.decorate(program => {
+            _program = program;
+            throw new Error();
+        });
+        bar();
+    } catch (e) {
+        // no-op
```
chrisprice: Might be worth throwing a known error message i.e. `new Error('HORRIBLE HACK')` and asserting that's the received error i.e. `catch (e) { if (e.message !== 'HORRIBLE HACK') { throw e; } ... }`

packages/d3fc-webgl/package.json
```diff
@@ -18,7 +18,8 @@
   },
   "dependencies": {
     "@d3fc/d3fc-rebind": "^5.0.8",
-    "d3": "^4.1.1"
+    "d3": "^4.1.1",
+    "glsl-transpiler": "^1.8.5"
```
chrisprice: As this is only required for testing, please can you make it a devDependency?

chrisprice (changes requested):
Very nice. I think this will make a big difference to the maintainability of these shaders. Just a few minor comments

packages/d3fc-webgl/package.json
jamesmporter: Sure.

packages/d3fc-webgl/test/util/vertextShaderUtil.js
jamesmporter: Good idea.

chrisprice (approved):

OliverForeman (approved):

PR #1401, opened 2020-01-24
feat: add equals function to d3fc-series webgl implementations

OliverForeman:
Adds an `equals` function to each of the d3fc-series WebGL implementations, the function takes the previous data and current data as arguments.

We use the `equals` function to skip the creation of `xValues`, `yValues`, etc. data arrays if the data hasn't changed. By default we return `false`, assuming that the data has always changed.

This implements point 1 of https://github.com/d3fc/d3fc/issues/1366

### DIFF
--- a/packages/d3fc-series/src/webgl/area.js
+++ b/packages/d3fc-series/src/webgl/area.js
@@ -1,4 +1,5 @@
 import xyBase from '../xyBase';
+import isIdentityScale from '../isIdentityScale';
 import { glArea, scaleMapper } from '@d3fc/d3fc-webgl';
 import { rebindAll, exclude, rebind } from '@d3fc/d3fc-rebind';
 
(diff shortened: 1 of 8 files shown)


packages/d3fc-series/src/webgl/area.js
```diff
@@ -7,33 +8,51 @@ export default () => {
 
     const draw = glArea();
 
+    let equals = (previousData, data) => false;
+    let previousData = '';
```
chrisprice: Would an empty array be more appropriate?

packages/d3fc-series/src/webgl/area.js
```diff
@@ -7,33 +8,51 @@ export default () => {
...
-            y0Values[i] = yScale.scale(base.baseValue()(d, i));
-            defined[i] = base.defined()(d, i);
-        });
-
-        draw.xValues(xValues)
-            .yValues(yValues)
-            .y0Values(y0Values)
-            .defined(defined)
-            .xScale(xScale.glScale)
+        const validXScale = isIdentityScale(xScale.scale);
```
chrisprice: I'd inline these into the conditional

chrisprice (changes requested):

chrisprice (approved):

PR #1402, opened 2020-01-24
test: new examples w/ perf. & regression tests

chrisprice:
Ultimately I'd like to move all of the following into this structure -
* packages/*/examples
* packages/*/site
* packages/d3fc-site/examples
* bl.ocks.org referenced by packages/d3fc-site/examples/index.md


chrisprice:
Spotted a few problems with this branch which I've just pushed fixes for -
* `d3fc-element` contained a bug when using `use-device-pixel-ratio`. I've fixed the bug as a new commit.
* `emptyRedrawQueue` was setup to work with a monkey-patched version of `d3fc-element`. I've updated it to use the current version of `d3fc-element`.
* The rendered bar series looked odd because of a fix to bandwidth.

examples/README.md
```diff
@@ -0,0 +1,32 @@
...
+## Regression testing
+
+First the performance tests need baselining for the machine they are being run
+on. Check out the code prior to the change under test and run -
+
+```bash
+npm run examples -- --updateSnapshot
+```
+
+At this stage the `*/__tests__/__snapshots__` files will have being updated
```
jamesmporter: **Typo:**

```suggestion
At this stage the `*/__tests__/__snapshots__` files will have been updated
```

examples/__helpers__/emptyRedrawQueue.js
```diff
@@ -0,0 +1,6 @@
+module.exports.emptyRedrawQueue = async () => {
+    await page.waitFor(
+        () =>
+            document['__d3fc-elements__'] == null
```
jamesmporter: Should this use `packages\d3fc-element\src\data.js\key`, or at least move that functionality to this examples folder?

chrisprice: I can't access `key` as it is not currently exported from that module nor from the package (`packages/d3fc-element/index.js`). I can't see a good reason other than testing to access the value so I'd rather not change that.

As this is test code, if the value ever changes then the tests will fail so I'm happy to have the duplication

jamesmporter: Sounds good.

jamesmporter (approved):

examples/README.md
```diff
@@ -0,0 +1,32 @@
+# Examples
+
+This directory contains a set of examples utilising various d3fc features.
+
+## Viewing the examples
+
+Just load up one of the `index.html` files within an example directory.
```
ColinEberhardt: I misread this as "from within an example directory" - might be worth clarifying ...

"Just load up one of the example index.html files from an example directory, served from the project root"

chrisprice: There's currently no need to run up a web server, you can literally just open the file. I was expecting we'd add some more detailed instructions or a package script if/when that is required.

chrisprice: I've reworded it to hopefully make that clear

chrisprice:
@OliverForeman I've pushed a jest config change to run with `maxConcurrency = 1` to fix the issue you observed when you added a second example.

examples/README.md
ColinEberhardt: > There's currently no need to run up a web server, you can literally just open the file
Oh - that's cool, you should definitely make that clear. I assumed a dev server was required!

ColinEberhardt (approved):
Tried this out locally and it worked just fine 👍 

I did see a false positive just once, where the timing had deviated from the baseline beyond the default tolerance. However, I'm assuming these tests are intended to be a guide, and that you might occasionally have to make repeat measurements.

PR #1407, opened 2020-01-24
Use drawArraysInstancedANGLE()

josted:
Should no longer be blocked - line and point now working, though perhaps not in the most graceful way for the latter.

Uses ANGLE instancing extension to draw vertex and element constants with fewer draw calls.
See here: https://github.com/d3fc/d3fc/issues/1369.

### DIFF
--- a/packages/d3fc-webgl/src/buffers/baseAttributeBuilder.js
+++ b/packages/d3fc-webgl/src/buffers/baseAttributeBuilder.js
@@ -8,8 +8,9 @@ export default () => {
     let stride = 0;
     let offset = 0;
     let validSize = 0;
+    let divisor = 0;
 
-    const base = (gl, program, name) => {
+    const base = (gl, program, name, verticesPerElement) => {
         if (buffer == null || !gl.isBuffer(buffer)) {
             buffer = gl.createBuffer();
             validSize = 0;
@@ -19,13 +20,17 @@ export default () => {
         const location = gl.getAttribLocation(program, name);
         gl.vertexAttribPointer(
             location,
-            base.size(),
-            base.type(),
-            base.normalized(),
-            base.stride(),
-            base.offset()
+            size,
+            type,
+            normalized,
+            stride,
+            offset
         );
         gl.enableVertexAttribArray(location);
+        if (verticesPerElement !== 1) {
+            const ext = gl.getExtension('ANGLE_instanced_arrays');
+            ext.vertexAttribDivisorANGLE(location, divisor);
+        }
     };
 
     base.validSize = (...args) => {
@@ -87,6 +92,14 @@ export default () => {
         }
         offset = args[0];
         validSize = 0;
+        return base;
+    };
+
+    base.divisor = (...args) => {
+        if (!args.length) {
+            return divisor;
+        }
+        divisor = args[0];
         return base;
     };
 
--- a/packages/d3fc-webgl/src/buffers/vertexConstantAttributeBuilder.js
+++ b/packages/d3fc-webgl/src/buffers/vertexConstantAttributeBuilder.js
@@ -5,15 +5,12 @@ export default () => {
     const base = baseAttributeBuilder();
 
     let data = null;
-    let validSize = 0;
 
-    const project = (elementCount, verticesPerElement) => {
+    const project = verticesPerElement => {
         const components = base.size();
         const offset = base.offset();
         const componentsPerElement = verticesPerElement * components;
-        const projectedData = new Float32Array(
-            offset + elementCount * componentsPerElement
-        );
+        const projectedData = new Float32Array(offset + componentsPerElement);
         if (data.length !== verticesPerElement) {
             throw new Error(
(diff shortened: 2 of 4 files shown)


packages/d3fc-webgl/src/buffers/baseAttributeBuilder.js
```diff
@@ -19,13 +20,17 @@ export default () => {
...
-            base.stride(),
-            base.offset()
+            size,
+            type,
+            normalized,
+            stride,
+            offset
         );
         gl.enableVertexAttribArray(location);
+        if (verticesPerElement !== 1) {
```
josted, line 30: This and line 33 in programBuilder fixed the issue with Point, but it feels a bit hacky - I'll speak to you later because I've forgotten exactly what but think you had something else in mind @chrisprice?

packages/d3fc-webgl/src/buffers/vertexConstantAttributeBuilder.js
```diff
@@ -7,12 +7,12 @@ export default () => {
...
     let validSize = 0;
 
-    const project = (elementCount, verticesPerElement) => {
+    const project = verticesPerElement => {
         const components = base.size();
         const offset = base.offset();
         const componentsPerElement = verticesPerElement * components;
         const projectedData = new Float32Array(
-            offset + elementCount * componentsPerElement
+            offset + 1 * componentsPerElement
```
chrisprice: Do we need the `1 *`?

packages/d3fc-webgl/src/buffers/baseAttributeBuilder.js
chrisprice: Please can you raise an issue to cover fixing this up? I think there might be an interaction between this and the change @jamesmporter made in #1414, so I'm keen to see what the combined version of these changes is.

chrisprice (dismissed):

packages/d3fc-webgl/src/buffers/vertexConstantAttributeBuilder.js
josted: Missed that, nope.

packages/d3fc-webgl/src/buffers/vertexConstantAttributeBuilder.js
```diff
@@ -31,32 +31,22 @@ export default () => {
...
-                offset + componentsPerElement
-            );
-        }
         return projectedData;
     };
 
     const build = (gl, program, name, verticesPerElement, count) => {
         base(gl, program, name);
 
-        if (validSize >= count) {
```
OliverForeman, line 51: Why does this need to change from `count`?

packages/d3fc-webgl/src/buffers/vertexConstantAttributeBuilder.js
```diff
@@ -7,12 +7,12 @@ export default () => {
     let data = null;
     let validSize = 0;
```
OliverForeman: Not necessarily part of this PR but can we change this to use `validSize` from `baseAttributeBuilder`, the same way that `elementConstantAttributeBuilder` does

packages/d3fc-webgl/src/buffers/vertexConstantAttributeBuilder.js
OliverForeman: Does this need to be `(offset + 1) * componentsPerElement`

OliverForeman: If I remove the `+ 1` then the charts don't render

packages/d3fc-webgl/src/buffers/baseAttributeBuilder.js
josted: Done https://github.com/d3fc/d3fc/issues/1421

packages/d3fc-webgl/src/buffers/vertexConstantAttributeBuilder.js
josted: Sure

packages/d3fc-webgl/src/buffers/vertexConstantAttributeBuilder.js
josted: I don't know. It works either way so I'll change it to be consistent, though just in case, it's possible this appeared when Chris and I paired on it last Friday, do you remember if there was a particular reason for this? @chrisprice

packages/d3fc-webgl/src/buffers/vertexConstantAttributeBuilder.js
OliverForeman: This is working fine for me now, I must have made another change that affected it without realising

packages/d3fc-webgl/src/buffers/vertexConstantAttributeBuilder.js
josted: Actually makes most sense as verticesPerElement, changed.

chrisprice (approved):

PR #1409, opened 2020-01-27
refactor: move boxplot web-gl logic from JS into GLSL

jamesmporter:


packages/d3fc-webgl/src/shaders/vertexShaderSnippets.js
```diff
@@ -247,12 +247,45 @@ export const area = {
...
-        attribute float aYDirection;
-        attribute float aBandwidth;
+    header: `
+        attribute float aXValue;
+        attribute float aHigh;
+        attribute float aUpperQuartile;
+        attribute float aMedian;
+        attribute float aLowerQuartile;
+        attribute float aLow;
+        attribute float aBandwidthNew;
```
chrisprice: Strange name?

jamesmporter: Indeed... A relic of the refactor I think.

chrisprice (approved):

PR #1410, opened 2020-01-27
refactor: move defined call to after dirty check

OliverForeman:
Changes d3fc-series WebGL implementations to cache `filteredData` so that it doesn't have to be calculated on every render


packages/d3fc-series/src/webgl/bar.js
```diff
@@ -10,15 +10,15 @@ export default () => {
 
     let equals = (previousData, data) => false;
     let previousData = [];
+    let filteredData = [];
```
chrisprice, line 13: I'm not sure this is the right scope, can this be moved into the function below?

chrisprice (changes requested):

packages/d3fc-series/src/webgl/bar.js
OliverForeman: If this is in the function below and the data is equal (we aren't reprocessing) then we will be using an empty array for the `decorate` function and the `draw` call.

If we don't store the filtered data between renders then I think we will have to call `defined` every render

chrisprice: Deferred until https://github.com/d3fc/d3fc/issues/1412

chrisprice (approved):

PR #1414, opened 2020-01-29
feat: add element indexing

jamesmporter:


packages/d3fc-webgl/src/program/programBuilder.js
```diff
@@ -26,7 +27,20 @@ export default () => {
         );
         buffers(context, program, verticesPerElement, count);
 
-        context.drawArrays(mode, 0, count * verticesPerElement);
+        let elementBuffer = null;
+        if (elementIndices != null) {
+            elementBuffer = elementIndices(context, verticesPerElement, count);
+            context.bindBuffer(context.ELEMENT_ARRAY_BUFFER, elementBuffer);
+
+            context.drawElements(
```
chrisprice: This doesn't look like the right argument list https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/drawElements

packages/d3fc-webgl/src/program/programBuilder.js
```diff
@@ -26,7 +27,20 @@ export default () => {
         );
         buffers(context, program, verticesPerElement, count);
 
-        context.drawArrays(mode, 0, count * verticesPerElement);
+        let elementBuffer = null;
+        if (elementIndices != null) {
+            elementBuffer = elementIndices(context, verticesPerElement, count);
+            context.bindBuffer(context.ELEMENT_ARRAY_BUFFER, elementBuffer);
```
chrisprice: Is this being done twice?

packages/d3fc-webgl/src/program/programBuilder.js
```diff
@@ -10,6 +10,7 @@ export default () => {
     let mode = drawModes.TRIANGLES;
     let buffers = bufferBuilder();
     let verticesPerElement = 1;
+    let elementIndices = null;
```
chrisprice: Is prefer all the buffers stayed in buffer builder or moved into here. As keeping them in buffer builder is the smaller change can we go for that for now?

packages/d3fc-webgl/src/buffers/elementIndicesBuilder.js
```diff
@@ -0,0 +1,47 @@
...
+            return;
+        }
+        if (buffer == null) {
+            buffer = gl.createBuffer();
+        }
+        gl.bindBuffer(gl.ELEMENT_ARRAY_BUFFER, buffer);
+
+        const bufferArray = [];
+        for (let index = 0; index < count; index++) {
+            bufferArray.push(...data.map(x => x + index * verticesPerElement));
```
chrisprice: Am I right in thinking this is only required until we get the ANGLE stuff in?

packages/d3fc-webgl/src/buffers/elementIndicesBuilder.js
```diff
@@ -0,0 +1,47 @@
...
+    const base = (gl, verticesPerElement, count) => {
+        if (!changed) {
+            return;
+        }
+        if (buffer == null) {
+            buffer = gl.createBuffer();
+        }
+        gl.bindBuffer(gl.ELEMENT_ARRAY_BUFFER, buffer);
+
+        const bufferArray = [];
```
chrisprice: Can we assign directly to a Uint16Array to save the latter copy operation?

chrisprice (changes requested):
Nice one on getting to the bottom of why this wasn't working. Can you roll it out to the other sites types as well?

packages/d3fc-webgl/src/program/programBuilder.js
jamesmporter: Good point. I'll change the offset to `0`.

packages/d3fc-webgl/src/program/programBuilder.js
jamesmporter: Yes, it doesn't need to be. I've removed this occurrence.

packages/d3fc-webgl/src/program/programBuilder.js
jamesmporter: Agreed - I'll change that over.

packages/d3fc-webgl/src/buffers/elementIndicesBuilder.js
jamesmporter: Yes

packages/d3fc-webgl/src/buffers/elementIndicesBuilder.js
jamesmporter: Done.

jamesmporter:
@chrisprice I've addressed all of your comments.

packages/d3fc-webgl/src/program/programBuilder.js
```diff
@@ -26,7 +26,16 @@ export default () => {
         );
         buffers(context, program, verticesPerElement, count);
 
-        context.drawArrays(mode, 0, count * verticesPerElement);
+        if (buffers.elementIndices() != null) {
+            context.drawElements(
+                mode,
+                buffers.elementIndices().data().length * count,
```
chrisprice: `buffers.elementIndices().data().length` === `verticesPerElement`?

packages/d3fc-webgl/src/buffers/elementIndicesBuilder.js
```diff
@@ -0,0 +1,54 @@
...
+    const base = (gl, cornerCount, elementCount) => {
+        if (!changed) {
+            return;
+        }
+        if (buffer == null) {
+            buffer = gl.createBuffer();
+        }
+        gl.bindBuffer(gl.ELEMENT_ARRAY_BUFFER, buffer);
+
+        const vertexCount = data.length;
```
chrisprice: Is this what we'd normally call this `verticesPerElement`? Should we pass it in here and assert that data is the right length?

chrisprice (changes requested):
A few minor comments on the fallout from the changes. I'm keen to get #1407 in first if we can so that we can test the interaction I mentioned in a comment on the PR.

packages/d3fc-webgl/src/program/programBuilder.js
jamesmporter: No.

`verticesPerElement` = number of corners defined.
`elementIndices length` = number of vertices defined (via indices).

I found this quite confusing when writing the code so perhaps we should change one of these names to clarify what is going on.

packages/d3fc-webgl/src/buffers/elementIndicesBuilder.js
jamesmporter: Unfortunately no, as per my comment above.

Also, due to the *ANGLE work, this line has been removed.

packages/d3fc-webgl/src/program/programBuilder.js
chrisprice: Of course. I think we should now be able to remove `verticesPerElement`, although let's do that in another PR.

chrisprice (approved):

PR #1415, opened 2020-01-29
IDEA - add helper functions to vertex shaders

jamesmporter:
This is an idea for a refactor we could do in order to improve the readability and maintainability of the shader code.

The idea is to use helper functions to hide the boolean logic and make it more obvious that conditional blocks are effectively being performed.


chrisprice:
Interesting idea. I think we already have this in one of the shaders. If possible (i.e. no impact on performance tested using examples) I'd much rather we just use conditionals and Boolean logic.

Issue #1396, opened 2020-01-21
Formalise naming convention for attributes/uniforms


chrisprice:
As the shaders are so dependent on the attribute names, we should formalise the naming convention and ensure we follow it.

chrisprice:
* Define conventions in `packages/d3fc-webgl/README.md` for attribute & uniform naming
* Explain that variables used in the shaders are implementation details and subject to change
* Ensure all shader code follows conventions
* Ensure similar variables/methods found elsewhere in the code are consistent

PR #1416, opened 2020-01-29
Shader naming convention

jamesmporter:
Here's an idea for a naming convention for the shader attributes. I have opened this as a draft PR so you can give your feedback on the convention so far.

Addresses #1396

### DIFF
--- a/packages/d3fc-webgl/src/series/glArea.js
+++ b/packages/d3fc-webgl/src/series/glArea.js
@@ -48,12 +48,12 @@ export default () => {
 
     program
         .buffers()
-        .attribute('aXValue', xValueAttribute)
-        .attribute('aXPrevValue', xPreviousValueAttribute)
-        .attribute('aYValue', yValueAttribute)
-        .attribute('aYPrevValue', yPreviousValueAttribute)
-        .attribute('aY0Value', y0ValueAttribute)
-        .attribute('aY0PrevValue', y0PreviousValueAttribute)
+        .attribute('aCrossValue', xValueAttribute)
+        .attribute('aCrossPrevValue', xPreviousValueAttribute)
+        .attribute('aMainValue', yValueAttribute)
+        .attribute('aMainPrevValue', yPreviousValueAttribute)
+        .attribute('aBaseValue', y0ValueAttribute)
+        .attribute('aBasePrevValue', y0PreviousValueAttribute)
         .attribute('aCorner', cornerAttribute)
         .attribute('aDefined', definedAttribute);
 
--- a/packages/d3fc-webgl/src/series/glBar.js
+++ b/packages/d3fc-webgl/src/series/glBar.js
@@ -68,10 +68,10 @@ export default () => {
             .buffers()
             .elementIndices(elementIndices)
             .attribute('aCorner', cornerAttribute)
-            .attribute('aXValue', xValueAttribute)
-            .attribute('aYValue', yValueAttribute)
-            .attribute('aY0Value', y0ValueAttribute)
-            .attribute('aWidthValue', widthValueAttribute);
+            .attribute('aCrossValue', xValueAttribute)
+            .attribute('aMainValue', yValueAttribute)
+            .attribute('aBaseValue', y0ValueAttribute)
+            .attribute('aBandwidth', widthValueAttribute);
 
         xScale.coordinate(0);
         xScale(program);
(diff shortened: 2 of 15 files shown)


packages/d3fc-webgl/src/series/glBar.js
```diff
@@ -66,9 +66,9 @@ export default () => {
         program
             .buffers()
             .attribute('aXValue', xValueAttribute)
-            .attribute('aYValue', yValueAttribute)
-            .attribute('aY0Value', y0ValueAttribute)
-            .attribute('aWidthValue', widthValueAttribute)
+            .attribute('aXWidth', widthValueAttribute)
```
chrisprice: Does series call this bandwidth?

packages/d3fc-webgl/src/series/glArea.js
```diff
@@ -52,8 +52,8 @@ export default () => {
         .attribute('aXPrevValue', xPreviousValueAttribute)
         .attribute('aYValue', yValueAttribute)
         .attribute('aYPrevValue', yPreviousValueAttribute)
-        .attribute('aY0Value', y0ValueAttribute)
-        .attribute('aY0PrevValue', y0PreviousValueAttribute)
+        .attribute('aYBase', y0ValueAttribute)
+        .attribute('aYPrevBase', y0PreviousValueAttribute)
```
chrisprice: Maybe it's time to go the whole hog and use main/cross/base/orient?

packages/d3fc-webgl/src/series/glBar.js
chrisprice: I'm not keen on the x/y prefixing because we'll ultimately need to support orientation

jamesmporter:
The cross has stroke-width instead of line-width.
@chrisprice Do you know why this is?

packages/d3fc-webgl/src/series/glBar.js
```diff
@@ -66,9 +66,9 @@ export default () => {
         program
             .buffers()
             .attribute('aXValue', xValueAttribute)
```
chrisprice: How do you feel about the (in)consistency of when we use the value suffix?

chrisprice:
> The cross has stroke-width instead of line-width.
> @chrisprice Do you know why this is?

I think it's a by-product of svg using `stroke` and canvas using `line`. We should pick one and stick with it. I'd recommend `stroke`

packages/d3fc-webgl/src/series/glBar.js
chrisprice: https://d3fc.io/api/series-api.html#ohlc

jamesmporter:
I'd like to differ from the [D3FC docs](https://d3fc.io/api/series-api.html#seriesPoint_size) on the point of `aSize`. The shaders will be clearer if we use `aArea`.

packages/d3fc-webgl/src/series/glBar.js
jamesmporter: I've changed this to cross and main

packages/d3fc-webgl/src/series/glArea.js
jamesmporter: I've implemented this.

packages/d3fc-webgl/src/series/glBar.js
jamesmporter: Implemented.

jamesmporter:
I've renamed all instances of `uLineWidth` to `uStrokeWidth`.

chrisprice:
> I'd like to differ from the [D3FC docs](https://d3fc.io/api/series-api.html#seriesPoint_size) on the point of `aSize`. The shaders will be clearer if we use `aArea`.

Yea I agree it's weird but we didn't come up with that one - https://github.com/d3/d3-shape#symbol_size

packages/d3fc-webgl/src/series/glPoint.js
```diff
@@ -19,9 +19,9 @@ export default () => {
 
     program
         .buffers()
-        .attribute('aXValue', xValueAttribute)
-        .attribute('aYValue', yValueAttribute)
-        .attribute('aSize', sizeAttribute);
+        .attribute('aCrossValue', xValueAttribute)
+        .attribute('aMainValue', yValueAttribute)
+        .attribute('aArea', sizeAttribute);
```
chrisprice: Please can you make this `aSize` and then I think we're good

chrisprice (changes requested):

packages/d3fc-webgl/src/series/glPoint.js
jamesmporter: Done.

chrisprice (approved):

PR #1417, opened 2020-01-30
refactor: change elementConstantAttributeBuilder loop comparison to less than

OliverForeman:
Changes the comparison in the loop for `elementConstantAttributeBuilder` from `<=` to `<`


josted (approved):
Well spotted.

chrisprice (approved):

PR #1419, opened 2020-01-30
refactor: added float array "factory" to recycle arrays

josted:
New and improved version of (https://github.com/d3fc/d3fc/issues/1374). 
Example of it working with logging with glBar:
image

Let me know any opinions on better approaches/further optimisations.


packages/d3fc-webgl/src/buffers/elementConstantAttributeBuilder.js
```diff
@@ -10,9 +11,9 @@ export default () => {
     const project = (elementCount, verticesPerElement) => {
         const components = base.size();
         const offset = base.offset();
-        const projectedData = new Float32Array(
-            offset + elementCount * verticesPerElement * components
-        );
+        const requiredLength =
+            offset + elementCount * verticesPerElement * components;
+        const projectedData = floatArrayFactory(data, requiredLength);
```
jamesmporter: `data` is not a `Float32Array` - it's just a standard javascript list. I think that @cprice meant that we should recycle `projectedData`.

josted: As discussed, `data` is actually a `Float32Array`, but you might be right about re-using projected data between attributeBuilders, if that's possible. Let's get more info and I can make changes if need be.

packages/d3fc-webgl/src/buffers/floatArrayFactory.js
```diff
@@ -0,0 +1,2 @@
+export default (data, requiredLength) =>
+    data.length >= requiredLength ? data : new Float32Array(requiredLength);
```
chrisprice: This makes presumptions about `data`. I'd expect function to be a component which caches the previously requested array and optionally returns it if it's still useful.

chrisprice (changes requested):

packages/d3fc-webgl/src/buffers/elementConstantAttributeBuilder.js
josted: Changed the whole thing, I'll resolve this for now.

packages/d3fc-webgl/src/buffers/floatArrayFactory.js
josted: Think it's closer to that vision now.

packages/d3fc-webgl/src/buffers/float32ArrayFactory.js
```diff
@@ -0,0 +1,12 @@
+function float32ArrayFactory(requiredLength) {
```
chrisprice: I'm not sure this is doing what you expect it to. The `cachedArray` is effectively global.

josted: Thanks, makes more sense now, changed.

chrisprice (approved):

PR #1420, opened 2020-01-31
feat: add unit and performance tests for webgl line series

OliverForeman:
- Adds a unit test for the line shader
- Adds a line example rendering 100,000 data points
- Adds a performance test for the line example


packages/d3fc-webgl/test/shaders/line/shaderSpec.js
```diff
@@ -0,0 +1,39 @@
+import * as fc from '../../../index';
+import { expectVertexShader, getShaders } from '../../util/vertextShaderUtil';
+
+it('test', () => {
```
jamesmporter: Could you rename this to describe the test?

packages/d3fc-webgl/test/shaders/line/shaderSpec.js
```diff
@@ -0,0 +1,39 @@
+import * as fc from '../../../index';
+import { expectVertexShader, getShaders } from '../../util/vertextShaderUtil';
+
```
jamesmporter, line 3: Could you add more test cases?

- 1st point
- 2nd point
- A middle point
- Last point

packages/d3fc-webgl/test/shaders/line/shaderSpec.js
```diff
@@ -0,0 +1,39 @@
...
+        linear0Scale: [1, 1, 1, 1],
+        linear1Offset: [0, 0, 0, 0],
+        linear1Scale: [1, 1, 1, 1]
+    };
+
+    expectVertexShader(
+        shaders.vertexShader(),
+        attributes,
+        uniforms
+    ).toHaveGlPosition([1, 4.991055728090001, 0, 1]);
```
jamesmporter: @jamesmporter Can you make another method which takes a rounded input value to 4dp?

packages/d3fc-webgl/test/shaders/line/shaderSpec.js
OliverForeman: Yeah, missed that 👍

packages/d3fc-webgl/test/shaders/line/shaderSpec.js
chrisprice: As we're always dealing with floats, does it make sense to modify `toHaveGlPosition` to use `toBeCloseTo` or equivalent?

jamesmporter: Addressed in: https://github.com/d3fc/d3fc/pull/1422

jamesmporter: @chrisprice I that it's better to use the map/round solution than `toBeCloseTo` as the former gives us error information about the whole `gl_Position`.

```
shaderOutput.gl_Position = shaderOutput.gl_Position.map(
    x => Math.round((x + Number.EPSILON) * 10000) / 10000
);

return expect(shaderOutput).toEqual({
    gl_Position: glPosition
});
```

```
shaderOutput.gl_Position.forEach((x, i) => {
    expect(x + Number.EPSILON).toBeCloseTo(glPosition[i], 4);
});
```

jamesmporter: However, I've made the 4dp function the default (`toHaveGlPosition()`) and added another (`toHaveGlPositionExact()`), which compares the non-rounded float value.

chrisprice (approved):

PR #1424, opened 2020-01-31
feat: webgl earth and mars example

OliverForeman:
Adds the Earth and Mars orbit data example that @mattstobbs created

Additionally this adds an `isProgram` check in `programBuilder` to ensure that the program is valid as otherwise this causes issues when switching between canvas and webgl implementations


examples/earth-and-mars/index.js
```diff
@@ -0,0 +1,202 @@
...
+
+const crossValue = d => d.date;
+const mainValue = d => d.distance;
+
+const blueGradientDecorate = program => {
+    program.fragmentShader().appendBody(`
+        vec4 color = vec4(70.0/255.0, 130.0/255.0, 180.0/255.0, 1.0);
+        float colorModifier = 1.0 - gl_FragCoord.y / 421.0;
+        colorModifier = smoothstep(0.875, 1.0, colorModifier);
+        gl_FragColor = vec4(color + ((vec4(1.0, 1.0, 1.0, 1.0) - color) * colorModifier));
```
jamesmporter: I think you can do `1.0 - color` here.

examples/earth-and-mars/index.js
```diff
@@ -0,0 +1,202 @@
...
+    `);
+};
+
+const webglArea = fc
+    .seriesWebglArea()
+    .mainValue(mainValue)
+    .crossValue(crossValue)
+    .defined(() => true)
```
jamesmporter, line 43: Does defined have a default value? I think it probably should.

examples/earth-and-mars/index.js
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    render();
+});
+addShowMultiColor();
+
+const chart = fc
+    .chartCartesian(x, y)
+    .yOrient('left')
+    .yLabel('Distance (m KM)')
```
jamesmporter: Usually, units of measure are given in Propercase, so this would read `m Km`. However, assuming this means "million kilometres", this should be written as "giga-metres" `Gm`. On the other hand, as most programmers won't understand that perhaps `Km x10^6` or something to that effect may be more clear.

jamesmporter:
There appears to be no axis label on the horizontal axis.

jamesmporter:
Could you add a title to the chart? Something along the lines of "The distance between Earth and Mars over time".

examples/earth-and-mars/index.js
jamesmporter: Perhaps `million Km` works best.

examples/earth-and-mars/index.js
OliverForeman: `defined` has a default implementation that involves a `for` loop. This optimises the defined check so we don't run a loop for every data point

examples/earth-and-mars/index.js
OliverForeman: As discussed this will be changed to `million Km` for clarity

OliverForeman:
Currently blocked as it contains an error when switching from the WebGL implementation to Canvas and back again. This should be addressed by PR #1475

packages/d3fc-webgl/src/program/programBuilder.js
```diff
@@ -90,7 +90,7 @@ export default () => {
     return build;
 
     function newProgram(program, vertexShader, fragmentShader) {
```
chrisprice: This is no longer required.

examples/earth-and-mars/index.js
```diff
@@ -0,0 +1,204 @@
...
+
+    d3.select('#chart')
+        .datum(currentData)
+        .call(chart);
+
+    zoom.transform(d3.select('.plot-area'), d3.zoomIdentity);
+};
+
+// eslint-disable-next-line no-undef
+const gui = new dat.GUI();
```
chrisprice, line 148: Can we just use standard DOM elements for this?

examples/earth-and-mars/worker.js
```diff
@@ -0,0 +1,214 @@
...
+};
+
+const parseDate = d3.timeParse('%d %B %Y');
+
+onmessage = ({ data }) => {
+    const { numPoints } = data;
+
+    setup();
+
+    const newData = d3.range(numPoints).map(() => {
```
chrisprice, line 204: `Array.from({ length: numPoints })`

examples/earth-and-mars/worker.js
```diff
@@ -0,0 +1,214 @@
...
+        draw();
+        const date = dayToDate(Math.round(_day), _year);
+        return {
+            distance: Math.round(getKM(d)),
+            date: parseDate(date[0] + ' ' + date[1] + ' ' + _year)
```
chrisprice, line 209: Don't roundtrip via string

examples/earth-and-mars/worker.js
```diff
@@ -0,0 +1,214 @@
+// eslint-disable-next-line no-undef
+importScripts('../../node_modules/d3/dist/d3.js');
```
chrisprice, line 2: Don't think this is required.

examples/earth-and-mars/index.js
```diff
@@ -0,0 +1,204 @@
...
+
+const requestRedraw = () => {
+    d3.select('d3fc-group')
+        .node()
+        .requestRedraw();
+};
+
+const zoom = d3.zoom().on('zoom', () => {
+    x.domain(d3.event.transform.rescaleX(x2).domain());
+    requestRedraw();
```
chrisprice, line 90: This should just call render

examples/earth-and-mars/index.html
```diff
@@ -0,0 +1,14 @@
+<html>
```
chrisprice, line 1: Match HTML content from other examples

chrisprice (changes requested):

PR #1426, opened 2020-02-03
test: add puppeteer config and run in band

OliverForeman:
Adds the `runInBand` flag to prevent the tests running in parallel

Adds a puppeteer config file to allow headless mode and executable path to be configured for puppeteer (set via the command line using `HEADLESS` and `EXECUTABLE_PATH`).
This is needed as puppeteer does not use the GPU when ran in headless mode, see this issue https://github.com/puppeteer/puppeteer/issues/3637


package.json
```diff
@@ -14,7 +14,7 @@
     "screenshots": "lerna run screenshots --stream --concurrency 1",
     "start": "lerna run bundle && npm start --prefix ./packages/d3fc",
     "commitlint": "commitlint --from HEAD~1 to HEAD --verbose",
-    "examples": "cd examples && npx jest"
+    "examples": "cd examples && npx jest -i"
```
chrisprice: Please can you use the long version of the run in band flag? Long flags makes it easier for the next person reading the script, short flags make it easier for you to type into a console 😄

chrisprice (approved):

Issue #1404, opened 2020-01-24
elementConstantAttributeBuilder doesn't allow size > 1


chrisprice:
The `value` function does not consider `component`. Therefore an attribute built by a `elementConstantAttributeBuilder().size(2)` would not be able to supply values for the second component of the `vec2`.

PR #1431, opened 2020-02-04
feat: allow multiple components with elementConstantAttributeBuilder

jamesmporter:
Addresses: #1404

**Example usage:**
```js
    const xYValueAttribute = elementConstantAttributeBuilder()
        .size(2)
        .value((data, element, component) => data[element][component]);
```

### DIFF
--- a/packages/d3fc-webgl/src/buffers/elementConstantAttributeBuilder.js
+++ b/packages/d3fc-webgl/src/buffers/elementConstantAttributeBuilder.js
@@ -6,23 +6,33 @@ export default () => {
     const base = baseAttributeBuilder().divisor(1);
     const factory = float32ArrayFactory();
 
-    let value = (data, element, vertex, component, index) => data[element];
+    let value = (data, element) => data[element];
     let data = null;
 
     const project = elementCount => {
-        const components = base.size();
+        const componentCount = base.size();
         const offset = base.offset();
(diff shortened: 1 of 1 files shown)


chrisprice (changes requested):
As discussed please can we remove the `component` argument from the `value` function and return a tuple of component values instead.

packages/d3fc-webgl/src/buffers/elementConstantAttributeBuilder.js
```diff
@@ -6,23 +6,33 @@ export default () => {
...
+        if (componentCount > 1) {
+            for (let element = 0; element < elementCount; element++) {
+                for (
+                    let component = 0;
+                    component < componentCount;
+                    component++
+                ) {
+                    projectedData[
+                        offset + element * componentCount + component
+                    ] = value(data, element)[component];
```
chrisprice, line 27: Won't this invoke the function `componentCount` times more than required?

chrisprice (changes requested):

packages/d3fc-webgl/src/buffers/elementConstantAttributeBuilder.js
OliverForeman: This is now addressed in #1436

OliverForeman:
This is now duplicated by #1436

Issue #1412, opened 2020-01-27
Handling of defined is inconsistent in d3fc-series WebGL implementations


OliverForeman:
For `area` and `line`, `defined` is handled as a property of `glArea` and `glLine` respectively. For the other implementations it is handled via the `base`.

This also affects the data passed to the decorate function as `area` and `line` are passed the full data, whereas the other implementations are passed `filteredData`

chrisprice:
For consistency, always add a `defined` buffer and discard at the shader level. This means that `decorate` will always see the unfiltered `data`.

PR #1432, opened 2020-02-04
refactor: add defined attribute to all webgl series

OliverForeman:
Adds `definedAttribute` to all WebGL series and discards undefined elements in the `fragmentShader`.

This solves #1412 

This can cause inconsistencies in rendering between SVG/Canvas implementations and WebGL implementations, for example if the index is used for the cross value (as seen below).

Canvas rendering:
image

WebGL rendering:
image

### DIFF
--- a/packages/d3fc-series/src/webgl/bar.js
+++ b/packages/d3fc-series/src/webgl/bar.js
@@ -10,7 +10,6 @@ export default () => {
 
     let equals = (previousData, data) => false;
     let previousData = [];
-    let filteredData = [];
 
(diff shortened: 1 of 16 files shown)


OliverForeman:
As an alternative to this implementation the changes to the vertex and fragment shaders could be added to the default shaders in `shaderBuilder` as they are consistent across all implementations.
This would reduce the shader code in each individual implementation but would hide the details of the defined implementation.

I would be interested in thoughts on this idea?

chrisprice (approved):

PR #1441, opened 2020-02-07
feat: add webgl and canvas plot area example

### LINKED ISSUE #1438: Webgl based gridlines
Is there a built-in way to render grid lines using webgl? I'm trying to move over from canvas and noticed there is no annotationWebglGridline.

I'm aware of the lack of feature parity with the webgl components, but is there any progress or beta code I could use for this one? Generally, what are some of the next things being worked on for d3fc-webgl? 

Alternatively, is there any way to combine them? Like:

`seriesMulti().series([ canvasGridlineSeries, webglCandlesSeries ])`

This doesn't work because ultimately I need to call .webglPlotArea() on the cartesian

OliverForeman:
Adds a chart demonstrating how to use both `webglPlotArea` and `canvasPlotArea` together.

This resolves #1438

### DIFF
--- a/examples/svg-canvas-and-webgl-chart/README.md
+++ b/examples/svg-canvas-and-webgl-chart/README.md
@@ -1,0 +1,3 @@
+# Webgl and Canvas Plot
+
+This example demonstrates how a Webgl series and a Canvas series can be rendered together by drawing a Webgl candlestick series with a Canvas gridline series.
--- a/examples/svg-canvas-and-webgl-chart/index.html
+++ b/examples/svg-canvas-and-webgl-chart/index.html
@@ -1,0 +1,24 @@
+<html>
+
+<head>
+	<script src="../../node_modules/seedrandom/seedrandom.js"></script>
+	<script>Math.seedrandom('a22ebc7c488a3a47');</script>
+    <script src="../../node_modules/d3/dist/d3.js"></script>
+    <script src="../../packages/d3fc/build/d3fc.js"></script>
+    <style>
+        body {
+            display: flex;
+        }
+
+        body>* {
+            flex: auto;
+        }
+    </style>
+</head>
+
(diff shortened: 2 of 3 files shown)


chrisprice (changes requested):
I think it would be good to round the example out by adding in an SVG component as well. I was thinking the crosshair but if you think something else is more inspiring feel free to use that.

examples/svg-canvas-and-webgl-chart/index.html
```diff
@@ -0,0 +1,24 @@
...
+        }
+
+        body>* {
+            flex: auto;
+        }
+    </style>
+</head>
+
+<body>
+	<div id="chart"></div>
```
chrisprice, line 20: Whitespace nitpic...

examples/svg-canvas-and-webgl-chart/index.html
```diff
@@ -0,0 +1,24 @@
+<html>
+
+<head>
+	<script src="../../node_modules/seedrandom/seedrandom.js"></script>
```
chrisprice, line 4: Whitespace nitpic

examples/svg-canvas-and-webgl-chart/README.md
```diff
@@ -0,0 +1,3 @@
+# Webgl and Canvas Plot
```
chrisprice, line 1: This file needs updating to reference the SVG series as well.

chrisprice (changes requested):

examples/svg-canvas-and-webgl-chart/README.md
chrisprice, line 1: Also needs `__tests__` adding.

chrisprice (changes requested):

PR #1452, opened 2020-02-11
refactor: glScales take identifier and component arguments

### LINKED ISSUE #1362: scale/scale.coordinate/scale.component naming is confusing
Calling `scale(program)` applies the scale to the `scale.coordinate`th value of `vec4 gl_Position`. Calling `scale.component(program, component)` applies the scale to the `scale.coordinate`th value of `vecN component` (making the implicit assumption that `N` > `scale.coordinate`).

The scales generate scaling uniforms based upon the `scale.coordinate`, with no-op values filling in the other values. Internally the `scale.coordinate` value is also used to generate the uniform identifiers. A templated shader is then appended to the program containing the generated identifiers.

I find the naming and the scope of `scale.coordinate` to be confusing. I also think that special-casing `scale(program)` to implicitly apply to `gl_Position` feels like magic (in a bad way). 

I propose the following changes -

* The signature of `scale(...)` is changed to always accept an `identifier`.
* The `scale.coordinate` property is removed and `scale(...)` is additionally changed to accept an `index`.

OliverForeman:
Refactors `glScaleX` to add an identifier and component argument, this removes the `coordinate` property and `scaleComponent` function.

New use looks like:
```javascript
xScale(program, 'gl_Position', 0);
yScale(program, 'gl_Position', 1);
```

This resolves #1362

### DIFF
--- a/packages/d3fc-webgl/src/scale/glScaleBase.js
+++ b/packages/d3fc-webgl/src/scale/glScaleBase.js
@@ -1,7 +1,6 @@
 export default () => {
     let domain = [0, 1];
     let range = [-1, 1];
-    let coordinate = 0;
 
     const base = () => {};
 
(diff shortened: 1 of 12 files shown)


chrisprice:
Fixes #1362

chrisprice (approved):

PR #1454, opened 2020-02-11
refactor: renamed webgl package exports

OliverForeman:
- Rename `buffers` directory to `buffer`
- Refactor top level exports from `d3fc-webgl`


packages/d3fc-webgl/src/scale/glScaleLinear.js
```diff
@@ -1,4 +1,4 @@
-import uniform from '../buffers/uniform';
```
chrisprice, line 1: Does it make sense to rename these files as well? E.g. linear.js

chrisprice (changes requested):

chrisprice (approved):

PR #1457, opened 2020-02-12
d3fc-pointer - Set package.json main entry to build/d3fc-pointer.js

parliament718:
I was having trouble importing the pointer in my project via

```import { pointer } from '@d3fc/d3fc-pointer';```

When I compared d3fc-pointer to other d3fc packages, I noticed this minor difference, which in turn solved my import problem.


chrisprice:
Thanks for the report and fix. I've tweaked your change to match the other packages and our commit conventions in #1459.

PR #1467, opened 2020-02-18
chore: updated links in webgl docs

OliverForeman:
Changes the link for the `normalized` property to the correct page


chrisprice (approved):

chrisprice:
@OliverForeman If it's docs changes I generally opt for `docs:` rather than `chore:`

Issue #1470, opened 2020-02-18
Handling of errors for shader compiling and attribute/uniform binding


OliverForeman:
Currently there is no handling for any errors/warnings that occur during shader compilation or when binding attributes/uniforms. This results in the program potentially exiting without rendering and outputting a WebGL generated message.

Ideally we want to be able to automatically recover from these potential error states but not if it comes with a significant performance hit.

OliverForeman:
If we ever encounter an error that results in a blank screen/no draw we should evaluate how to handle this. If the default error given by the browser/WebGL is appropriate then this can be left, otherwise we should throw an error ourselves to indicate what caused the error.

chrisprice:
> Currently there is no handling for any errors/warnings that occur during shader compilation or when binding attributes/uniforms. This results in the program potentially exiting without rendering and outputting a WebGL generated message.
> 
> Ideally we want to be able to automatically recover from these potential error states but not if it comes with a significant performance hit.

Compilation/linking errors triggers warnings from the browser in the console. For further detail, set the [`program.debug(true)` flag](https://github.com/d3fc/d3fc/tree/master/packages/d3fc-webgl#webglProgramBuilder_debug).

chrisprice:
> If we ever encounter an error that results in a blank screen/no draw we should evaluate how to handle this. If the default error given by the browser/WebGL is appropriate then this can be left, otherwise we should throw an error ourselves to indicate what caused the error.

We should raise specific examples of this happening as separate issues

PR #1473, opened 2020-02-25
Webgl boxplot fill

adilparvez:
Adds fill to webgl box plot.

New default.
webgl-boxplot-defualt

With stroke and fill.
webgl-boxplot-fill-and-stroke

### DIFF
--- a/packages/d3fc-annotation/package.json
+++ b/packages/d3fc-annotation/package.json
@@ -24,7 +24,7 @@
   },
   "devDependencies": {
     "babel-polyfill": "^6.16.0",
-    "canvas": "2.3.1",
+    "canvas": "2.6.1",
     "d3": "^4.3.0",
     "webshot": "^0.18.0"
   },
--- a/packages/d3fc-series/examples/boxplot.js
+++ b/packages/d3fc-series/examples/boxplot.js
@@ -71,5 +71,4 @@ var webglBoxPlot = fc.autoBandwidth(fc.seriesWebglBoxPlot())
     .lowerQuartileValue(function(d) { return d.lowerQuartile; })
     .highValue(function(d) { return d.high; })
     .lowValue(function(d) { return d.low; });
-
 webglBoxPlot(data);
--- a/packages/d3fc-series/examples/boxplot.js
+++ b/packages/d3fc-series/examples/boxplot.js
@@ -70,9 +70,6 @@ var webglBoxPlot = fc.autoBandwidth(fc.seriesWebglBoxPlot())
     .upperQuartileValue(function(d) { return d.upperQuartile; })
     .lowerQuartileValue(function(d) { return d.lowerQuartile; })
     .highValue(function(d) { return d.high; })
-    .lowValue(function(d) { return d.low; })
-    .decorate((program) => {
-        // fc.webglFillColor([0.6, 0.6, 0.3, 1.0])(program);
-        // fc.webglStrokeColor([0.4, 0.2, 0.0, 1.0])(program);
-    });
+    .lowValue(function(d) { return d.low; });
+
 webglBoxPlot(data);
--- a/packages/d3fc-series/examples/boxplot.js
+++ b/packages/d3fc-series/examples/boxplot.js
@@ -70,5 +70,9 @@ var webglBoxPlot = fc.autoBandwidth(fc.seriesWebglBoxPlot())
     .upperQuartileValue(function(d) { return d.upperQuartile; })
     .lowerQuartileValue(function(d) { return d.lowerQuartile; })
     .highValue(function(d) { return d.high; })
-    .lowValue(function(d) { return d.low; });
+    .lowValue(function(d) { return d.low; })
+    .decorate((program) => {
+        // fc.webglFillColor([0.6, 0.6, 0.3, 1.0])(program);
+        // fc.webglStrokeColor([0.4, 0.2, 0.0, 1.0])(program);
+    });
 webglBoxPlot(data);
--- a/packages/d3fc-webgl/src/buffer/attributeProjector.js
+++ b/packages/d3fc-webgl/src/buffer/attributeProjector.js
@@ -49,14 +49,7 @@ export default () => {
         if (!args.length) {
             return size;
         }
-
         size = args[0];
-        if (![1, 2, 3, 4].includes(size)) {
-            throw new Error(
-                'Number of components per vertex attribute must be 1, 2, 3 or 4.'
-            );
-        }
-
         dirty = true;
         return projector;
     };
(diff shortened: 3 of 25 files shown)


packages/d3fc-annotation/package.json
```diff
@@ -24,7 +24,7 @@
   },
   "devDependencies": {
     "babel-polyfill": "^6.16.0",
-    "canvas": "2.3.1",
```
OliverForeman, line 27: Does this canvas version need bumping up? I'm not sure what its relation is to this PR?

packages/d3fc-series/examples/boxplot.js
```diff
@@ -70,5 +70,9 @@ var webglBoxPlot = fc.autoBandwidth(fc.seriesWebglBoxPlot())
     .upperQuartileValue(function(d) { return d.upperQuartile; })
     .lowerQuartileValue(function(d) { return d.lowerQuartile; })
     .highValue(function(d) { return d.high; })
-    .lowValue(function(d) { return d.low; });
+    .lowValue(function(d) { return d.low; })
+    .decorate((program) => {
+        // fc.webglFillColor([0.6, 0.6, 0.3, 1.0])(program);
+        // fc.webglStrokeColor([0.4, 0.2, 0.0, 1.0])(program);
+    });
```
OliverForeman: Can we remove this?

packages/d3fc-webgl/src/buffer/attributeProjector.js
```diff
@@ -49,7 +49,14 @@ export default () => {
         if (!args.length) {
             return size;
         }
+
         size = args[0];
+        if (![1, 2, 3, 4].includes(size)) {
+            throw new Error(
+                'Number of components per vertex attribute must be 1, 2, 3 or 4.'
```
OliverForeman: I like the introduction of an error message for this case however I don't know if it needs to be included in this PR. Issue #1470 should cover cases like this.

packages/d3fc-webgl/src/series/boxPlot.js
```diff
@@ -105,79 +105,82 @@ export default () => {
...
-            [-1, 2, 1, 1]
+            [-1, 2, 1, 1],
+            // Fill
+            [-1, -1, 1, 0],
+            [-1, 1, 1, 0],
+            [1, -1, 1, 0],
+            [1, 1, 1, 0]
+        ]);
+
+    /* eslint-disable */
```
OliverForeman: Can we follow the linting rules here please

packages/d3fc-webgl/src/series/boxPlot.js
```diff
@@ -105,79 +105,82 @@ export default () => {
...
+    const canFillAttribute = vertexAttribute()
```
OliverForeman: I think we can implement fill without having to introduce a new attribute. The vertices we need for the fill can be uniquely identified as a group from the `aCorner` values and we can then manipulate the fill/stroke colour within the shader.

I have a rough idea of how this should work and I'm happy to talk through it in person

packages/d3fc-webgl/src/series/boxPlot.js
```diff
@@ -105,79 +105,82 @@ export default () => {
...
+            // Bottom whisker line
+            0, 0, 0, 0,
+            // Bottom cap line
+            0, 0, 0, 0,
+            // Fill
+            1, 1, 1, 1
         ]);
+    /* eslint-enable */
 
+    /* eslint-disable */
```
OliverForeman: Can we follow the linting rules here please

packages/d3fc-webgl/src/shaders/fragmentShaderSnippets.js
```diff
@@ -82,15 +94,27 @@ export const ohlc = {
...
+    header: `
+        varying float vDefined;
+        varying float vCanFill;
+    `,
     body: `
+        float canFill = vCanFill;
+        float canStroke = 1.0 - vCanFill;
+
+        vec4 defaultFillColor = vec4(0.86, 0.86, 0.86, 1.0);
+        gl_FragColor = (canFill * defaultFillColor) + ((1.0 - canFill) * gl_FragColor);
```
OliverForeman: `(1.0 - canFill)` can be replaced with `canStroke`

OliverForeman (changes requested):
Can we change the target of this PR to the webgl branch please.

I've mentioned in a comment that I think this can be done slightly differently to avoid adding an extra attribute, I'm happy to talk through it

packages/d3fc-annotation/package.json
adilparvez: This was causing npm install to fail on node 12, since the older canvas version doesn't have prebuilt binaries for this node version. It attempts to build from source which needs a bunch of other native dependencies to succeed, so I bumped the version up.

packages/d3fc-series/examples/boxplot.js
adilparvez: Done.

packages/d3fc-webgl/src/buffer/attributeProjector.js
adilparvez: Removed.

packages/d3fc-webgl/src/series/boxPlot.js
adilparvez: Done. I had grouped them to make it clear which triangles were being referenced.

packages/d3fc-webgl/src/series/boxPlot.js
adilparvez: Done.

packages/d3fc-webgl/src/series/boxPlot.js
adilparvez: Done.

packages/d3fc-webgl/src/shaders/fragmentShaderSnippets.js
adilparvez: Done.

packages/d3fc-webgl/src/shaders/fragmentShaderSnippets.js
```diff
@@ -128,6 +135,10 @@ export const bar = {
     body: `
         float canFill = 1.0;
         float canStroke = 0.0;
+
+        vec4 defaultFillColor = vec4(0.60, 0.60, 0.60, 1.0);
+        gl_FragColor = (canFill * defaultFillColor) + ((1.0 - canFill) * gl_FragColor);
```
OliverForeman: Can we just assign the default colour directly to `gl_FragColor` here, so `gl_FragColor = vec4(0.60, 0.60, 0.60, 1.0);`

At this point we know that `canFill = 1.0` and `canStroke = 0.0` so we know that this will always evaluate to using `defaultFillColor`

OliverForeman (changes requested):
Minor change and the branch needs rebasing, otherwise looks good

OliverForeman (approved):

Issue #1490, opened 2020-03-17
scaleMapper.js not working when project compiled using webpack 


gosthell:
The scaleMapper.js use the string representation of the scaler's copy function in order to determine the type of scaler (linear, log, ..). However, the copy function can be different for two similar scalers if a project is compiled using webpack and doesn't use the same d3 module as the one in d3fc-webgl.

`
function() {
   return Object(_continuous__WEBPACK_IMPORTED_MODULE_1__["copy"])(scale, linear());
}
`
`
function() {
    return Object(_continuous__WEBPACK_IMPORTED_MODULE_2__["copy"])(scale, linear());
  }
`

gosthell:
Proposed change
```
import * as d3Scale from 'd3-scale';
import linear from './linear';
import log from './log';
import pow from './pow';

export default scale => {
    let outScale = d3Scale.scaleIdentity();
    let glScale = null;
    const scaleCopyString = scale.copy.toString();

    // compare the stringified 'copy' function of the scale to
    // determine the scale type.
    if (scaleCopyString.search('linear') >= 0) {
        glScale = linear().domain(scale.domain());
    } else if (scaleCopyString.search('log') >= 0) {
        glScale = log()
            .domain(scale.domain())
            .base(scale.base());
    } else if (scaleCopyString.search('pow') >= 0) {
        glScale = pow()
            .domain(scale.domain())
            .exponent(scale.exponent());
    } else if (scaleCopyString.search('calendar') >= 0) {
        glScale = linear().domain(scale.domain());
    } else {
        glScale = linear().domain(scale.range());
        outScale = scale;
    }

    return {
        scale: outScale,
        glScale: glScale
    };
};
```

chrisprice:
Thanks for the report and the suggested fix. It's very frustrating that there isn't a "nice" way to do this and the current technique was the best we could come up with.

One thought that might be root issue of the behaviour your seeing would be 2 slightly different versions of d3-scale being imported. It might be worth checking that there's only 1 version in your `node_modules` tree (i.e. there might be a second version lurking in `node_modules/@d3fc/d3fc-webgl/node_modules`).

My concern with your suggested approach is that minification will most likely rewrite the identifiers that we're matching on. If the above doesn't work I can see 2 remaining options -

* Allow users to register explicit mappings of source scale to something (e.g. gl scale) with `scaleMapper`. If we store them internally in a `WeakMap`, I can't see any issues with lifecycle/memory management.
[...]

PR #1492, opened 2020-03-19
feat: improve rendering speed

gosthell:
Avoid rescaling main / cross values if the data hasn't changed and xScale or yScale is an identity scale

### DIFF
--- a/packages/d3fc-series/src/webgl/line.js
+++ b/packages/d3fc-series/src/webgl/line.js
@@ -41,7 +41,7 @@ export default () => {
         const xScale = webglScaleMapper(base.xScale());
         const yScale = webglScaleMapper(base.yScale());
 
-        if (!isIdentityScale(xScale.scale) || !isIdentityScale(yScale.scale) || !equals(previousData, data)) {
+        if (!equals(previousData, data)) {
             previousData = data;
 
             if (base.orient() === 'vertical') {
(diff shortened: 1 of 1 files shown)


gosthell:
I can do the same for the other series if you guys approve the change, and add those to the merge request

packages/d3fc-series/src/webgl/line.js
```diff
@@ -41,7 +41,7 @@ export default () => {
         const xScale = webglScaleMapper(base.xScale());
         const yScale = webglScaleMapper(base.yScale());
```
adilparvez, line 43: Setting `crossPreviousValueAttribute`, and `mainPreviousValueAttribute` is duplicated, what do you think of something like
```js
const setCrossPreviousValueAttribute = () => crossPreviousValueAttribute.value((d, i) => xScale.scale(base.crossValue()(d, i))).data(data);
const setMainPreviousValueAttribute = () => mainPreviousValueAttribute.value((d, i) => yScale.scale(base.mainValue()(d, i))).data(data);
```
And then calling these instead?

adilparvez:
> I can do the same for the other series if you guys approve the change, and add those to the merge request

After you've looked at the comment it would be great to have this applied to the other series too.

packages/d3fc-series/src/webgl/line.js
gosthell: Yes but depending on orientation you would need to change the argument of the scaler. So you would need to define 4 const
```
const setCrossPreviousValueAttributeCross = () => crossPreviousValueAttribute.value((d, i) => xScale.scale(base.crossValue()(d, i))).data(data);
const setCrossPreviousValueAttributeMain = () => crossPreviousValueAttribute.value((d, i) => xScale.scale(base.mainValue()(d, i))).data(data);
const setMainPreviousValueAttributeCross = () => mainPreviousValueAttribute.value((d, i) => yScale.scale(base.crossValue()(d, i))).data(data);
const setMainPreviousValueAttributeMain = () => mainPreviousValueAttribute.value((d, i) => yScale.scale(base.mainValue()(d, i))).data(data);
```

If you don't think this is too verbose I can do that

chrisprice (changes requested):
Sorry can we hold off this change until we have an agreed solution for #1490. I'm concerned that the original contract for when to run the projection wasn't clear and we may inadvertently change it in this PR which could make #1490 harder to fix without a breaking change.

chrisprice:
Superseded by #1500

PR #1493, opened 2020-03-20
refactor: inital migration of site to gatsby

mattstobbs:
This is the initial stage of migrating the d3fc website to use [GatsbyJS](https://www.gatsbyjs.org/).

### Examples
This PR adds all the examples used in the [d3fc introduction](https://d3fc.io/introduction/getting-started.html) pages. These examples include frontmatter for `title` and `section`, which is used to classify the examples on the [examples](https://d3fc.io/examples/) page. This frontmatter should be added to all future examples.

### Plugins

Because the examples are html files, which include `<head>` tags, they are currently being rendered using iframes. In order to do this, I've written a [custom plugin](https://www.gatsbyjs.org/docs/creating-plugins/), `examples-to-public`, which takes the required files from the examples folder and into the public folder at build time. This makes them available to use in iframes.

### src/Components

Contains React components used throughout the app.

The `SEO` component is a special component which uses [React Helmet](https://github.com/nfl/react-helmet) to define the site's metadata in the `<head>` tags.

### src/Pages

Any js file in here will attempt to create a page from the React in the file. Currently used for the home page and the examples page. Also holds the markdown for the introduction pages, which are generated in the [`gatsby-node`](#gatsby-node.js) file.

### src/static

Contains constant variables, such as information about external example pages and page urls.

### src/styles

Contains local scss files.

### src/templates

Contains the [page templates](https://www.gatsbyjs.org/tutorial/part-seven/).

### static

Contains static assets. These are automatically moved over to the public folder during build time by gatbsy.

### gatsby-config.js

[Defines the site's metadata](https://www.gatsbyjs.org/docs/api-files-gatsby-config/). Currently just defines all the [plugins](https://www.gatsbyjs.org/plugins/) used by gatsby, including the [custom one](#Plugins). 

### gatsby-node.js

[Used to help build the site](https://www.gatsbyjs.org/docs/api-files-gatsby-node/). This is where the introduction and individual example pages are programatically created.

Also creates an express server to host the public folder when running in develop mode, which is needed for the iframes to render correctly.

## Remaining Work
- Add the [API](https://d3fc.io/api/annotation-api.html) pages
- Examples
  - The rest of the examples from master will need to be merged in (with frontmatter added to each one).
  - Ensure that the data sources used by the examples are pointing to the package file (e.g. `bubble.index.js`, `stacked/index.js`).
  - Because the examples are being rendered in iframes, the site's global styling is not being applied to them. Therefore, any needed styling will need to be added to the example's `style.css` file. 
  - The examples currently use [unpkg](https://unpkg.com/) to import the external libraries. This should be replaced with the currently used method.
  - Rename "Non-block examples" to something more meaningful.
  - Generate the screenshots for all examples.
- A [404](https://www.gatsbyjs.org/docs/add-404-page/) page.
- Add d3fc-site to the npmignore file.
- Ensure package.json scripts are complete.
- Add remaining script and link tags to SEO component.
- Remove any unused static resources.
- I'm not 100% sure but I think to run the code, you'll need to run `npm install` within the custom package file as well in order to download the required dependencies. In the future, it would be good if running `npm install` in the root directory also ran `npm install` in the custom plugins.


adilparvez:
No longer needed.

PR #1494, opened 2020-03-20
D3fc series examples (part 2)

OliverForeman:
This builds on #1487 moving the D3FC series examples.

In this PR:
* Decorate Append
* Decorate
* Errorbar
* Grouped
* Heatmap

PR #1487, opened 2020-03-17
D3fc series examples

OliverForeman:
The start of moving all examples to the examples directory.

In this PR:
* Area
* Autobandwidth
* Bar
* Box Plot
* Candlestick

Each example is being moved to 3 directories (one for each rendering method), for example `candlestick` examples are now under `series-svg-candlestick`, `series-canvas-candlestick`, and `series-webgl-candlestick`.

Every new directory contains the example files, a README, and a test file that also generates a screenshot for the example. In some cases a screenshot is also generated for the `d3fc-series/screenshots` directory for use on the website.

PR #1495, opened 2020-03-20
D3fc series examples (part 3)

OliverForeman:
This builds on #1487 and #1494 moving the D3FC series examples.

In this PR:
* Line
* Multi
* OHLC
* Point size type
* Point
* Repeat
* Stacked


chrisprice (changes requested):
Looks like they're not transparent backgrounds

OliverForeman:
> Looks like they're not transparent backgrounds

It seems the majority of the series examples screenshots don't have transparent backgrounds. As this change (and the handling of context) needs to be rolled out across all examples this will be handled in a separate PR.

chrisprice (approved):

PR #1501, opened 2020-03-24
D3fc annotation examples

OliverForeman:
Moves the examples from the annotation directory to the examples directory


examples/annotation-canvas-crosshair/index.js
```diff
@@ -0,0 +1,49 @@
...
+        context.fillStyle = 'rgba(0, 0, 0, 0)';
+        context.scale(95, 95);
+    });
+
+const canvas = container.querySelector('canvas');
+
+d3.select(container)
+    .on('draw', () => {
+        crosshair(data);
+        canvas.onmousemove = ({ offsetX, offsetY }) => {
```
chrisprice: Can we be consistent in how we add event handlers? So that within a given example we only use one style i.e. `.onx =`/`addEventListener`/`.on(`.

chrisprice (changes requested):

examples/annotation-canvas-crosshair/index.js
```diff
@@ -0,0 +1,50 @@
...
+        context.fillStyle = 'rgba(0, 0, 0, 0)';
+        context.scale(95, 95);
+    });
+
+const canvas = container.querySelector('canvas');
+
+d3.select(container)
+    .on('draw', () => {
+        crosshair(data);
+        d3.select(canvas).on('mousemove', () => {
```
chrisprice: The canvas should be the same size as the container so there's no need to reference it. I think that would allow you to move this handler to be a peer of 'draw', 'measure'.

chrisprice (approved):

PR #1503, opened 2020-03-24
Webgl point star

adilparvez:
image


adilparvez:
PR looks messed up for some reason, will rebase onto master.

chrisprice (changes requested):
Seems weird that any package-locks/screenshots should be updated

chrisprice:
As discussed I really like the idea of distance functions. Please could you raise an issue for moving all of the symbol rendering over to using it? It would be great if you could detail the changes we'd likely want to make to stroke/fill while we're at it.

chrisprice (approved):

PR #1510, opened 2020-03-27
D3fc axis examples

OliverForeman:
Moves the examples from `d3fc-axis` to the `examples` directory


examples/axis-center/index.js
```diff
@@ -0,0 +1,56 @@
...
+
+const bandScale = d3.scaleBand().domain(domain);
+
+const bandAxis = fc.axisOrdinalBottom(bandScale).tickPadding(5);
+
+d3.select(bandContainer)
+    .on('draw', () => {
+        d3.select(bandContainer)
+            .select('svg')
+            .append('g')
```
chrisprice: The `g` elements are redundant in this case. I think they'd only be required for these examples if there are multiple components being rendered.

examples/axis-color/index.js
```diff
@@ -0,0 +1,28 @@
...
+        .select('text')
+        .style('fill', d => (d >= 100 ? 'red' : 'black'));
+});
+
+d3.select(container)
+    .on('draw', () => {
+        d3.select(container)
+            .select('svg')
+            .append('g')
+            .attr('transform', 'translate(0, 10)')
```
chrisprice: I'm not sure what this transform is for? I can understand if we're positioning a few on screen or shifting stuff around to be visible.

examples/axis-label-offset/index.html
```diff
@@ -44,15 +46,16 @@
             grid-row: 3;
         }
     </style>
-
 </head>
+
 <body>
     <div class="container">
         <div id="top"><svg data-orient="horizontal"></svg></div>
```
chrisprice: Could these all be made `d3fc-svg`s?

examples/axis-label-offset/index.js
```diff
@@ -0,0 +1,65 @@
...
+
+    renderAxis(
+        scale => fc.axisLabelOffset(fc.axisBottom(scale)),
+        '#bottom svg',
+        'translate(0, 0)'
+    );
+    renderAxis(
+        scale => fc.axisLabelOffset(fc.axisTop(scale)),
+        '#top svg',
+        'translate(0, 80)'
```
chrisprice: These transforms should be handled like they are in cartesian - https://github.com/d3fc/d3fc/blob/master/packages/d3fc-chart/src/cartesian.js#L126

examples/axis-offset/index.js
```diff
@@ -0,0 +1,36 @@
...
+        .select('text')
+        .attr('transform', (_, i) => `translate(0, ${i % 2 === 0 ? 20 : 10})`);
+});
+
+d3.select(container)
+    .on('draw', () => {
+        d3.select(container)
+            .select('svg')
+            .append('g')
+            .attr('transform', 'translate(0, 10)')
```
chrisprice: Again not sure this is necessary

examples/axis-label-offset/index.html
chrisprice: I think we should be able to get rid of the `div`s if we do that as well.

examples/axis-rotate-auto/index.html
```diff
@@ -55,27 +56,25 @@
...
-                <svg id="bottomAuto" class="bottom"></svg>
-            </div>
-            <div class="ordinal-group">
-                <div class="label">Fixed rotation</div>
-                <svg id="topFixed" class="top"></svg>
-                <svg id="leftFixed" class="left"></svg>
-                <svg id="rightFixed" class="right"></svg>
-                <svg id="bottomFixed" class="bottom"></svg>
-            </div>
+<body>
```
chrisprice, line 54: As before `div>svg` -> `d3fc-svg`

examples/axis-rotate-auto/index.js
```diff
@@ -0,0 +1,92 @@
+const width = 400;
```
chrisprice: We shouldn't need these if we do the above

chrisprice (changes requested):

examples/axis-rotate-auto/index.js
```diff
@@ -0,0 +1,82 @@
...
+        '#leftFixed'
+    );
+    renderAxis(
+        scale => fc.axisLabelRotate(fc.axisOrdinalRight(scale)).labelRotate(30),
+        '#rightFixed'
+    );
+};
+
+draw();
+window.addEventListener('resize', () => draw());
```
chrisprice: Just spotted this line. If you wrap all of the `d3fc-svg`s in a `d3fc-group` (i.e. 1 x `d3fc-group` with n x `d3fc-svg` children) with an `auto-resize` attribute, it should handle the redrawing for you so you can remove the `draw` function and just make the code inline.

chrisprice (changes requested):

chrisprice (approved):

PR #1513, opened 2020-03-30
wip: Streaming attribute

chrisprice:
I'm not sure if this should be added to the project or just added as an example. There are a few tradeoffs to be aware of -

* Retaining the data outside of the GPU to allow recovering from context lost versus just streaming it straight in. Recovery could happen by re-requesting a snapshot from the server instead. This implementation does neither.
* Transferring the data to the GPU immediately rather than enqueuing for the next render. This is hard because of our internal API.
* As we're enqueuing for the next render and `attributeProjector` attempts to reuse buffers, the consumer can not reuse the `data` until they can be sure the data has been enqueued.

Additionally, in implementing this I noticed that we should profile the difference between using `buferData` on each render versus initialising with `bufferData` and then calling `bufferSubData` for loading data.

### DIFF
--- a/examples/series-webgl-point/index.js
+++ b/examples/series-webgl-point/index.js
@@ -1,4 +1,4 @@
-const data = fc.randomGeometricBrownianMotion().steps(1e4)(1);
+const data = fc.randomGeometricBrownianMotion().steps(1e5)(1);
 
 const extent = fc.extentLinear();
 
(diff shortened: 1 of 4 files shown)


adilparvez (approved):

chrisprice:
Whilst an approach similar to this one is likely to lead to the most highly performing code possible it comes at the cost of complicating the API. #1560 offers an alternative which whilst not as optimal offers adequate performance and is a much better fit with the existing API.

PR #1514, opened 2020-03-30
refactor: move webgl examples

adilparvez:
Move and update examples from the d3fc-webgl subpackage to the root examples folder.


examples/webgl-overlay/__tests__/index.js
```diff
@@ -0,0 +1,25 @@
+const { join } = require('path');
+
+it('should have consistent performance', async () => {
```
chrisprice: I'm not sure this is doing what you expect it to be doing.

If it was just copy/pasted, I'm happy to see it deleted. If it was intentional, a few things will need fixing up -

* We need the code which reads the rendered canvas pixel data back as it is the only way to accurately measure the performance. 
* We also need to make sure the clicking that's been done is meaningful to the test.

examples/webgl-overlay/index.js
```diff
@@ -8,33 +8,20 @@ d3.json('star-data.json', data => {
...
-    let i = 0;
-    data.forEach(d => {
-        colors[i++] = d.color[0];
-        colors[i++] = d.color[1];
-        colors[i++] = d.color[2];
-        colors[i++] = d.color[3];
-    });
-    const colorBuilder = fc.attributeBuilder(colors);
-
     const getWebglSeries = () =>
```
chrisprice: Can we remove the `get**()` functions? I think all of the code can be inlined.

I appreciate they pre-date your involvement but can we tidy it up while we're moving it over.

examples/webgl-overlay/index.js
```diff
@@ -8,33 +8,20 @@ d3.json('star-data.json', data => {
...
-                    .fragmentShader()
-                    .appendHeader(fc.fragmentShaderSnippets.multiColor.header)
-                    .appendBody(fc.fragmentShaderSnippets.multiColor.body);
-
-                program.buffers().attribute('aColor', colorBuilder);
+                fc.webglFillColor()
+                    .value(d => d.color)
+                    .data(data)(program);
 
                 const context = program.context();
```
chrisprice: What do these lines do?

examples/webgl-performance/README.md
```diff
@@ -0,0 +1,3 @@
+# WebGL Performance
```
chrisprice: Can we drop this example?

examples/webgl-squares/index.js
```diff
@@ -13,29 +13,11 @@ y.domain([0, 1]);
...
 
-const crossValue = d => d.x;
-const mainValue = d => d.y;
-
 const getWebglSeries = () =>
-    fc
-        .seriesWebglPoint()
-        .crossValue(crossValue)
-        .mainValue(mainValue)
+    fc.seriesWebglPoint()
```
chrisprice: Sorry same for this one, not sure it adds anything on top of the starts example

examples/webgl-overlay/README.md
```diff
@@ -0,0 +1,6 @@
+# WebGL Overlay
```
chrisprice: Please can we rename this to webgl-with-svg-overlay

chrisprice (changes requested):

chrisprice (approved):

PR #1522, opened 2020-04-01
fix: stale divisor values if using non-instanced series

### LINKED ISSUE #1521: Bug when using points series w/ any other series
This commit introduced a weird bug and I have no idea where it comes from. When I do a plot with seriesWebglLine and seriesWebglPoint, the points' main value all take the main value of the first point. The cross value is correct, but they all get the same main value which is the one of the first point. Once I remove the seriesWeblLine, the seriesWebglPoint plots normally. I think the WebglPoint's shader is reusing the elementIndices buffer of the seriesWebglLine ? Just a guess, I don't know webgl that much

https://github.com/d3fc/d3fc/commit/a53813f561644e1f57e9e4340655072160282a54#commitcomment-38199777

@melaurent

chrisprice:
Fixes #1521. 

This only fixes the immediate bug. I think it's worth continuing with the larger refactor described in #1521 but sharing this version as a workaround.

### DIFF
--- a/packages/d3fc-webgl/README.md
+++ b/packages/d3fc-webgl/README.md
@@ -599,6 +599,14 @@ If *offset* is specified, sets the offset property and returns this attribute bu
 
 The offset property is used to specify the offset in bytes of the first value in the vertex attribute array. If set, the offset must be a multiple of the byte length of [type](#base-attribute).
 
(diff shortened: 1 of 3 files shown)


chrisprice:
Superseded by #1523

PR #1524, opened 2020-04-02
refactor: move d3fc-element examples

adilparvez:
Moves the examples from the d3fc-element subpackage to the root level examples directory.


examples/element-dual-axes/README.md
```diff
@@ -0,0 +1 @@
+# Element Dual Axes
```
chrisprice: This example seems to be a nice aggregation of the other examples into a useful chunk of functionality rather than the bits and pieces in the others. Could we rename it, drop the others and just keep this one?

# Element Bespoke Chart Layout

This examples shows how the d3fc-element components can be used directly to create completely bespoke chart layouts. In this case a flexbox layout has been used but any CSS layout technique could be used e.g. flow, grid, columns, etc..

PR #1541, opened 2020-04-14
Examples refactor

adilparvez:


adilparvez:
Work cleaned up on `examples-rebased` branch, this was kept just in case.

PR #1558, opened 2020-04-29
A couple of small fixes

ColinEberhardt:
The stacked bar example looks better with auto-bandwidth, also a d3 band scale is more appropriate


chrisprice (approved):
Was going to ask you to update the screenshot then realised we don't have any instructions for doing that...

PR #1559, opened 2020-04-30
Fix/improve examples

### LINKED ISSUE #1549: Template for website examples
It would be good to have more of a blocks or blockbuilder style interface on top of our examples. To ease maintenance this should be entirely client-side and require minimal build-time code. This should be relatively easy as the examples all follow the same file structure.

As it stands, without the README content the lack of context is causing confusion #1531 .

### LINKED ISSUE #1554: Examples unstable for snapshots
Nearly all of the examples use randomFinancial which is [initialised with `new Date()`](https://github.com/d3fc/d3fc/blob/master/packages/d3fc-random-data/src/financial.js#L8). This makes the image snapshots unstable if they're not run against a very recent baseline.

chrisprice:
Fixes #1549.
Fixes #1554.

### DIFF
--- a/examples/annotation-canvas-band/index.html
+++ b/examples/annotation-canvas-band/index.html
@@ -10,16 +10,6 @@
     <script src="../index.js"></script>
     <link rel="stylesheet" href="../index.css">
     <link rel="icon" type="image/png" href="data:image/png;base64,iVBORw0KGgo=">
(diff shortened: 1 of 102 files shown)


ColinEberhardt (approved):
Looks very good 👍

PR #1560, opened 2020-04-30
feat: optimise seriesWebglRepeat performance

chrisprice:
The seriesWebglRepeat.series property can now be specified
as a function that returns a series instance rather than a series
instance. This allows the repeat series to allocate a series instance
per data series, preventing the unnecessary cache-evictions which can
occur if only one series instance is used.

### DIFF
--- a/packages/d3fc-series/src/webgl/repeat.js
+++ b/packages/d3fc-series/src/webgl/repeat.js
@@ -1,0 +1,52 @@
+import { rebindAll, exclude } from '@d3fc/d3fc-rebind';
+import multiSeries from '../canvas/multi';
+import line from './line';
+
+export default () => {
+
+    let orient = 'vertical';
+    let series = () => line();
+    const multi = multiSeries();
+    let seriesCache = [];
+
+    const repeat = (data) => {
+        if (orient === 'vertical') {
+            const previousSeriesCache = seriesCache;
+            seriesCache = data[0].map((d, i) => i < previousSeriesCache.length ? previousSeriesCache[i] : series());
+            multi.series(seriesCache)
+              .mapping((data, index) => data.map(d => d[index]));
+        } else {
+            const previousSeriesCache = seriesCache;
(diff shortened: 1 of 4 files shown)


packages/d3fc-series/src/webgl/repeat.js
```diff
@@ -0,0 +1,52 @@
+import {rebindAll, exclude} from '@d3fc/d3fc-rebind';
```
mattstobbs: Nitpick: Missing spaces after and before curly brackets

packages/d3fc-series/src/webgl/repeat.js
```diff
@@ -0,0 +1,52 @@
...
+    const multi = multiSeries();
+    let seriesCache = [];
+
+    const repeat = (data) => {
+        if (orient === 'vertical') {
+            const previousSeriesCache = seriesCache;
+            seriesCache = data[0].map((d, i) => i < previousSeriesCache.length ? previousSeriesCache[i] : series());
+            multi.series(seriesCache)
+              .mapping((data, index) => data.map(d => d[index]));
+        } else {
```
mattstobbs, line 18: I'm assuming this is how it's implemented in the other series, but this means the default is horizontal (i.e. if you set the orient to anything other than `'vertical'`, it will be horizontal). This behaviour seems to conflict with the fact that if no orient is set, the series uses `'vertical'` as the default. Would it be better to have `else if (orient === 'horizontal') {` and then an else statement that reports to the user than a unacceptable orient value has been used?

chrisprice: I completely agree but this seems to be the pattern across the whole code base so it would be a breaking change. I'll raise an issue for it #1562

packages/d3fc-series/src/webgl/repeat.js
chrisprice: Not sure why the linter didn't pick this up

mattstobbs (approved):
LGTM

PR #1568, opened 2020-05-06
docs: adding a small multiples example

ColinEberhardt:
starting to migrate some of my old examples


chrisprice (changes requested):
Looks like some lining errors as well.

examples/small-multiples/index.js
```diff
@@ -0,0 +1,87 @@
...
+    .value(d => d);
+
+// bring all these renderers together into one using the multi-series. The area/point/line
+// are mapped to the 'values' returned by d3.nest, and the line annotation maps to the average
+const multi = fc.seriesSvgMulti()
+    .series([area, line, point, average])
+    .mapping((data, index, series) => {
+        switch (series[index]) {
+        case average:
+            return [d3.mean(data.values.map(function(d) { return d.Sales; }))];
```
chrisprice: Second argument to mean is an optional accessor. Probably also worth an arrow function

ColinEberhardt: good point - fixed

chrisprice (approved):

PR #1569, opened 2020-05-06
Examples bar chart race

ColinEberhardt:
I think [this example](https://bl.ocks.org/ColinEberhardt/2d9168600094e50ac379838f7cebc183) is a good illustration of the transitions support - although as it has a timer, it upsets the screenshot code :-)


chrisprice:
Looks cool! Should be possible to get screenshots working without too much trouble. Just make sure to do an initial render outside of the setInterval callback do there's something to screenshot and wait for the chart element to appear before taking the screenshot (see any of the other examples that load data in using fetch).

ColinEberhardt:
> Just make sure to do an initial render outside of the setInterval callback do there's something to screenshot and wait for the chart element to appear before taking the screenshot

Done and screenshotted

chrisprice (approved):

PR #1572, opened 2020-05-12
docs: added label layout example

ColinEberhardt:


chrisprice (approved):
Something weird with the build, might need a kick

PR #1573, opened 2020-05-13
docs: added boxplot example

ColinEberhardt:


chrisprice (approved):
Same thing with the build I think

PR #1574, opened 2020-05-14
fix: label rotation and offset needs to support the logic around tick value and formatting

ColinEberhardt:
The current implementation simply requests the ticks from the scale, as a result it doesn't work if the user adds a formatter or sets a specific tick array.

This PR ensures that the rotation / offset code uses exactly the same tick / formatter logic as the axis itself.


packages/d3fc-axis/src/axisTickUtils.js
```diff
@@ -0,0 +1,25 @@
...
+// properties: https://github.com/d3/d3-axis#axis_ticks 
+
+const identity = d => d;
+
+const tryApply = (scale, fn, args, defaultVal) =>
+    scale[fn] ? scale[fn].apply(scale, args) : defaultVal;
+
+const ticksArrayForAxis = axis =>
+    axis.tickValues() == null
+        ? tryApply(
```
chrisprice: Perfect use case for https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator if you fancy doing the build dance?

packages/d3fc-axis/src/axisTickUtils.js
```diff
@@ -0,0 +1,25 @@
...
+              axis.scale(),
+              'ticks',
+              axis.tickArguments(),
+              axis.scale().domain()
+          )
+        : axis.tickValues();
+
+const tickFormatterForAxis = axis =>
+    axis.tickFormat() == null
```
chrisprice: As above

packages/d3fc-axis/src/axisTickUtils.js
ColinEberhardt: build dance complete 👍

chrisprice (approved):

PR #1576, opened 2020-05-22
Handle device pixel ratio within cartesian chart

### LINKED ISSUE #1564: Bandwidth differs between SVG and WebGL series
Here's an SVG series with a 15px bandwidth:

image

And here's a WebGL series with the same:

image

Also, the high / low 'wick' looks like it has a sub-pixel width

ColinEberhardt:
also fixes #1564


ColinEberhardt:
ah - linting, need to get VSCode set up properly for this project again.

ColinEberhardt:
replaced with #1596

PR #1578, opened 2020-05-31
fix: update SVG gridline stroke to be set on enter

### LINKED ISSUE #1577: Could SVG gridline annotation stroke be set on enter not update?
The `annotationSvgGridline` component sets the `stroke` attribute of both the x and y lines on update rather than enter:

https://github.com/d3fc/d3fc/blob/f8ddaf6a5f0b58aa0460c5d045364d3f32e17a53/packages/d3fc-annotation/src/svg/gridline.js#L48-L53

https://github.com/d3fc/d3fc/blob/f8ddaf6a5f0b58aa0460c5d045364d3f32e17a53/packages/d3fc-annotation/src/svg/gridline.js#L74-L79

When using the decorate pattern to set the stroke color of the gridlines, the stroke must be set on update rather than enter (I've created [an example on CodeSandbox](https://codesandbox.io/s/pensive-mcclintock-k0g4f?file=/src/index.js) to illustrate this). Given that the stroke set by the component is contant (`#bbb`) it seems unecessary to do this on update; could this be set on enter?

I think this would also be consistent with the line annotation, where the `stroke` attribute is set on enter:

https://github.com/d3fc/d3fc/blob/250e3455624ffbce3ed7e381af6d8946906afafd/packages/d3fc-annotation/src/svg/line.js#L50-L52

jleft:
Resolves #1577

### DIFF
--- a/packages/d3fc-annotation/src/svg/gridline.js
+++ b/packages/d3fc-annotation/src/svg/gridline.js
@@ -43,14 +43,14 @@ export default () => {
                 .attr('x1', xScaleOld)
                 .attr('x2', xScaleOld)
                 .attr('y1', yScale.range()[0])
-                .attr('y2', yScale.range()[1]);
+                .attr('y2', yScale.range()[1])
+                .attr('stroke', '#bbb');
 
             xLines
                 .attr('x1', xScale)
                 .attr('x2', xScale)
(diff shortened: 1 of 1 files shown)


chrisprice (approved):

ColinEberhardt:
Thanks @jleft

PR #1580, opened 2020-06-01
docs: render axes in correct containers in axis transition example

jleft:
`d3OrdinalAxis` was being rendered within the `fcLinearAxisJoin` group and `fcLinearAxis` was being rendered within the `d3OrdinalAxisJoin` group. This change swaps them around so that each axis is rendered within its corresponding group.


chrisprice (approved):
Can I just check how you generated the screenshot?

jleft:
> Can I just check how you generated the screenshot?

Sure - I ran the commands in the [Regenerating screenshots](https://github.com/d3fc/d3fc/blob/96fb2dd73a8300e1a3953902543345323c5bed05/examples/README.md#regenerating-screenshots) section of the Examples README:

> Start up an HTTP server e.g. -
> 
> ```bash
> npx http-server
> ```
> 
> Then run the following -
> 
> ```bash
> npm run bundle
> npm run examples -- --testNamePattern image --updateSnapshot
> npm run examples -- --testNamePattern image
> ```

I wasn't enirely sure if this was the right thing to do. I did notice that many screenshot had been modified, but I couldn’t notice any obvious visual differences except for the screenshot that is committed for this specific example.

chrisprice:
Sorry still need to double check this when I'm in front of a computer

chrisprice:
Looking through the screenshots there's still differences creeping in. I think we should merge this as is and then fixup the images in #1590.

PR #1585, opened 2020-06-18
docs: fix chart labels in cartesian chart ordinal example

jleft:
Update chart labels to correspond with data.


ColinEberhardt (approved):
well spotted

PR #1586, opened 2020-06-25
chore: Move from Jasmine to Jest

### LINKED ISSUE #1296: Move to jest
#1295 has reminded me how old some of our tooling is. It would be nice to get the test runner upgraded as well.

jleft:
Resolves #1296.

- Uses pre-existing top-level Jest dependencies
- Removes per pacakage `test` script per package
- One top level test script that runs all of the tests across all packages, excluding examples
- Removes `jasmine` dependency
- Does not collect code coverage (but could be added)
- Uses `jsdom` [`testEnvironment`](https://jestjs.io/docs/en/configuration#testenvironment-string); explicit `jsdom` usage removed for all specs except `bundleSpec`s
  - `bundleSpec`s are somewhat inconsistent, e.g. some packages have them and some don't, and they duplicate a lot of code; I've tried to keep changes to a minimum here, but this could be an area for future refactoring?
- Retained pre-existing test directory and filename patterns
- Moved custom matches to top-level
- Migrated from Jasmine to Jest expectations, spies etc. where appropriate


package.json
```diff
@@ -7,7 +7,7 @@
     "publish": "lerna publish --conventional-commits --message \"chore: publish\" --yes",
     "bundle": "lerna run bundle --stream --concurrency 1",
     "bundle-min": "lerna run bundle --stream --concurrency 1 -- -- --configEnv=prod",
-    "test": "lerna run test --stream --concurrency 1",
+    "test": "npx jest -c scripts/jest/jest.config.js",
```
chrisprice: I don't think this needs the `npx` on the start and can we use the long version of the jest flag (i.e. `--config`)

package.json
```diff
@@ -58,7 +58,6 @@
     "eslint-plugin-promise": "^4.2.1",
     "eslint-plugin-standard": "^4.0.1",
     "glsl-transpiler": "^1.8.5",
-    "jasmine": "^2.4.1",
     "jest": "^24.9.0",
     "jest-image-snapshot": "^2.12.0",
     "jest-puppeteer": "^4.4.0",
```
chrisprice, line 63: Are there any other dependencies we could remove e.g. jsdom? Happy to see this raised as a separate issue

chrisprice (commented):
Just a couple of minor comments otherwise looks awesome to me. Especially like this bit -

image

package.json
jleft: Good spot, thanks - that's fixed now. I've also updated to use `--config` instead of `-c`.

package.json
jleft: Not that I'm aware of. `jsdom` is still used in the various `bundleSpec` tests that most packages have, e.g. in d3fc-annotation:

https://github.com/d3fc/d3fc/blob/bb3958f2df147e4614a4d8289a9819d8e8e7145c/packages/d3fc-annotation/test/bundleSpec.js#L1

I've opened https://github.com/d3fc/d3fc/issues/1591 which lists a few potential areas for improvement for the bundle specs, including potentially using Jest's `jsdom` [`testEnvironment`](https://jestjs.io/docs/en/configuration#testenvironment-string) and removing the explicit `jsdom` dependency.

chrisprice (approved):

chrisprice:
Thanks @jleft

PR #1595, opened 2020-07-23
docs: add axis annotation example

chrisprice:


examples/axis-annotation/index.js
```diff
@@ -0,0 +1,61 @@
...
+const annotation = new Annotation(45, 0.01);
+
+const scale = d3.scaleLinear().domain([0, 100]);
+
+const axis = fc
+    .axisBottom(scale)
+    .tickValues([...scale.ticks(), annotation])
+    .decorate(g => {
+        g.filter(d => d === annotation)
+            .select('path')
```
ColinEberhardt, line 38: should this use `enter()` to ensure that these style / attributes are only set when a new element is created?

examples/axis-annotation/index.js
```diff
@@ -0,0 +1,61 @@
+class Annotation {
+    constructor(value, velocity) {
+        this.value = value;
+        this.velocity = velocity;
+    }
+
+    animate() {
+        this.velocity += (Math.random() - 0.5) * 0.01;
+        this.value += this.velocity;
+        if (this.value < scale.domain()[0]) {
```
ColinEberhardt: <minor/>

I find indexed values returned by accessors really hard to read `scale.domain()[0]`, perhaps pull out the scale range as a variable?
```
const range = [0, 100];
...
this.value = range[0] + (range[0] - this.value)
...
const scale = d3.scaleLinear().domain(range);
```

ColinEberhardt (approved):
Useful example, a couple of minor comments, I'll leave to your discretion.

examples/axis-annotation/index.js
chrisprice: I could for the styles but the axis code sets the `d` attribute on each update so I figured I'd just lump them together for simplicity

PR #1627, opened 2020-09-16
D3 v6 

### LINKED ISSUE #1606: latest version of d3fc is incompatible with D3v6
With new d3 version (6.0.0) release, examples under https://d3fc.io/examples/ stopped working as didn't specify the version of d3 to use from unpkg

chrisprice:
Fixes #1606


ColinEberhardt (approved):
I've spent a good few hours using the build - haven't hit any issues yet.

felixvictor:
- packages/d3fc-label-layout/src/label.js
- packages/d3fc-label-layout/src/textLabel.js

still use the old argument list for `selection.each((data, index, group)` instead of `selection.each((event, data)`.

chrisprice:
@felixvictor thanks for the report but I don't believe the signature for selection.each has changed with V6 i.e. I think those files are correct as they are. Please raise an issue with an example of any broken behaviour you're seeing though and we can take a look.

felixvictor:
@chrisprice Yes, you are right. It works as expected.

Issue #1464, opened 2020-02-16
Ticks are disappeared if discontinuous scale


tamtakoe:
This is impossible to use discontinuous scale for stock charts because of the ticks issue
![bug (1)](https://user-images.githubusercontent.com/2244675/74605162-83a3e080-50d6-11ea-8fe5-47936fb1c1e0.gif)

I tried this way, but it doesn't help a lot. But make application very slow
```
const dayTicks = d3Time.timeHour
    .filter(d => d.getHours() === 10);

$axis.call(d3.axisBottom(options.x).ticks(dayTicks).tickFormat(d3.timeFormat('%b %d')))
```

It will be good to fix the issue and it will be also good if nice parameter will add ticks at the borders of discontinuous periods :-)

tamtakoe:
I solved the issue - dropped  discontinuous scale (and all d3fs project) and use native Linear scale instead. Just I generate correct tickValues and tickFormat myself. It works much faster

Issue #1167, opened 2018-08-13
Axis with skipped weekends doesn't have any ticks


ColinEberhardt:
When updating teh examples to v14 as per https://github.com/d3fc/d3fc/issues/1109 - the following example shows a regression:

http://blockbuilder.org/ColinEberhardt/0a5cc7ca6c256dcd6ef1e2e7ffa468d4

When weekends are skipped, no tick marks are rendered

chrisprice:
So this [commit](https://github.com/d3fc/d3fc/commit/d8062b0e56e4fa3036ed3819bb94228d7abe60dd#diff-3953acd00e77a16487a3b95613dc3d2c) explicitly adds this behaviour. I'm not sure what the behaviour should be?

chrisprice:
After a bit of research, I think it makes sense to push the decision on to the user. I can think of a number of different ways in which a user would want the replacement ticks to be selected (e.g. consistent pixel spacing, consistent domain spacing, domain specific spacing, etc.) so I don't think there's a sensible default.

A user can choose which ticks are displayed with something like -

```js
// ...

const xTickFilter = d3.timeDay.filter((d) => d.getDay() === 1);

// Create the gridlines and series
const gridlines = fc.annotationSvgGridline()
	.xTicks(xTickFilter);

// ...

// adapt the d3 time scale in a discontinuous scale that skips weekends
const skipWeekendScale = fc.scaleDiscontinuous(d3.scaleTime())
  .discontinuityProvider(fc.discontinuitySkipWeekends());

function renderChart() {
  // create a chart
  const chart = fc.chartSvgCartesian(
      checkbox.checked ? skipWeekendScale : d3.scaleTime(),
      d3.scaleLinear()
    )
    .xDomain(xExtent(data))
    .xTicks(xTickFilter)
    .yDomain(yExtent(data))
    .plotArea(multi);

  // ...
}

// ...
```

chrisprice:
@ColinEberhardt as I can't edit the block, can you update with the above code and add this to the end of the description -

> N.B. By default, the automatic tick positioning of the scale will try to place the ticks on Sunday's at midnight. Unfortunately this they will be filtered out by the discontinuous scale because the ticks fall within the discontinuities. To work around this you can manually specify the tick interval using d3-time -

```
const xTickFilter = d3.timeDay.filter((d) => d.getDay() === 1);
gridlines.xTicks(xTickFilter);
chart.xTicks(xTickFilter);
```

[...]

PR #1631, opened 2020-09-24
fix: ticks that fall within a discontinuity are no longer removed

### LINKED ISSUE #1601: Discontinuities can result in an axis without any ticks
xScale = fc.scaleDiscontinuous(d3.scaleTime())
            .discontinuityProvider(fc.discontinuitySkipWeekends());

When I change the scale(zoom), I get a bug. Labels and text on the x-scale disappear. Data is an array with a step of one day. The data is sorted. UTC format.

 How to fix?
example

ColinEberhardt:
fixes #1601 #1167 #1464

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -9,13 +9,17 @@
   "env": {
     "es6": true
   },
+  "globals": {
+    "console": true
+  },
   "parserOptions": {
     "ecmaVersion": 2020,
     "sourceType": "module"
   },
   "rules": {
     "no-unused-vars": ["error", { "args": "none" }],
-    "semi": ["error", "always"]
+    "semi": ["error", "always"],
+    "no-console": ["error", { "allow": ["warn", "info"] }]
   },
   "overrides": [
     // Additional globals for tests
--- a/packages/d3fc-discontinuous-scale/src/tickFilter.js
+++ b/packages/d3fc-discontinuous-scale/src/tickFilter.js
@@ -1,11 +1,12 @@
 export default function tickFilter(ticks, discontinuityProvider) {
-    const discontinuousTicks = [];
-    for (const tick of ticks) {
-        const up = discontinuityProvider.clampUp(tick);
-        const down = discontinuityProvider.clampDown(tick);
-        if (up === down) {
-            discontinuousTicks.push(up);
-        }
+    const discontinuousTicks = ticks.map(discontinuityProvider.clampUp);
+    if (
+        discontinuousTicks.length !==
+        new Set(discontinuousTicks.map(d => d?.valueOf())).size
+    ) {
+        console.warn(
+            'There are multiple ticks that fall within a discontinuity, which has led to them being rendered on top of each other. Consider using scale.ticks to explicitly specify the ticks for the scale.'
+        );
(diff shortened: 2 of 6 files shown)


packages/d3fc-discontinuous-scale/src/tickFilter.js
```diff
@@ -1,11 +1,19 @@
+import { pairs } from 'd3-array';
+
+function areEqual([a, b]) {
+    if (a.valueOf) {
```
chrisprice: Everything in JS has a `valueOf` i.e. [`Object.prototype.valueOf`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/valueOf) other than `null`/`undefined`. However, they'd cause a `TypeError` here so I guess that's not what you're aiming to catch.

How about a `values.map(d => d?.valueOf())`?

.eslintrc.json
```diff
@@ -31,7 +31,8 @@
     {
       "files": [
         "./packages/d3fc-chart/src/**/*.js",
-        "./packages/d3fc-element/src/**/*.js"
+        "./packages/d3fc-element/src/**/*.js",
+        "./packages/d3fc-discontinuous-scale/src/**/*.js"
```
chrisprice: Is this just to support `console.warn`? Maybe it's worth us adding a utility package to wrap this up so we don't need to add every package to this list?

packages/d3fc-discontinuous-scale/src/tickFilter.js
```diff
@@ -1,11 +1,19 @@
...
 export default function tickFilter(ticks, discontinuityProvider) {
-    const discontinuousTicks = [];
-    for (const tick of ticks) {
-        const up = discontinuityProvider.clampUp(tick);
-        const down = discontinuityProvider.clampDown(tick);
-        if (up === down) {
-            discontinuousTicks.push(up);
-        }
+    const discontinuousTicks = ticks.map(discontinuityProvider.clampUp);
+    if (pairs(discontinuousTicks).some(areEqual)) {
```
chrisprice: How about `discontinuousTicks.length !== new Set(discontinousTicks.map(d => d?.valueOf())).length`?

(You can drop the new dependency in that case)

chrisprice (changes requested):

.eslintrc.json
ColinEberhardt: Is your main concern that by moving this package into this particular override section that it now permits all browser globals? How about we just enable console as a global variable across all packages?

https://github.com/d3fc/d3fc/pull/1631/files#diff-df39304d828831c44a2b9f38cd45289cR12

chrisprice: It was and that definitely solves it in a nicer way. However, we now risk accidently leaving console logs in the code. I wonder if we should just excuse the individual log lines?

ColinEberhardt: We can prevent `console.log` explicitly:

https://github.com/d3fc/d3fc/pull/1631/files#diff-df39304d828831c44a2b9f38cd45289cR22

chrisprice (approved):

Issue #1603, opened 2020-08-27
Is there an example of pan / drag / zoom on axes independently? 


artemiusgreat:
There is an example of how to do zoom on one axis and pan on another. 
https://stackoverflow.com/a/35445448/437393

Meanwhile, it's for D3 v3 and it uses `behavior` which doesn't seem to be available now, so I was wondering if there is an example of how to do it in the latest version?

ColinEberhardt:
The way that D3 brush (pan) and zoom works makes this a little tricky - they apply a transformation that is 'stashed' on the DOM node itself. As a result, applying multiple interactions causes a conflict.

I created a solution to this in response to a StackOverflow question a little while back, which might be what you need: https://stackoverflow.com/a/61140796/249933

artemiusgreat:
Thank you

PR #1633, opened 2020-09-26
feat: added an interactive chart component

ColinEberhardt:
An interactive chart - see the code and discussions here:

https://stackoverflow.com/questions/61071276/d3-synchronizing-2-separate-zoom-behaviors
https://observablehq.com/@d3/x-y-zoom

There are currently a couple of issues:

1. Needs a rework to allow the component to render multiple charts (e.g. small multiples) - a big issue here is that the chart mutates the component properties (x/y domain)
2. There is some nasty code that determines if the user has changed the domain.

I think this is a potentially quite useful component if 1&2 can be resolved, or some compromise found!

see #1603


ColinEberhardt:
I've also included a different take on this, creating a chart decorator component that adds this functionality:

https://github.com/d3fc/d3fc/pull/1633/files#diff-6150ad9a09807415321cdfb87b9fbce9R15

chrisprice:
Do we still want to go with adding d3fc zoom to Cartesian chart?

ColinEberhardt:
@chrisprice I'm not sure it's needed now - the [diamonds example](https://github.com/d3fc/d3fc/blob/master/examples/diamonds/index.js) shows how easy it is to add zoom functionality to a chart.

PR #1634, opened 2020-10-09
feat: add d3fc-domain-zoom

chrisprice:
I'm open to a better name 😄


examples/chart-domain-zoom/index.js
```diff
@@ -0,0 +1,69 @@
...
+            .call(zoom, x, y);
+        sel.enter()
+            .selectAll('.x-axis')
+            .call(zoom, x, null);
+        sel.enter()
+            .selectAll('.y-axis')
+            .call(zoom, null, y);
+    });
+
+function clamp(
```
ColinEberhardt: Unfortunately this clamp doesn't work very well for me. When the chart first loads, if I perform a pan, the scale starts skewing. Likewise, if I zoom to the scale extent, the translation starts to drift.

examples/diamonds/index.js
```diff
@@ -10,11 +10,7 @@ d3.tsv('diamond-data.tsv', d => ({
     const yScale = d3.scaleLog().domain(yExtent(data));
     const yScaleCopy = yScale.copy();
 
-    const zoom = d3.zoom().on('zoom', event => {
-        xScale.domain(event.transform.rescaleX(xScaleCopy).domain());
-        yScale.domain(event.transform.rescaleY(yScaleCopy).domain());
-        render();
-    });
+    const zoom = fc.domainZoom().on('zoom', render);
```
ColinEberhardt: It's good to see how it simplifies this example.

packages/d3fc-domain-zoom/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+A [d3.zoom](https://github.com/d3/d3-zoom#zoom) component is used internally to bind the necessary event listeners. Therefore you can unbind the behaviour as follows:
+
+```javascript
+selection.on(".zoom", null);
+```
+
+<a name="domainZoom_on" href="#domainZoom_on">#</a> *domainZoom*.**on**(*typenames*[, *callback*])
+
+This component dispatches `zoom` events. There are no arguments are passed to the event handler. Typically this would be used to re-render the chart.
```
ColinEberhardt: There are no arguments ~are~ passed to the event handler

packages/d3fc-domain-zoom/src/domainZoom.js
```diff
@@ -0,0 +1,122 @@
...
+                    previousYScale: yScaleCopy,
+                    yScale,
+                    previousTransform: zoomIdentity
+                };
+            })
+            .call(zoomer);
+    };
+
+    rebind(instance, dispatcher, 'on');
+    rebind(
```
ColinEberhardt: are there fewer properties that are no re-bound? Is it easier to achieve this through exclusion?

packages/d3fc/package.json
```diff
@@ -29,7 +29,8 @@
     "@d3fc/d3fc-brush": "^3.0.3",
     "@d3fc/d3fc-chart": "^5.0.6",
     "@d3fc/d3fc-data-join": "^6.0.3",
-    "@d3fc/d3fc-discontinuous-scale": "^4.0.2",
+    "@d3fc/d3fc-discontinuous-scale": "^4.0.1",
+    "@d3fc/d3fc-domain-zoom": "^1.0.0",
```
ColinEberhardt: How about simply `d3fc-zoom`? It's unlikely we'll create a different zoom implementation.

ColinEberhardt (approved):

examples/chart-domain-zoom/index.js
chrisprice: Agreed it isn't ideal but as it's not part of the component, I think this can be iterated in a subsequent PR.

packages/d3fc-domain-zoom/src/domainZoom.js
chrisprice: I don't know why but being explicit feels a bit nicer to me. I'm happy to change it if you feel strongly about it?

PR #1697, opened 2021-04-19
feat: add d3fc-chart types

moxon6:
* Sets up the initial infrastructure for supporting d3fc Typescript types
* Add types to `@d3fc/d3fc-chart`


chrisprice (approved):
Cracking work. Did you try the js/tsconfig.json in the examples folder to replace the ```<reference...``` stuff?

moxon6:
> Cracking work. Did you try the js/tsconfig.json in the examples folder to replace the `<reference...` stuff?

I had a little play around with `jsconfig.json` but without success: from what I can see in docs and issues it seems like using `jsconfig.json` and `tsconfig.json` together is not an intended use case for a single workspace

ColinEberhardt:
Nice one, thank you @moxon6

PR #1710, opened 2021-07-29
Prevent registering components multiple times

matthijsgroen:
When using webcomponents that are using d3-fc the different
webcomponents don't know if another component has
already registered the d3fc components,
causing an error.

### DIFF
--- a/packages/d3fc-element/index.js
+++ b/packages/d3fc-element/index.js
@@ -13,9 +13,23 @@ if (
     );
 }
 
-const registerElement = (name, element) =>
-    customElements.get(name) || customElements.define(name, element);
+const alreadyRegistered = [];
+const registerElement = (name, element) => {
+    if (customElements.get(name)) {
+        alreadyRegistered.push(name);
+    } else {
+        customElements.define(name, element);
+    }
+};
 
 registerElement('d3fc-canvas', Canvas);
 registerElement('d3fc-group', Group);
 registerElement('d3fc-svg', Svg);
+
+if (alreadyRegistered.length > 0) {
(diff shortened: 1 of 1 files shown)


Haosvit:
👍👍

Haosvit (approved):

chrisprice:
Thanks for the PR 👍 

Can I check in what situation you're coming across this problem e.g. what bundler you're using, how you are importing the library, what framework you're using, etc.?

My concern is that with this change we're silently failing in the case that two different versions of the elements, with potentially breaking changes, are being included in the same document. I'm also concerned that even if the two versions are non-breaking or identical, there's still a hidden performance issue in that the library is being unnecessarily evaluated twice in the same context.

matthijsgroen:
> My concern is that with this change we're silently failing in the case that two different versions of the elements, with potentially breaking changes, are being included in the same document. I'm also concerned that even if the two versions are non-breaking or identical, there's still a hidden performance issue in that the library is being unnecessarily evaluated twice in the same context.

Yes that is a valid concern. I am working in an 'enterprise' organisation where multiple teams are creating webcomponents for a single UI. So for us we will move from an 'Exception and total breakage' towards a 'We need to be careful to have the versions synchronized' would be a big improvement.

chrisprice:
That makes sense. In which case, would it work for you if the code was changed to console.warn if the component is already registered rather than fail silently? And if so, are you happy to update the PR?

matthijsgroen:
Yes that would be a good thing... I will update the PR.

chrisprice (approved):

PR #1720, opened 2021-12-03
WeeklyPatternsDiscontinuityProvider

lorivic:
Sometimes the discontinuities follow weekly recurring patterns.
For instance, in trading, ES does not trade between 4 and 5 pm Central Time, between Monday and Thursday, and it does not trade between Fridays 5pm and Sundays 4pm Central Time.
Besides, certain symbols can have multiple non trading ranges during a day, e.g.: Corn which does not trade between 7:45 - 8:30 and 01:20 - 07:00 pm Central time.
Important to notice that these ranges should be relative to a specific timezone, which may not be the local or utc.
discontinuity.weeklypatterns.js implements the @d3fc/discontinuityProvider interface to address the above requirements.
time.js is a utility file to handle time functions. It is dependent on d3-time for floor-ing and ceil-ing dates and luxon for timezone conversion.

### DIFF
--- a/packages/d3fc-discontinuous-scale/src/discontinuity.weeklypatterns.js
+++ b/packages/d3fc-discontinuous-scale/src/discontinuity.weeklypatterns.js
@@ -1,0 +1,556 @@
+import "./node_modules/d3-time/dist/d3-time.min.js";
+import { TimeUtils } from "./time.js";
+
+class WeeklyPatternsDiscontinuityProvider {
+
+    //this class implements the DiscontinuityProvider interface to identify
+    //all dates processed in this class are expressed in UTC (see below, at the end of this file, how specific timezones are managed)
+
+    //we define:
+    //  calendarDistance:= (endDate-startDate)
+    //  continuousDistance := the calendarDistance minus the discontinuity ranges
+    //  gap:= a specific discontinuos range [startDate, endDate], startDate < endDate
+
+    constructor(TZ, NTH)
+    {
+        this.Time = new TimeUtils(TZ);
+        this.parse(NTH);
+
+        this.includedMillisPerWeek = this.getIncludedMillisInFullWeek();
+    }
+
+    //discontinuity range functions
+    parse(NTH) { //from json to proper data structure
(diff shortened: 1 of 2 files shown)


lorivic:
Usage as follows

let tz ='CT'; //central time

let nth = JSON.parse('{"Monday":[{"StartTime":"07:45","EndTime":"08:30"},{"StartTime":"13:20","EndTime":"19:00"}],"Tuesday":[{"StartTime":"07:45","EndTime":"08:30"},{"StartTime":"13:20","EndTime":"19:00"}],"Wednesday":[{"StartTime":"07:45","EndTime":"08:30"},{"StartTime":"13:20","EndTime":"19:00"}],"Thursday":[{"StartTime":"07:45","EndTime":"08:30"},{"StartTime":"13:20","EndTime":"19:00"}],"Friday":[{"StartTime":"07:45","EndTime":"08:30"},{"StartTime":"13:20","EndTime":"EOD"}],"Saturday":[{"StartTime":"SOD","EndTime":"EOD"}],"Sunday":[{"StartTime":"SOD","EndTime":"19:00"}]}');
//"SOD" = special string representing the start of day (internally converted in 00:00:00.000)
//"EOD"= special string representing the end of day (internally converted in 00:00:00.000, although conceptually different)

xScale.discontinuityProvider(skipWeeklyPatterns(tz, nth));

ColinEberhardt:
Thanks @lorivic that looks like a really useful addition, we'd be very happy to include this in D3FC.

There is quite a bit of logic within `WeeklyPatternsDiscontinuityProvider`, this will need some testing to ensure it works correctly. See the following for an example:

https://github.com/d3fc/d3fc/blob/master/packages/d3fc-discontinuous-scale/test/discontinuousSpec.js

Is this something you'd have time to do?

Other than that, there are a few minor stylistic changes needed, but nothing major.

ColinEberhardt:
Thanks you so much for this @lorivic - while we haven't used your code directly, it is a really useful feature which has been added via this PR #1753

PR #1730, opened 2022-05-16
Modified ohlc hlc

niteeshkr20:
Added a new Chart Type named HLC. That is more like OHLC but it does not show the open value.

**OHLC Changes -** 
Added a new parameter **ohlcShape** which will take either **_line_** or **_bar_** values. Based on that we will decide whether to draw a bar for OHLC or a line.


ColinEberhardt:
Thanks @niteeshkr20 - I think an HLC chart would be a really good addition. Ideally we'd want an implementation across canvas, SVG and WebGL.

PR #1794, opened 2023-06-21
chore: add publish script

chrisprice:


chrisprice:
I have low confidence this will work but there's only one way to find out...

PR #1808, opened 2024-04-04
chore: update lerna, migrate from lerna bootstrap

### LINKED ISSUE #1814: Modernise libraries
Many dependencies need updating, most of which have triggered an automatic PR. Merge these PRs to resolve the initial burse of dependency updates, then Investigate whether more updates are required. 

- [x] Merge chore(deps) PRs
- [ ] Fix deps with warnings
- [ ] D3
- [ ] Sveltekit?
- [ ] React?

jisted-scottlogic:
Updatjng Lerna to >v7 removes the lerna bootstrap command.
Their suggestion is to [migrate to using npm workspaces](https://lerna.js.org/docs/legacy-package-management)  for package management.

From what I can gather this PR does this migration correctly, but since the tests have been changed I would appreciate a look from someone more familiar with the project/package management to see that it still behaves as expected.


packages/d3fc-annotation/test/bundleSpec.js
```diff
@@ -9,10 +9,10 @@ describe('bundle', function() {
...
             scripts: [
                 require.resolve('d3/dist/d3.js'),
-                require.resolve('../node_modules/@d3fc/d3fc-data-join/build/d3fc-data-join.js'),
-                require.resolve('../node_modules/@d3fc/d3fc-rebind/build/d3fc-rebind.js'),
-                require.resolve('../node_modules/@d3fc/d3fc-series/build/d3fc-series.js'),
-                require.resolve('../node_modules/@d3fc/d3fc-shape/build/d3fc-shape.js'),
+                require.resolve('../../../node_modules/@d3fc/d3fc-data-join/build/d3fc-data-join.js'),
+                require.resolve('../../../node_modules/@d3fc/d3fc-rebind/build/d3fc-rebind.js'),
+                require.resolve('../../../node_modules/@d3fc/d3fc-series/build/d3fc-series.js'),
+                require.resolve('../../../node_modules/@d3fc/d3fc-shape/build/d3fc-shape.js'),
```
cfisher-scottlogic, line 15: How come this is necessary?

cfisher-scottlogic (commented):
My understanding of lerna is very limited, I just know it's responsible for 'building'. Does this mean if the app is building, running & testing successfully locally and in this CI pipeline, we can presume it will work in prod?

packages/d3fc-annotation/test/bundleSpec.js
jisted-scottlogic: It did seem a bit much to have ../../../ but I think that's what it takes to make these tests pass now that the node_modules are stored elsewhere. If we upgrade lerna and move away from bootstrap as the docs suggest then I think this is just where the new node_modules is located when using npm for the packages; they're managed a couple of levels up now instead.

I wasn't too sure how lerna worked either, but they did provide some migration info for this specific situation that I've been leaning on:
https://lerna.js.org/docs/legacy-package-management

cfisher-scottlogic (approved):

jisted-scottlogic:
> My understanding of lerna is very limited, I just know it's responsible for 'building'. Does this mean if the app is building, running & testing successfully locally and in this CI pipeline, we can presume it will work in prod?

I've experimented locally on a clean clone with this version and it behaves as I would expect, with npm doing what lerna did previously. We should probably do a further sanity check if merging, it should be obvious if it doesn't work, but I don't currently see why it wouldn't work the same in prod.

cfisher-scottlogic:
> > My understanding of lerna is very limited, I just know it's responsible for 'building'. Does this mean if the app is building, running & testing successfully locally and in this CI pipeline, we can presume it will work in prod?
> 
> I've experimented locally on a clean clone with this version and it behaves as I would expect, with npm doing what lerna did previously. We should probably do a further sanity check if merging, it should be obvious if it doesn't work, but I don't currently see why it wouldn't work the same in prod.

Cool, lets merge and do a sanity check after 🙂

PR #1832, opened 2024-06-07
Ignore dependabot commits

bethpritchard:
Dependabot is consistently breaking the `Code Standards` job. This should get commitlint to ignore all Dependabot commits.

image

image

### DIFF
--- a/.commitlintrc.js
+++ b/.commitlintrc.js
@@ -1,0 +1,5 @@
+module.exports = {
+    extends: ['@commitlint/config-conventional'],
+    ignores: [(message) => /^Bumps \[.+]\(.+\) from .+ to .+\.$/m.test(message)],
+  }
+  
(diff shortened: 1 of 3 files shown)


.commitlintrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+    extends: ['@commitlint/config-conventional'],
+    ignores: [(message) => /^Bumps \[.+]\(.+\) from .+ to .+\.$/m.test(message)],
```
jisted-scottlogic: I like this but I'm not sure all of the commit messages from dependabot start with this pattern, I don't think this would match the commits starting like this, for example:
`chore(deps): bump d3-color, d3-scale and @d3fc/d3fc-webgl`

You might have tried this but could we match the end of the message instead? ('Signed-off-by: dependabot... etc') The last line seems more consistent and it's more obvious what this ignores: block is doing. Something like this _should_ work but we might still have to merge to test on this branch:

```
module.exports = {
  extends: ['@commitlint/config-conventional'],
  ignores: [(message) => /Signed-off-by: dependabot\[bot]\s+<support@github\.com>$/m.test(message)],
}
```

bethpritchard:
Not sure if this will work but have to test on merge

jisted-scottlogic (approved):

PR #1842, opened 2024-06-17
WIP fix: specify publishConfig for main package

jisted-scottlogic:
Noticed there was an inconsistency with how our npm settings were set up, where the main package doesn't specify the 'access' field in its "publishConfig": {}, which all the other packages do.


bethpritchard:
Is this needed now?

jisted-scottlogic:
> Is this needed now?

It was just a shot in the dark but we're past that issue now, so I suspect if it's working without then this can only end worse!

PR #1855, opened 2024-06-25
Publish script

jisted-scottlogic:
Attempt to fix the bundle steps, and lack of /build directory on unpkg after the migration from lerna bootstrap.


jisted-scottlogic:
Current release isn't working because of missing build files after lerna and github actions update. Need to bypass rules to merge this but it seems justifiable to try to understand and fix the issue than leave it broken for longer.

PR #1857, opened 2024-06-26
fix: publish permissions

jisted-scottlogic:


jisted-scottlogic:
Further to last PR, which had incorrect permissions:

Current release isn't working because of missing build files after lerna and github actions update. Need to bypass rules to merge this but it seems justifiable to try to understand and fix the issue than leave it broken for longer.

PR #1858, opened 2024-06-27
fix: move to changesets

jisted-scottlogic:
After updating the pipelines and dependencies to the point where we could actually get to the publish step, I found 'lerna publish' to be deprecated now - one possible replacement seems to be changesets, which this branch is working towards migrating to if it works out.


jisted-scottlogic:
Release isn't working because of a deprecated publish command and missing build files after lerna and github actions update. Need to bypass rules to merge this again but it seems justifiable to try to understand and fix the issue than leave it broken for longer - there's only so much I can try to accomplish locally. So far these post-15.2.11 PRs haven't managed to do a release yet, even a bad one, but hoping the move to changesets works as advertised with this config.

PR #1861, opened 2024-07-02
chore: update rollup dependencies

bethpritchard:
Updates dependencies


.github/workflows/development.yml
```diff
@@ -13,14 +13,14 @@ jobs:
...
+        uses: actions/checkout@v4
         with:
             fetch-depth: 0
 
       - name: "Setup Node"
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
-          node-version: 21.7.3
+          node-version: 20.15.0
```
jisted-scottlogic, line 23: Just checking, does this need to be an earlier version than we had before?

bethpritchard: This deprecation warning was from the new version of Node - Stack Overflow suggested to use the LTS version of Node

image

jisted-scottlogic (approved):

PR #1863, opened 2024-07-03
chore: update jsdom

bethpritchard:
Updates JSDOM and fixes tests


packages/d3fc-data-join/test/bundleSpec.js
```diff
@@ -1,34 +1,41 @@
...
-                const selection = win.d3.select(win.document);
-                dataJoin(selection, []);
-                done();
-            }
+    it('should correctly wire-up all the dependencies via their UMD-exposed globals', function(done) {
+        const virtualConsole = new VirtualConsole().sendTo({
+            error: done
         });
-    });
-    it('should correctly wire-up all the dependencies via their UMD-exposed globals (w/ d3-transition)', function(done) {
```
bethpritchard, line 19: Couldn't for the life of me tell the difference between the `w/ d3-transition` and `w/o d3-transition` so removed

jisted-scottlogic: Well spotted, not sure what the intention was behind this extra test case either, but since it's identical let's leave it out.

packages/d3fc-series/test/bundleSpec.js
```diff
@@ -1,25 +1,41 @@
-import jsdom from 'jsdom';
+import { JSDOM, VirtualConsole } from 'jsdom';
+import fs from 'fs';
 
 describe('bundle', function() {
     it('should corectly wire-up all the dependencies via their UMD-exposed globals', function(done) {
```
jisted-scottlogic: Not too important but this file (d3fc-series) and the two below (d3fc-shape and d3fc-technical-indicator) don't have this 'corectly' typo fixed.

scripts/jest/utils.js
```diff
@@ -0,0 +1,21 @@
+import { JSDOM, VirtualConsole } from 'jsdom';
+import fs from 'fs';
+
+exports.createJSDOM = done => {
```
jisted-scottlogic: Does this work? I was trying to do something similar because the repetition was annoying me, but kept hitting a brick wall. I just tried locally again fiddling around with this branch instead but had similar results.
I can't see this currently being used or hooked up to anything, unless there's a plan for it?

.github/workflows/development.yml
```diff
@@ -7,6 +7,8 @@ on:
     branches: [ "master" ]
   workflow_call:
 
+  workflow_dispatch:
```
jisted-scottlogic: Is this deliberate? I first thought this might not do anything on its own but I see it might let us run the action manually, if that's the idea?
https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow

jisted-scottlogic (approved):

PR #1864, opened 2024-07-04
fix: add workflow to add release notes to fix releases

bethpritchard:
A separate workflow has been added to do releases. The workflow is triggered on the push of a tag, as getting the created tag from Changesets was difficult.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -53,12 +53,3 @@ jobs:
         run: npx changeset publish
         env:
           CI: true
-          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
-
-      - name: "Git tagging the main package"
-        run: |
-          MAIN_PACKAGE_VERSION=$(node -p "require('./packages/d3fc/package.json').version")
-          git tag "d3fc@$MAIN_PACKAGE_VERSION"
-          git push origin "d3fc@$MAIN_PACKAGE_VERSION"
-        env:
-          GH_TOKEN: ${{secrets.GITHUB_TOKEN}}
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -4,6 +4,8 @@ on:
   push:
     branches: [ "master" ]
 
+  workflow_dispatch:
```
bethpritchard: change before merge

jisted-scottlogic (approved):
Can't know for certain until we push tags but all looks good afaik.

README.md
```diff
@@ -100,11 +100,62 @@ This project is equipped with a Continuous Integration (CI) pipeline using GitHu
...
+```
+
+Changesets should include: 
+
+* WHAT the change is
+* WHY the change was made
+* HOW a consumer should update their code 
+
+#### Adding a changeset
+To add a changeset, before putting up a PR, contributors should:
```
jisted-scottlogic, line 131: Not really a comment, just been looking at how this works and want to check I understand something. So the changeset _doesn't_ need to be done just before merging to capture the most recent commit or anything?

My understanding is that the changeset will still capture all the commits on that PR, even if some commits were done after the changeset was created - and as long as the PR doesn't change drastically to the point where the changeset _description_ is no longer accurate then it should all work?

Suppose we could also add a github action check for this in future to check it's been done right. (to be clear I mean just checking a changeset of some kind has been created)

jisted-scottlogic (approved):
Looks good, just a clarification question.

README.md
bethpritchard: I've just done a test - the `changeset` script creates a changeset markdown file, in the format of

```md
---
'd3fc': minor
---

Test feature
``` 

`changeset publish` then captures all the commits - so as far as I am aware, you are correct and as long as commits after the changeset file is created don't touch another package or change anything drastically, contributors won't need to change the changeset. 

There's a changeset bot and github action that we can use for checking changesets - #1866

jisted-scottlogic: Ok thanks, based on what it generates that seems to be the way it works, cool.

PR #1868, opened 2024-07-05
Update D3 to 7.9.0

jisted-scottlogic:
Updates D3 to 7.9.0 - initial changes seem to be to locations, the moduleNameMapper and a lot of edits to node_modules paths. Need to test a bit more and read the changelogs for D3 to try identify any other issues.

I've refactored the base node_modules path out to a global jest setup variable to prevent having to change it everywhere again in future, though I could just revert to the string path if we prefer clarity within tests, it's an optional part of this update.

From what I've seen so far everything still works but a sanity check would be appreciated.


packages/d3fc-extent/test/bundleSpec.js
```diff
@@ -18,9 +18,11 @@ describe('bundle', function() {
             window.document.head.appendChild(scriptElement);
         };
         
-        const scripts = [
-            require.resolve('d3/dist/d3.js'),
-            require.resolve('../build/d3fc-extent.js')
+        const nodeModules = global.nodeModulesPath;
+
+const scripts = [
```
jisted-scottlogic: note to self: indentation seems to have gone wrong on some of these,

jisted-scottlogic:
Needs a changeset.

PR #1869, opened 2024-07-08
refactor: use changesets action for publish

bethpritchard:
Refactors the CD pipeline to use the Changesets github action, and added Changeset-bot. See [here](https://github.com/d3fc/d3fc/actions/runs/9837215380/job/27154624509) for test release pipeline.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches: [ "master" ]
 
+concurrency:  ${{ github.workflow }}-${{ github.ref }}
+
 jobs:
   build-test-lint:
     name: Build, test, and lint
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -14,42 +16,29 @@ jobs:
...
-        run: |
-          git config --global user.name 'github-actions[bot]'
-          git config --global user.email 'github-actions[bot]@users.noreply.github.com'
-          git remote set-url origin git@github.com:${{github.repository}}.git
 
       - name: "Install dependencies"
         run: npm ci
 
-      - name: "Running build"
-        run: npm run build
```
jisted-scottlogic, line 47: Maybe a silly question but do we not still need to do this build step as part of this? Can't see anywhere else it's done, so wondered if you figured out we don't need it for the publish?

bethpritchard: Good spot! My bad

jisted-scottlogic (approved):
Looks sensible to me, let's find out.

PR #1896, opened 2026-02-09
fix(webgl): add canFill and canStroke to diamnd point

HerrNaN:
## Symptoms

When using `webglFillColor()` or `webglStrokeColor()` via `.decorate()` on a WebGL point series with `d3.symbolDiamond`, the diamond points fail to render. The canvas is blank and the browser console shows WebGL errors: `useProgram: program not valid` and `no valid shader program in use`.

## Cause

The diamond fragment shader in `fragmentShaderSnippets.js` was missing `canFill` and `canStroke` variable declarations. When `webglFillColor()` or `webglStrokeColor()` is applied, it appends shader code that references these variables. Since diamond didn't define them, the GLSL shader failed to compile. Every other point type (circle, square, star, triangle, cross, wye) already defined both variables.

## Solution

Added `float canFill = 1.0;` and `float canStroke = smoothstep(vSize - 2.0, vSize, distance * vSize);` to the diamond fragment shader body, matching the pattern used by all other point types.

## Before (broken — blank canvas, WebGL shader errors in console)

before

## After (fixed — diamond points render correctly)

after

### DIFF
--- a/packages/d3fc-webgl/src/shaders/fragmentShaderSnippets.js
+++ b/packages/d3fc-webgl/src/shaders/fragmentShaderSnippets.js
@@ -123,6 +123,7 @@ export const diamond = {
         float b = 1.0;
     `,
     body: `
+        float canFill = 1.0;
         if (vDefined < 0.5) {
             discard;
         }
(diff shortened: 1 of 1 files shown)


HerrNaN:
@bethpritchard this is a very small fix do you know if it could be merged somewhat soon? I'll be happy to fix it up more of needed :)

